PR #7, opened 2017-02-02
Add gradle support

yanokwa:
* Added gradle support to build jar (`gradle jar`), run tests (`gradle test`) , and generate javadocs (`gradle javadoc`)
* Added .gitignore to ignore gradle directories
* Tried setting `sourceSets.test.resources.srcDirs = ['core/resources']` and that did not work, so changed PathConst.java so test resources could be found
* Confirmed the created jar can be put in ODK Collect and app can fill, save, and edit Widgets test form
* Confirmed all tests pass and outputs can be seen at `/build/reports/tests/test/index.html`
* Confirmed docs build and outputs can be seen at `/build/docs/javadoc/index.html`
* Did not test UML diagram generation because this might be complex and not important and should likely be filed as a separate issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+.gradle
+build
+gradle
+*.class
--- a/build.gradle
+++ b/build.gradle
@@ -1,0 +1,30 @@
+apply plugin: 'java'
+
+repositories {
+    mavenCentral()
+}
+
+sourceSets.main.java.srcDirs = ['core/src']
+sourceSets.test.java.srcDirs = ['core/test']
+
+dependencies {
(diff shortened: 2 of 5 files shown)


yanokwa:
@mitchellsundt, since you have the most experience with this, I want to make sure my changes to PathConst.java and my decision to punt on UML (are there any useful ones?) are reasonable.

If these changes are good, my plan is to add CircleCI support, then remove all the Eclipse, hg, ant files, and then trim and restructure the repo around `core/src` `core/test` and `core/resources`. Anything I should be careful about here?

PR #12, opened 2017-03-09
Fixed issue with daylight saving gap at midnight

### LINKED ISSUE #11: Daylight saving gap at midnight
When working on [Daylight saving gap](https://github.com/opendatakit/collect/issues/356) and [Daylight saving gap test](https://github.com/opendatakit/collect/pull/458) I noticed a bug in Javarosa.

The problem is in [this method](https://github.com/opendatakit/javarosa/blob/master/core/src/org/javarosa/core/model/utils/DateUtils.java#L116). When we pass 01/01/1960 00:00:00 in Nairobi timezone (there is a gap at midnight) the result is 31/12/1959. As I know the gap is between 00:00:00 - 00:15:00 if we use [DateTimeWidget](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/widgets/DateTimeWidget.java) in ODK we can change minutes to 20 for example and it's ok but if we use [DateWidget](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/widgets/DateWidget.java) it's a big problem and we can't use mentioned date.
Curiously the problem is realated only to Android and everything works well on JVM!
I added two tests to show that.
[This test](https://github.com/grzesiek2010/collect/commit/7ddb4ec95963ecbeb45c7e4214574afba9beaafd#diff-ebcd662d8b853be5586cd97fad0ab44e) is an unit test (unit tests run on the JVM) 
[This test](https://github.com/grzesiek2010/collect/commit/7ddb4ec95963ecbeb45c7e4214574afba9beaafd#diff-2d4dc61a97b9b58cd76c9d0f7e980767) in an instrumented test and it runs on the AVD.

In both test I test [current method](https://github.com/opendatakit/javarosa/blob/master/core/src/org/javarosa/core/model/utils/DateUtils.java#L116) and a new version of that method. 
As you can see in my tests the current method doesn't work well on Android but my version is ok.

grzesiek2010:
This PR is in reference to resolve #11

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,7 @@ sourceCompatibility = '1.7'
 dependencies {
     // Be sure to update dependencies in pom.xml to match
     compileOnly group: 'net.sf.kxml', name: 'kxml2', version: '2.3.0'
+    compile group: 'joda-time', name: 'joda-time', version: '2.9.9'
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'net.sf.kxml', name: 'kxml2', version: '2.3.0'
 }
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>joda-time</groupId>
       <artifactId>joda-time</artifactId>
-      <version>2.9.0</version>
+      <version>2.9.9</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,12 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <groupId>joda-time</groupId>
+      <artifactId>joda-time</artifactId>
+      <version>2.9.0</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>3.8.2</version>
--- a/src/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/org/javarosa/core/model/utils/DateUtils.java
@@ -25,6 +25,8 @@ import java.util.ArrayList;
 import org.javarosa.core.services.locale.Localization;
 import org.javarosa.core.util.MathUtils;
 
+import org.joda.time.LocalDateTime;
+
 /**
  * Static utility methods for Dates in j2me
  *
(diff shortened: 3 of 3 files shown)


mdudzinski:
I can confirm this works as expected and fixes the problem on Android.

lognaturel:
Thanks so much for taking a look at it @mdudzinski! Can you please let us know what API level you tried it with so we have an ongoing record of what has been verified? 

We do know there have been some surprising changes in date/time implementations between Android versions. I hope having a consistent version of `joda` shields us from any of those problems but I must admit that anything touching date/time and especially daylight savings time makes me nervous. I've been wanting to do a bit more reading on this approach to make sure it doesn't introduce any other unexpected behavior.

I do think we should get this fix in the May release so really appreciate your help looking at it!

mdudzinski:
@lognaturel 
We (in TaroWorks) tested ithis fix on Android 4.4, 5.1 and 6.0 and included it in the latest release.
So that's why I'm happy to confirm this code works. 

But I absolutely understand your concerns about messing with the date and date-time related code. I hope that the analysis below will resolve the doubts.

The Java Date API implementation on Android prior to 7 (API 24) treats DST gaps in a different way than it's expected. 

Example: 
1/1/1960 00:00 GMT+0300 (Africa/Nairobi)
This date is invalid because it lies within a DST gap (a time that does not
exist). If it happens that DateUtils#getDate is supposed to create such
Date then it should find the earliest valid time after the gap. 
It works on JVM when it returns 1/1/1960 00:15 but on Android it returns 31/12/1959 23:45 GMT+0300.

This fix is only about constructing the date object through the Joda time API. The logic - using DateFields and timezone parameters - stays the same. And as you noticed - both Collect and JavaRosa should rely on the same Date API. Either the default one or Joda. In my opinion Joda wins. It's widely used in the industry. I personally used it in Java projects (Android and Java EE as well) before and it always worked as expected. Plese also notice that there are several tests in JavaRosa that exercise this code and they pass.

We also have an instrumentation test case that tests this on Android. I think it's more valuable than the Roboeletric tests introducded in https://github.com/opendatakit/collect/pull/458 because it runs in JVM where everything is fine regardless of the fix added in https://github.com/opendatakit/collect/pull/382 so it never fails.
If you want I can share it in ODK Collect repo.  It may take me some time though because it's written in older manner (extending InstrumentationTestCase class).
The test fails without this fix in JavaRosa on Android versions prior to Android 7 so the Date API deviation seems to be fixed in Android 7. 

Moreover, with this fix in JavaRosa the manual skip the DST gap in the DateTime widget (https://github.com/opendatakit/collect/pull/382) is no longer needed.

lognaturel:
Thanks so much for your patience @grzesiek2010 and for that analysis @mdudzinski! 

@grzesiek2010 When you have a chance, please resolve the conflicts and then we'll be ready to merge. ✨

@mdudzinski We would very much appreciate the contribution of those Collect tests when you have a chance.

grzesiek2010:
@lognaturel Done

PR #14, opened 2017-03-16
Remove unnecessary files

yanokwa:
* Confirmed working by using clean jar in Collect build and trying widgets form
* Remove .hgignore because we use git
* Cleaned up notice.md for better presentation.
* Update location of src, resources, test files in gradle
* Update circle.yml to generate tests and jar artifacts
* Remove Eclipse files (.project, .classpath. build.xml)
* Remove core/tools/*.jar because we use gradle now
* Move src, resources, test from ./core/* to ./
* Remove util/* because it's unused or J2ME


lognaturel:
Hate to do this but this is not the kind of pull request we want to encourage. Can you please break it into smaller, targeted PRs that explain a group of related changes? For example, why doesn't the javadoc get generated on CircleCI anymore?

PR #27, opened 2017-04-04
Change all newlines to \n.

lognaturel:
Newline types are currently a mix of \n and \r\n. I think it would be easier moving forward to have that be uniform so that GitHub diffs aren't confusing when someone edits from NIX.

All I have done in this branch is run File > Line Separators > LF from IDEA J.

Is this a good idea to do? To do now? @dcbriccetti @yanokwa


dcbriccetti:
Fine with me.

lognaturel:
@yanokwa The explanation is [here](http://stackoverflow.com/questions/7013034/does-windows-carriage-return-r-n-consist-of-two-characters-or-one-character). When someone on *NIX opens a file with \r\n and saves it, \r is represented as ^M and that shows up in the GitHub diffs.

nribeka:
I really like this. Isn't there a git setting that force this for future commits? Or is it IDE setting?

lognaturel:
@nribeka [Here](https://github.com/checkstyle/checkstyle/issues/893) is a fun conversation on whether/how/where to enforce this kind of thing. Maybe we should file a separate issue and discuss there?

Interestingly, there have been a couple of times with Collect when I saw diffs like these and I couldn't figure out what was going on. Now all is clear. But it's only happened a couple of times so either we don't have a lot of Windows-based contributors or ...?

PR #28, opened 2017-04-04
Issue 25: Modernize XFormParser

dcbriccetti:
In progress


dcbriccetti:
@lognaturel, you might want to take a look and see if this work is going in the right direction for you. I’ve made significant code quality improvements.

src/org/javarosa/xform/parse/FormInstanceParser.java
```diff
@@ -0,0 +1,629 @@
+package org.javarosa.xform.parse;
+
+import org.javarosa.core.model.*;
```
lognaturel: I think for new files it makes sense to follow Google's [Java style guide](https://google.github.io/styleguide/javaguide.html) (no wildcard imports).

dcbriccetti: In general I agree with that (and certainly in this case). Not cluttering the namespace, showing exactly what’s used, etc. An exception is when the type contains something like constants, and they all are used. My view anyway.

src/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
```diff
@@ -48,39 +54,21 @@ DataBinding createBinding(IXFormParserFunctions parserFunctions, FormDef formDef
...
-            if ("true()".equals(xpathRO)) {
-                binding.readonlyAbsolute = true;
-            } else if ("false()".equals(xpathRO)) {
-                binding.readonlyAbsolute = false;
-            } else {
-                binding.readonlyCondition = getCondition(formDef, ref, xpathRO, "readonly");
-            }
-        }
-
+        setBoolAndCondition(ref, element, "relevant", formDef, new BoolAndConditionSetter() {
```
lognaturel: I'm not really convinced about this change. I feel it makes the code a lot harder to read for no significant gain that I can see!

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1631,127 +1631,34 @@ private boolean hasSpecialFormMapping(String textID, String locale) {
...
-                binding.relevantAbsolute = true;
-            } else if ("false()".equals(xpathRel)) {
-                binding.relevantAbsolute = false;
-            } else {
-                Condition c = buildCondition(xpathRel, "relevant", ref);
-                c = (Condition)_f.addTriggerable(c);
-                binding.relevancyCondition = c;
-            }
-        }
+    private final List<String> usedAtts = Collections.unmodifiableList(Arrays.asList(
```
lognaturel: I guess there are a number of different `usedAtts` lists. They could be sets but they're all relatively small so maybe it doesn't matter.

src/org/javarosa/xform/parse/IXFormParserFunctions.java
```diff
@@ -0,0 +1,17 @@
+package org.javarosa.xform.parse;
+
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.IDataReference;
+import org.javarosa.core.model.IFormElement;
+import org.javarosa.core.model.condition.Condition;
+import org.javarosa.core.model.condition.Recalculate;
+import org.javarosa.xpath.parser.XPathSyntaxException;
+
+/** Provides functions from XFormParser to classes split off from it, to avoid a cyclical dependency. */
```
lognaturel: So once it later became instantiable, then it became necessary, right?

lognaturel:
http://stackoverflow.com/a/3348855 might be useful for those wildcard imports.

src/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
dcbriccetti: I think my experience with functional languages led me here. This bit would look really clean in Scala. The point was to remove the code duplication among relevant, required, and readonly. Maybe improving the names would make it more readable? 

In any case, I’m happy to put it back as it was, if you like.

lognaturel: Makes sense, it does look like "code that is idiomatic in another language/context". I think if there were 17 or maybe even 7 or those things I'd be convinced that an abstraction was necessary but in this case it feels more like an obfuscation to me. :/

src/org/javarosa/xform/parse/IXFormParserFunctions.java
dcbriccetti: I’m trying to follow a practice where module A and module B each depend on the other. I don’t think it’s necessary as far as the compiler’s concerned. It’s a temporary solution.

src/org/javarosa/xform/parse/XFormParser.java
dcbriccetti: A fun fact: In some immutable Set implementations, when given a small set of values, no actual hash set or tree set is created. Instead the elements are saved in an array or something similar.

src/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
dcbriccetti: OK. I agree with your reasoning and will change this back.

dcbriccetti:
OK, I think I addressed everything. Let’s keep this open, thought, because after looking at Aggregate, I see I broke the interface with it, by changing code in overridden methods.

dcbriccetti:
Hi. I just tested with Collect and found no problems. Next are Aggregate and Validate. If you can accept some risk of breakage, you could merge this now.

lognaturel:
There's no rush so I think it's worth taking time to do Aggregate/Validate checks!

dcbriccetti:
This is now tested with Aggregate and Validate. Ready to merge.

PR #32, opened 2017-04-08
Add Abs function

dcbriccetti:
Fixes #4.

I did a rough, safe, cleanup pass on the affected files before implementing the feature.

Ready to merge. You may want to review the individual commit diffs with a better tool than github.com. None of IDEA, SourceTree, and Github Desktop are fooled by the cr/lf issues.

PR #33, opened 2017-04-08
Round problem, Issue #5

dcbriccetti:
This branch is off of `abs-function`. The changes for Issue #5 start at “Use assertEquals ...”.

See Issue #5 for discussion.


dcbriccetti:
This is ready to merge. It might be good to merge it now—it has improved tests—and then discuss in the issue what to do from here.

lognaturel:
Hmm, the #32 merge caused some conflicts.

dcbriccetti:
Sorry. Looks right now.

lognaturel:
Was going to merge this but I have pondered and I think we should see if the fix can come in this branch too. Would be nice to group them together!

dcbriccetti:
OK, this is ready if you agree with the solution.

Issue #16, opened 2017-03-17
Add support for external secondary instances


lognaturel:
#### Software versions 
JavaRosa v2016-04-21

#### Problem description
The status of external secondary instance support is unknown.

[External secondary instances](https://opendatakit.github.io/xforms-spec/#secondary-instances---external) are now part of the ODK XForms spec. As discussed in opendatakit/xforms-spec#50, there is at least partial support for the concept in JavaRosa currently.

I think the first step to addressing this should be to figure out what the state of the current implementation is and if/how it is usable. Then I believe there will need to be some design work done to make sure that the implementation is practical for large external secondary instances.

dcbriccetti:
I don’t see any evidence of support for external secondary instances in ODK’s JavaRosa. I looked at CommCare’s JavaRosa code and added just a bit of code to ODK’s to recognize the src attribute of the instance element. https://github.com/opendatakit/javarosa/pull/34 .

A next step might be to design a form and some tests, to define what we want to implement.

dcbriccetti:
Oh boy, it looks like Dimagi had to modify even the XPath code when generalizing FormInstance references to DataInstance (abstract parent of FormInstance and ExternalDataInstance). The concrete FormInstance occurs in more than 26 files in the xpath packgae, according to IDEA. I didn’t get a sense of this magnitude from Clayton Sims when we spoke with him on April 7. My notes from that call include the now-impenetrable “in code, ignore external data model, look for instance into actual dom”, which makes me want to have Clayton confirm that these extensive changes to the xpath package are required.

lognaturel:
Did you follow the links [here](https://github.com/opendatakit/xforms-spec/issues/50#issuecomment-277174668)? In particular, [the loading of secondary instances](https://github.com/opendatakit/javarosa/blob/3eb73b5c8c1d56b9bd635b1aca6a8e9ab1051fa9/core/src/org/javarosa/xform/parse/XFormParser.java#L439) and [the map they are stored in](https://github.com/opendatakit/javarosa/blob/3eb73b5c8c1d56b9bd635b1aca6a8e9ab1051fa9/core/src/org/javarosa/core/model/FormDef.java#L206). I believe `DataInstance` is for convenience/speed.

I agree that some sample forms and tests around them would be helpful.

dcbriccetti:
There is support for secondary instances on ODK JavaRosa, but not _external_ ones. Or are you talking about something else?

lognaturel:
[...]

PR #34, opened 2017-04-09
External secondary instances

dcbriccetti:
Work in progress. Builds on earlier branches. First commit is “	Begin parsing the src attribute...”.


lognaturel:
Any chance this can be a separate branch rather than building off the refactor?

dcbriccetti:
I don’t think so, but I’m happy to resolve conflicts before you merge. But first we have to test the first branch of changes against Aggregate. That was blocked because I couldn’t get Aggregate to build, but I see there’s been some conversation about that.

dcbriccetti:
@lognaturel @ctsims If you have a moment, I’d love your quick thoughts on my first attempt to read and parse an external XML file: https://github.com/opendatakit/javarosa/pull/34/commits/a7da3b7c26a9e8776b070bde32610f69ba16f94f

lognaturel:
Are there any specific decisions that you had to make that you're unsure about? Implications for this structure rather than another you might have considered? Were you able to learn from the way that CommCare approached it?

I notice that that `ExternalDataInstance` lives in the `utils` package but it seems that maybe it should be directly in `instance`?

dcbriccetti:
Some good progress: I found TreeElementParser from commcare-core, which translates the XML from the external secondary instance file into the right form. Now I’m trying to get this to not blow up:
```
        TreeReference treeReference = ((XPathPathExpr)
                parseXPath("instance('towns')/towndata/data_set")).getReference();
        EvaluationContext evaluationContext = formDef.getEvaluationContext();
        List<TreeReference> treeReferences = evaluationContext.expandReference(treeReference);
```

ctsims:
Hi @dcbriccetti One thing I will mention: 

In general we differentiate between two kinds of XML documents that we parse. One type are the "in-memory" parses, of which you are doing here (where you parse the full DOM into memory and then manipulate it). 

We also (in a few places) support parsing the XML stream "In-place" by walking over the elements with a fixed memory space. You can see this with our "ElementParser" and implementing classes. 

We use this in spots where we expect the documents that we are parsing to be too large to reliably fit in memory (our user data restores fall into this category), and we generally use that approach as the default now (v. parsing into a full virtual DOM) because it keeps the heap free and avoids issues when documents are larger than expected. If you are hoping to parse _large_ documents, I'd suggest that approach. If you are parsing small docs (100's of kb's) this approach should be fine.

dcbriccetti:
By modifying the xpath in the `parsesExternalSecondaryInstanceForm` test, I got past the exception that’s had me stuck for awhile. I’m not sure whether things are working right though. @ln or @yanokwa, maybe we could look at it together live, and then if needed, we can ask @ctsims for some more guidance?

lognaturel:
Yay for getting unstuck! I'm pretty tied up with the migration to Discourse but let's try to schedule for something early next week -- I'll message you in Slack. In the mean time, if you had some cycles to look into #60, that would be awesome.

dcbriccetti:
Hi @lognaturel and @ctsims! Could I trouble you to look at the XML and code below and see if you think the external XML file is being processed and tested correctly? Thanks!

screen shot 2017-06-30 at 9 52 19 pm

lognaturel:
@dcbriccetti That test is looking right to me. The GitHub diffs currently include a bunch of unrelated changes and I think that's because of some of the merges. I think rebasing on top of master would fix it, is that something you could try?

It would be great to get some other JR-familiar eyes on it and @mdudzinski is available to do some review. @mdudzinski as a first pass it would be helpful to get your thoughts on the approach and verify that the test added matches what the spec promises (longer discussion at opendatakit/xforms-spec#50 and some approach discussion and links to CommCare in #16).

Provided this is all looking good, @dcbriccetti, what do you think next steps are? One big question is performance as discussed [here](https://github.com/opendatakit/javarosa/pull/34#issuecomment-303224301). I think it would be worth doing some quick and dirty benchmarking, perhaps using a memory-limited Android emulator, to get a sense of how big of external instances this can handle.

dcbriccetti:
Hi! I’ll attempt a rebase on master now.

dcbriccetti:
That should complete the rebase on master.

dcbriccetti:
@lognaturel wrote:

> ... what do you think next steps are? One big question is performance ...

Sure, I can get started on this while @mdudzinski—Hi!—reviews the current work.

lognaturel:
@dcbriccetti Following up on a conversation on #javarosa-code this morning. You were saying it's getting hard to keep this branch in sync which is absolutely the case and I'm sorry about that! Let's commit to not merging anything else in until we have a decision on how this will move forward. I should really have thought about this before merging in #151.

Before we even tackle coming up with a plan, I'd like to understand the state of things a little better because I'm not seeing what I would expect in the diff. For example, .hgtags should have been added back in (7b9e2e1). It looks like this changeset [removes the checkstyle file](https://github.com/opendatakit/javarosa/pull/34/files#diff-8cb95faaaa6ece87877ab28d9a45bf36) which shouldn't happen. Line break changes are showing up when they should be merged in.

I'm not really able to tell quickly from the history how these things happened but they make me nervous that this is not on a path to being safe to merge. With so many invasive changes, I think it's really important to be able to look at only relevant code when evaluating it. @mdudzinski what do you think?

@dcbriccetti do you think it's possible to untangle all of this? A couple of alternate options that occur to me and which may not be any good are to treat what's in this branch as a prototype, not worry about merge conflicts and rebuild a thoroughly documented, mergeable version once the solution is well-understood. Another is to cherry pick the relevant changes onto a fresh branch off master.

dcbriccetti:
OK, how about when we finish up
https://github.com/opendatakit/javarosa/pull/153 (Issue 60: Allow http://openrosa.org/xforms namespace for meta)
and get it merged to master I’ll make a new branch from master and reapply these changes. I may recommend that I do the refactorings first (the bulk of the changes), and get that into master, then when we are confident the new external secondary instances feature is correctly implemented, we’ll merge that separately.

mdudzinski:
@lognaturel @dcbriccetti 
I agree the branch should be cleaned-up. Besides changes that seem to be wrong - like .hgtags and checkstyle.xml being deleted - there are a lot of formatting changes (mostly tabs to spaces) which completely obfuscate what has actually changed. If any. For example -> [src/org/javarosa/core/model/DataBinding.java](https://github.com/opendatakit/javarosa/pull/34/files#diff-157fecda8a4bc50f3d89649ea02d4766)

@dcbriccetti Having the refactoring first and the actual implementation next sounds like a great idea. I'd first remove the tabs - perhaps for the whole project? - and later do the refactoring that actually modernizes the code. This way we'll clearly see what code exactly has changed and what it impacts. And it will be easier to find eventual bugs during the review.

Cherry-picking onto a fresh branch off the master proposed by @lognaturel may be the easiest way to do this. 

@dcbriccetti Honestly the above sounds like a really cumbersome work. Let me know if I can help you somehow.

dcbriccetti:
Replacing this branch with a new one from master, with this feature’s changes manually reapplied.

ctsims:
Confirming that the tests you've run seem correct, and that looks like the
right structure to me.

If you are looking for stress tests I would also try to test using external
instances with attributes, and context switches (IE :

instance('a')/data/item[@id = instance('b')/data/id]/output

to make sure those work correctly, as that's where the faults are likely to
start showing themselves

PR #38, opened 2017-04-25
WIP Fixed bug with constarint and null

grzesiek2010:


lognaturel:
Upon further reflection, this appears to be by design. `constraint` only applies to non-null values and `required` must be used in the case of a value that is required.

PR #42, opened 2017-04-25
Issue 41: Add support for contains(), starts-with(), ends-with() functions

dcbriccetti:
This branches off of “round-problem”.  This commit contains the non-cosmetic part of the change:
https://github.com/opendatakit/javarosa/commit/1a4f7cd5e6dbbd69bfa01f00fa0f94e3d844a6c5


lognaturel:
The more focused the PR, the less likely we are to introduce issues! I like the refactoring here but it would have been a lot easier to review separately. Overall, looks good, but I get a failing test
```
junit.framework.AssertionFailedError: expected:<12345.2> but was:<12345.1>
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.failNotEquals(Assert.java:329)
	at junit.framework.Assert.assertEquals(Assert.java:120)
	at junit.framework.Assert.assertEquals(Assert.java:129)
	at junit.framework.TestCase.assertEquals(TestCase.java:288)
	at org.javarosa.xpath.test.XPathEvalTest.testEval(XPathEvalTest.java:93)
	at org.javarosa.xpath.test.XPathEvalTest.testEval(XPathEvalTest.java:63)
	at org.javarosa.xpath.test.XPathEvalTest.doTests(XPathEvalTest.java:315)
```

dcbriccetti:
Interesting! Of course, all the tests pass for me. I wonder how I can reproduce. I’ll investigate.

dcbriccetti:
Maybe a difference in our default rounding modes?
https://docs.oracle.com/javase/7/docs/api/java/math/RoundingMode.html
Where is this error coming from? Can we step through in the debugger?

lognaturel:
Oh wow, it didn't even occur to me that we might be getting different results. That's disconcerting! Trying it in the debugger in a sec. In the mean time, it would be interesting to get this branch run on continuous integration. That should now happen automatically for new pull requests but for this one, I think you'll have to go [here](https://circleci.com/gh/opendatakit/javarosa), log in with your GitHub account and then trigger the build.

dcbriccetti:
I’m making a pull request with explicitly setting the rounding mode to HALF_UP.

lognaturel:
The tests that fail are the following so you seem to be right about the root cause.
```
testEval("round('12345.15',     1)", 12345.2);
testEval("round('-12345.15',    1)", -12345.2);
```
I wonder how I didn't see that when merging `round`. 😖

dcbriccetti:
I was able to reproduce when I explicitly set the rounding mode to HALF_EVEN.

dcbriccetti:
Summarizing a conversation elsewhere: This is a Java 7/8 difference. We’ll disable the tests for now, and revisit when JavaRosa is built for Java 8.

PR #43, opened 2017-04-26
Fixed BooleanData

grzesiek2010:
https://github.com/opendatakit/collect/issues/908


lognaturel:
We are going to need to think long and hard about this one. Encouraging the use of inputs with boolean types and using `true` and `false` as the primitives might make for some surprising results when using these values in XPath expressions. opendatakit/xforms-spec#115

lognaturel:
JavaRosa also doesn't follow all XPath rules when coercing values to boolean: https://forum.getodk.org/t/relevance-calculated-not-working-on-enketo/26587. If we want to revive this we have to start with the spec work.

PR #46, opened 2017-05-05
Explicitly set the rounding mode to HALF_UP, to avoid potential probl…

dcbriccetti:
…ems with different defaults in JVMs.


lognaturel:
Hmm, the same tests are still failing!

I'm going to fix the checkstyle issues really quick so we can get a clean run on CircleCI.

lognaturel:
Looks like you're exploring another approach. The build is failing [on CI](https://circleci.com/gh/opendatakit/javarosa/32#artifacts/containers/0). Sorry I didn't notice this earlier, it would have made more sense to fix with the rounding!

PR #61, opened 2017-06-05
#45 repeat-groups are not respecting the jr:count setting after a change to a lower value

### LINKED ISSUE #45: repeat-groups are not respecting the jr:count setting after a change to a lower value
#### Software versions 
All JavaRosa versions

#### Problem description
It's reported in https://opendatakit.org/help/form-design/workarounds/ under "repeat-groups are not respecting the jr:count setting after a change".

Also reported [in ODK Collect issues](https://github.com/opendatakit/collect/issues/868).

* Open a form that has a repeat group driven by the `jr:count` attribute which is a path to a node that holds the actual value.
* Go to the question referenced by `jr:count` attribute and put any value. For example 5.
* Go to the 5th repeat group instance.
* Go back to the question referenced by `jr:count` attribute and put any value less that previously entered. For example 3.
* Go back to the 4th or the 5th repeat group instance.

#### Expected behavior
* The 4th and 5th repeat group instances should be deleted.

#### Other information 
It seems like repeat group instances are created inside `FormEntryModel.createModelIfNecessary(FormIndex index)` method. The following snippet
```
if (index.getTerminal().getInstanceIndex() < fullcount) {
   try {
     getForm().createNewRepeat(index);
    } catch (InvalidReferenceException ire) {
      ire.printStackTrace();
      throw new RuntimeException("Invalid Reference while creting new repeat!" + ire.getMessage());
    }
 }
```
checks if amount of repeat section instances is less than the `jr:count` value (`fullcount`). If it is less, a new repeat group instance is created.

The above code runs every time `FormEntryController.jumpToIndex(FormIndex index)` is called so mostly upon `FormEntryController#stepToNextEvent()` and `FormEntryController.stepToPreviousEvent()`. 
This means a new repeat instance is created every time a user navigates to it. 

I think it should be quite easy to remove extraneous repeat group instances in similar manner. However, if the user `jumps over` the repeat group, this code will not be called.
I think it can be fixed by an additional code that traverses the whole form instance before form instance is saved. I believe that the `IDag.validate(FormEntryController formEntryControllerToBeValidated, boolean markCompleted)` may do the trick.

mdudzinski:
Fixes #45.

I hope commits messages, code and comments are comprehensive but here goes a short summary. 

The code responsible for actual deletion of the extraneous repeats is simple:
```
   do {
     getForm().deleteRepeat(index);
   } while (getForm().getMainInstance().resolveReference(index.getReference()) != null);
```
but  after the deletion, `index` points to the non-existing repeat group which leads to an error shown on UI (ODK Collect) . To fix this, the index must be incremented after the deletion and before it's assigned to `currentFormIndex`. This leads to another problem - a non-relevant (hidden) form element may be present right after the deleted repeat group. Skipping over such indices takes place *before* updating the model so it must be repeated. It's achieved by checking the current index after `jumpToIndex(index)` and recursively calling `stepEvent(forward)` if needed.

### DIFF
--- a/resources/limited-repeats-count-non-relevant-index-after-repeat-group.xml
+++ b/resources/limited-repeats-count-non-relevant-index-after-repeat-group.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa">
+    <h:head>
+        <h:title>limited repeats count</h:title>
+        <model>
+            <instance>
+                <data id="limited-repeats-count">
+                    <repeat_count/>
+                    <limited_repeat>
+                        <free_text_1/>
+                        <free_text_2/>
+                    </limited_repeat>
+                    <non_relevant_text/>
+                </data>
+            </instance>
+            <bind nodeset="/data/repeat_count" type="int"/>
+            <bind nodeset="/data/limited_repeat" relevant="/data/repeat_count > 0"/>
+            <bind nodeset="/data/limited_repeat/free_text_1" type="string"/>
+            <bind nodeset="/data/limited_repeat/free_text_2" type="string"/>
+            <bind nodeset="/data/non_relevant_text" type="string" relevant="false()"/>
+        </model>
+    </h:head>
+    <h:body>
+        <input ref="/data/repeat_count">
+            <label>repeat count</label>
+        </input>
+        <group>
+            <label>limited repeat</label>
+            <repeat nodeset="/data/limited_repeat" jr:count="/data/repeat_count"
+                    jr:noAddRemove="true()" appearance="field-list">
+                <input ref="/data/limited_repeat/free_text_1">
+                    <label>free text 1</label>
+                </input>
+                <input ref="/data/limited_repeat/free_text_2">
+                    <label>free text 2</label>
+                </input>
+            </repeat>
+        </group>
+        <input ref="/data/non_relevant_text">
+            <label>should not be visible</label>
+        </input>
+    </h:body>
+</h:html>
--- a/resources/limited-repeats-count.xml
+++ b/resources/limited-repeats-count.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa">
+    <h:head>
+        <h:title>limited repeats count</h:title>
+        <model>
+            <instance>
+                <data id="limited-repeats-count">
+                    <repeat_count/>
+                    <limited_repeat>
+                        <free_text_1/>
+                        <free_text_2/>
+                    </limited_repeat>
+                </data>
+            </instance>
+            <bind nodeset="/data/repeat_count" type="int"/>
+            <bind nodeset="/data/limited_repeat" relevant="/data/repeat_count > 0"/>
+            <bind nodeset="/data/limited_repeat/free_text_1" type="string"/>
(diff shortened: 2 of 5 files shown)


lognaturel:
Thank you @mdudzinski! Will try to review soon. I want to make sure I get a chance to think through implications.

In the mean time, other comments and reviews are welcome!

lognaturel:
@mdudzinski This has not been forgotten, I promise. As you have seen on the forum, there are a lot of exciting things happening all at once in the ODK community. I want to make sure we are thoughtful in assessing the impact of this change.

mdudzinski:
@lognaturel no worries, I understand. This is being tested on our side so we'll let you know the results here. In the meantime, I'll appreciate any technical input because I still have a feeling that I missed something in the implementation:)

mdudzinski:
@lognaturel Just to let you know that this change has been released with the latest TaroWorks version. It's been working fine so far but TaroWorks uses only a subset of Collect/JavaRosa XForm capabilities and therefore it would be nice to have another pair of eyes to look at this.

lognaturel:
Thank you for that update, @mdudzinski!

Did you end up including some kind of notification or confirmation for the user? Since this is a change in existing behavior, I have mild concerns about data loss. It should probably be clear to a user when they change the value that they could be losing records.

mdudzinski:
@lognaturel Nope. TW didn't have a requirement to inform the users about potential data loss. But I can imagine that some Collect's could be surprised. 
That's something to be implemented on the client's. But since a field that is referenced by jr:count doesn't actually know it's referenced - it's not easy. The other challenge is creating or deleting repeat groups "in place". This means it happens where a user actually tries to go to the repeat group that is supposed to be created or deleted. So nothing happens if the user changes the value to a lower one, go somewhere (but not to the repeat group's instances that will be deleted:)) and go back and change the value to the original one. Perhaps JavaRosa should incorporate a some kind of observer pattern (RxJava?) so it could emit messages about things that's going to happen. Client apps could then respond to those messages. For example by displaying a pop-up informing about potential data loss. Just a rough idea.

grzesiek2010:
@mdudzinski 
I have got this form: 
[repeatTest.xml.txt](https://github.com/opendatakit/javarosa/files/1487736/repeatTest.xml.txt)
it contains:
1. Select one Question (Yes/No)
2. Numeric Question (visible only if you selected yes)
3. Repeat group where the count of this group depends on the second question.

If I select yes then I pass 3 for example I can see 3 groups. :white_check_mark: 
If I go back and change 3 to 2 I can see two groups. :white_check_mark: 
If I go back and select `no` instead of `yes` the second question is not visible but I stiill can see all repeat groups. :bug: 

The workeround here is to add:
      <bind nodeset="/repeatTest/q3" relevant="selected(/repeatTest/q1, 'yes')"/>
to the repeat group as well.

My question is: isn't it something you should/could fix as a part of this pr too?

mdudzinski:
@grzesiek2010 Hmm this is a quite different scenario. When `createModelIfNecessary` looks for the count value it gets null. If null, then method returns and no new group is added (in the existing code) and no group is deleted (in the code from this PR). It's technically possible to delete all repeats if this happens but I think it should be discussed. I believe @lognaturel has a lot of concerns about the data loss possibility.

lognaturel:
Agreed that is pretty different although it's true that the effect feels similar to users.

@mdudzinski is right that I'm concerned about accidental data loss. That's why this PR hasn't been merged -- the behavior has previously been considered a feature not a bug. Search for "repeat-groups are not respecting the jr:count setting after a change" on the page https://opendatakit.org/help/form-design/workarounds/. You'll see that the recommendation is to use relevance to hide any extra values. This works really well and means that if someone had entered 10, wanted to change the value to 12 but accidentally entered 2 they wouldn't lose a bunch of records.

lognaturel:
This continues to be considered intended behavior.

Issue #52, opened 2017-05-23
Limit and document the public interface


dcbriccetti:
There are many public types and methods in JavaRosa that could have a more restrictive access level. Knowing that something is not used from outside can allow removing or refactoring it. 

Determine exactly what is to be included in the JavaRosa public interface. Restrict access to the rest using access level modifiers, and document the public interface, perhaps using [Doc Comments](http://www.oracle.com/technetwork/articles/java/index-137868.html). Allow for public comment, to reduce the risk of making breaking changes.

dcbriccetti:
[JavaRosa Client Cross Reference.xlsx](https://github.com/opendatakit/javarosa/files/1164067/JavaRosa.Client.Cross.Reference.xlsx)
I created this by processing the output from IDEA’s Analyze Dependencies feature (with a little Scala program I whipped up to process the XML). I want to ensure that this is accurate, so perhaps people who work on the clients could do a quick sanity check. Thanks!

PR #63, opened 2017-06-06
Submission profile

grzesiek2010:
I'm working on https://github.com/opendatakit/collect/issues/791.
If I add `<submission orx:auto-delete="true" orx:auto-submit="true" />` cording to the description I can't open my form because the app crashes:

> java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
                                                                           at java.nio.charset.ModifiedUtf8.countBytes(ModifiedUtf8.java:75)
                                                                           at java.nio.charset.ModifiedUtf8.encode(ModifiedUtf8.java:119)
                                                                           at java.io.DataOutputStream.writeUTF(DataOutputStream.java:197)
                                                                           at org.javarosa.core.util.externalizable.ExtUtil.writeString(ExtUtil.java:147)
                                                                           at org.javarosa.core.model.SubmissionProfile.writeExternal(SubmissionProfile.java:81)
                                                                           at org.javarosa.core.util.externalizable.ExtUtil.write(ExtUtil.java:98)
                                                                           at org.javarosa.core.util.externalizable.ExtWrapMap.writeExternal(ExtWrapMap.java:152)
                                                                           at org.javarosa.core.util.externalizable.ExtUtil.write(ExtUtil.java:98)
                                                                           at org.javarosa.core.model.FormDef.writeExternal(FormDef.java:1276)
                                                                           at org.odk.collect.android.tasks.FormLoaderTask.serializeFormDef(FormLoaderTask.java:503)
                                                                           at org.odk.collect.android.tasks.FormLoaderTask.doInBackground(FormLoaderTask.java:169)
                                                                           at org.odk.collect.android.tasks.FormLoaderTask.doInBackground(FormLoaderTask.java:73)

I've fixed the issue and cleaned the code in that class.

### DIFF
--- a/src/org/javarosa/core/model/SubmissionProfile.java
+++ b/src/org/javarosa/core/model/SubmissionProfile.java
@@ -70,7 +70,7 @@ public class SubmissionProfile implements Externalizable {
 	@SuppressWarnings("unchecked")
 	public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
 		ref = (IDataReference)ExtUtil.read(in, new ExtWrapTagged(IDataReference.class));
-		method = ExtUtil.readString(in);
+		method = ExtUtil.nullIfEmpty(ExtUtil.readString(in));
 		action = ExtUtil.nullIfEmpty(ExtUtil.readString(in));
 		mediaType = ExtUtil.nullIfEmpty(ExtUtil.readString(in));
 		attributeMap = (HashMap<String, String>)ExtUtil.read(in, new ExtWrapMap(String.class, String.class));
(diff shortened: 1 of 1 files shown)


lognaturel:
The crash fix looks good but I'm nervous about the change in the public API. Did you verify that the tools in the immediate ODK ecosystem don't somehow rely on them? Hopefully not but we've found some strange things. Perhaps it would be best to do that with #52 and request comments before making the change?

grzesiek2010:
@lognaturel you are right I shouldn't implement additional - not related changes like code improvements. Sometimes I just can't stop myself :)

lognaturel:
There is a lot of great refactoring to do! 

If it doesn't affect public interfaces, go ahead and do it as a separate pull request, that is always welcome. But if it does affect public interfaces, I do think we should be careful about how it's approached.

PR #65, opened 2017-06-06
#64 The `round(num, decimalPlaces)` function doesn't work on locales that use comma as the decimal separator.

### LINKED ISSUE #64: The `round(num, decimalPlaces)` function doesn't work on locales that use comma as the decimal separator.
#### Software versions 
JavaRosa since 7d0e596 (PR #33 , Issue #5).

#### Problem description
The `round(num, decimalPlaces)` function doesn't work on locales that use comma as the decimal mark (Polish Locale for example:)).

I reproduced this problem by running the test suite on my machine which has PL locale.
The suite crashes with the following information:

> java.lang.NumberFormatException: For input string: "14,3"
	at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2043)
	at sun.misc.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
	at java.lang.Double.parseDouble(Double.java:538)
	at java.lang.Double.valueOf(Double.java:502)
	at org.javarosa.xpath.expr.XPathFuncExpr.round(XPathFuncExpr.java:1055)
	at org.javarosa.xpath.expr.XPathFuncExpr.eval(XPathFuncExpr.java:200)
	at org.javarosa.xpath.test.XPathEvalTest.testEval(XPathEvalTest.java:86)
	at org.javarosa.xpath.test.XPathEvalTest.testEval(XPathEvalTest.java:65)
	at org.javarosa.xpath.test.XPathEvalTest.doTests(XPathEvalTest.java:322)

This comes from test for "round('14.29123456789', 1)" expression but I checked and the suite will fall for any round expression that have more than 0 as the power parameter.

#### Expected behavior
The test just passes.

#### Other information 
The following code causes this behaviour:
```
            final NumberFormat nf = NumberFormat.getNumberInstance();
            nf.setMaximumFractionDigits(numDecimals);
            nf.setGroupingUsed(false);
            return Double.valueOf(nf.format(number));
```
It uses NumberFormat's capabilities to round a decimal number. It works great but this class follows the current locale settings and produces a readable string that matches the locale's rules. The result is later passed to `Double.valueOf` which doesn't handle comma as the decimal separator.

For example, in Poland `nf.format(number)` for `number=14.29123456789` and `numDecimals=1` will return `14,3` instead of `14.3`.

mdudzinski:
Fixes #64.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -49,6 +49,7 @@ import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.regex.Pattern;
 
 import static java.lang.Double.NEGATIVE_INFINITY;
@@ -1045,7 +1046,10 @@ public class XPathFuncExpr extends XPathExpression {
         }
 
         if (numDecimals >= 0) {
-            final NumberFormat nf = NumberFormat.getNumberInstance();
+            // Some locales use ',' instead of '.' as a decimal separator. Since Double.valueOf
+            // doesn't handle comma as the decimal separator, we must ensure a dot will be used by
+            // forcing the US format locale.
+            final NumberFormat nf = NumberFormat.getNumberInstance(Locale.US);
             nf.setMaximumFractionDigits(numDecimals);
             nf.setGroupingUsed(false);
             return Double.valueOf(nf.format(number));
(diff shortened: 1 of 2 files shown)


dcbriccetti:
Whoops! I caused this recently. Thanks for fixing it. I’ll make a note to always consider the effect of different locales.

lognaturel:
Ooh, I use Collect in French so I really should have noticed @dcbriccetti, don't worry.

Thanks for the detailed issue @mdudzinski and for the nice comments/tests! Will be good for avoiding this kind of problem in the future.

PR #67, opened 2017-06-07
TreeElement#populate now handles leafs attributes.

mdudzinski:
TreeElement#populate method was completely ignoring leafs attributes. This fix just takes the code responsible for populating attributes out from the if-else block so it runs for all kind of nodes.

XML attributes come in handy when there is a need to maintain a metadata in form instances.

Added a simple unit test that covers this changes.

### DIFF
--- a/resources/populate-nodes-attributes-instance.xml
+++ b/resources/populate-nodes-attributes-instance.xml
@@ -1,0 +1,7 @@
+<?xml version='1.0' ?>
+<data id="populate-nodes-attributes" xmlns:cns="custom_name_space">
+    <free_text_1 custom_attr_1="xyz1">Abc</free_text_1>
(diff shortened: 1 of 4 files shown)


lognaturel:
Very interesting, I had not noticed that this was a limitation. Do you have a use case in mind or already implemented? 

On the Collect end of things, we are exploring adding a lighter weight [timing audit](https://github.com/opendatakit/collect/issues/257) kind of feature and we might use something like a `total-time-in-question` attribute.

mdudzinski:
@lognaturel Yes. We've got a couple of features that rely on this. For example certain questions may be auto-populated when the form is loading so we want to "mark them" so later they can be "disabled" on the UI.

It sounds like a good idea to keep the time in an attribute like "total-time-in-question" and this code will definitely help.

PR #68, opened 2017-06-09
Issue 60: Allow http://openrosa.org/xforms namespace for meta

dcbriccetti:
So far, we have a failing test to illustrate the problem.

### DIFF
--- a/resources/meta-namespace-form.xml
+++ b/resources/meta-namespace-form.xml
@@ -1,0 +1,17 @@
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms">
+  <h:head>
+    <h:title>Audit test ORX</h:title>
+    <model>
+      <instance>
+        <data id="audit-test-orx">
+          <orx:meta>
+            <orx:audit/>
+          </orx:meta>
+        </data>
+      </instance>
+      <bind nodeset="/data/orx:meta/orx:audit" type="binary"/>
+    </model>
+  </h:head>
+  <h:body>
+  </h:body>
+</h:html>
--- a/resources/meta-namespace-form.xml
+++ b/resources/meta-namespace-form.xml
@@ -1,17 +1,20 @@
 <h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml"
-        xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xsd="http://www.w3.org/2001/XMLSchema"
-        xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms">
+        xmlns:orx2="http://openrosa.org/xforms">
     <h:head>
         <h:title>Namespace for Metadata</h:title>
         <model>
             <instance>
-                <data id="audit-test-orx">
-                    <orx:meta>
-                        <orx:audit/>
-                    </orx:meta>
+                <data id="test">
+                    <orx2:meta>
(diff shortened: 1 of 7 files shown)


lognaturel:
Thank you, @dcbriccetti! Looking like great progress. Let me first make sure I understand what's being accepted at a high level. In the test form, you have 
```
<orx2:meta>
    <orx2:audit/>
    <orx2:audit2/>
    <audit3/>
</orx2:meta>

<bind nodeset="/data/orx2:meta/orx2:audit"/>
<bind nodeset="/data/meta/audit2"/>
<bind nodeset="/data/orx2:meta/audit3"/>
```
This nicely shows that the prefix is irrelevant which is great.

The `orx2:audit` case is the ideal case -- both `meta` and `audit` have proper and matching namespaces. Looks like this is handled perfectly by what you have now.

The next two cases I'm less sure about. I see that the code accepts name-only matches before checking for namespaces. I'm not sure a mismatch between the node namespace and the path in the `nodeset` should be allowed. Did you see any documentation that suggested it should be accepted?

With the `audit3` case, `audit3` is implicitly in the xforms namespace. That should probably be allowed for backwards compatibility.

If I'm following correctly, the implementation so far allows for taking namespace into account when matching binds to nodes but it doesn't specify which nodes are allowed to be in which namespaces.

I was surprised to see the namespace map being stored in a `TreeElement` rather than at the `FormDef` level which seems it would be more analogous with what [`FormInstance` does](https://github.com/opendatakit/javarosa/blob/8313caa731acd7bb51c0fac48bbc1aa572ef3897/src/org/javarosa/core/model/instance/FormInstance.java#L52). Is that to avoid parsing the file more than once or am I missing something else?

dcbriccetti:
Hi @lognaturel. I did some guessing in this first implementation. Why don’t we decide exactly what should be allowed and what should not, encode examples in tests, and I’ll make things work.

I created a unit test for TreeElementNameComparator, which should make this easier than making lots of forms. Here’s the output:

```
comparing a with a
Match
comparing a in http://openrosa.org/xforms orx with orx:b
No match
comparing a in http://openrosa.org/xforms orx with orx:a
Match
comparing a with orx:a
No match
comparing a in http://openrosa.org/xforms orx with orx2:b
No match
comparing a with b
No match
comparing a with *
Match
```

>I was surprised to see the namespace map being stored in a TreeElement

Yes, this is clearly the wrong place for it, but I wasn’t able to quickly do it properly without changing public interfaces.

lognaturel:
Love the test and the results are as I would expect -- the namespaces must match for an overall match.

Makes sense about the namespace map location. Overall, looks like all this is going in the right direction!

I was asking about nesting but really, I realize it's probably not important for this case. Moving forward, nodes or attributes introduced in the OpenRosa Xforms namespace will be unambiguously allowed only in that namespace. But unfortunately, there wasn't agreement on that namespace when `meta` was introduced so I think pretty much anything goes for `meta` and `audit` at this point.

dcbriccetti:
Glad this is looking good to you. What more do you see we need to do here?

lognaturel:
I think basically it's getting that map to a more expected home. That should be enough for supporting currently-implemented features (the audit).

dcbriccetti:
OK, I’ll move the map. Are we willing to change public interfaces if necessary?

lognaturel:
I haven't looked at the code deeply enough to know what kind of change would be needed. It wouldn't be just additive?

dcbriccetti:
There are many paths to the code in question, and if the prefix to namespace association data isn’t present in the TreeElement, it’ll need to be passed in via all those callers. I’ll keep looking at it.

dcbriccetti:
I’m not finding a low-impact alternative to what I have done in this branch. I’ll create a pull request from another, experimental branch where I started changes to pass the map in where needed. There are many calls, in many layers, involving public methods, and interfaces. It might be best to go with what we have, and plan to improve things in a later refactoring.

dcbriccetti:
Hi @lognaturel! 
What do you thinking about going with this change as is, and improving it later, after reducing the public interface and further refactoring.

PR #69, opened 2017-06-10
Fixed LeftCurly

grzesiek2010:


src/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -78,7 +78,9 @@ public void processAction(FormDef model, TreeReference contextRef) {
 		}
 
 		AbstractTreeElement node = context.resolveReference(qualifiedReference);
-		if(node == null) { throw new NullPointerException("Target of TreeReference " + qualifiedReference.toString(true) +" could not be resolved!"); }
+		if(node == null) {
```
dcbriccetti: If you’re improving formatting, how about a space after if?

src/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -116,8 +116,12 @@ public void setOriginalContext(TreeReference ref) {
 	}
 
 	public TreeReference getOriginalContext() {
-		if (this.original == null) { return this.contextNode; }
-		else { return this.original; }
+		if (this.original == null) {
```
dcbriccetti: Is this the indentation amount we want?

src/org/javarosa/core/model/actions/SetValueAction.java
grzesiek2010: I've fixed only LeftCurly checkstyle. A space is another one and I can fix it in the future as a separate pr.

src/org/javarosa/core/model/condition/EvaluationContext.java
grzesiek2010: As above it's only `LeftCurly` checkstyle. I don't care about anything else here.

dcbriccetti: Oh, I see. Thanks.

lognaturel:
@dcbriccetti Looks like the current open pull requests all require longer discussion that probably won't conclude in the short term. Would now be an ok time for @grzesiek2010 to add a few this check and a few others so that we have fewer style inconsistencies to work around?

dcbriccetti:
>Would now be an ok time

Sure.

PR #70, opened 2017-06-13
Prepare snapshot

yanokwa:


yanokwa:
When you publish snapshots, it timestamps them on the Sonatype side and so fresh pulls of the library will get the latest snapshot.

PR #72, opened 2017-06-15
Properly check for NaN in XPathFuncExpr.round, and add a test for it.

dcbriccetti:
See discussion in issue #71


lognaturel:
Please remove the couple of no-longer needed imports and we can merge!

Issue #150, opened 2017-07-13
Enable redirecting stderr and stdout usage without VM-wide effects


batkinson:
#### Software versions 
JavaRosa 2.2.0

Due to the library's direct dependence on `System.out` and `System.err`, and indirect dependence on `System.err` when calling `printStackTrace()` on a `Throwable` object, programs that use the javarosa library have no recourse to reclaim the console streams from its content without redirecting *all* content for the VM. 

Ideally, in the future it would be nicer to replace print calls with appropriate logging. However, that is likely to be more challenging technically considering the contexts the library is used in. In the mean time, by providing a simple method for *redirecting only the library's print calls* and *not the VM's* clients can improve performance and quality of their diagnostics and still leverage the library.

batkinson:
This was closed in the PR #151, but commits appeared to be squashed so it wasn't auto-closed.

PR #151, opened 2017-07-13
Javarosa 150

batkinson:
Enhancement enabling client redirection of output sent to stderr and stdout. See #150 for more information.

### DIFF
--- a/src/org/javarosa/core/io/Std.java
+++ b/src/org/javarosa/core/io/Std.java
@@ -25,4 +25,8 @@ public class Std {
     public static void setErr(PrintStream err) {
         Std.err = err;
     }
+
+    public static void printStack(Throwable t) {
+        t.printStackTrace(Std.err);
+    }
 }
--- a/src/org/javarosa/core/io/Std.java
+++ b/src/org/javarosa/core/io/Std.java
@@ -1,0 +1,28 @@
+package org.javarosa.core.io;
+
+import java.io.PrintStream;
+
+/**
+ * Used to break dependency directly on System.out and System.err.
+ */
+public class Std {
+
+    public static PrintStream out = System.out;
+    public static PrintStream err = System.err;
+
+    /**
+     * Use this to replace stdout in case anything more than assignment is necessary in the future.
+     * @param out
+     */
+    public static void setOut(PrintStream out) {
+        Std.out = out;
+    }
+
+    /**
+     * Use this to replace stderr in case anything more than assignment is necessary in the future.
+     * @param err
+     */
+    public static void setErr(PrintStream err) {
+        Std.err = err;
+    }
+}
--- a/src/org/javarosa/core/io/StreamsUtil.java
+++ b/src/org/javarosa/core/io/StreamsUtil.java
@@ -131,12 +131,9 @@ public class StreamsUtil {
 		}
 
 		if (len > 0 && read < len) {
-			// System.out.println("WARNING: expected " + len + "!!");
 			throw new RuntimeException("expected: " + len + " bytes but read "
 					+ read);
 		}
-		// replyS
-		// System.out.println(new String(data, "UTF-8"));
 
 		return data;
 	}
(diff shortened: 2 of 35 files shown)


batkinson:
Not sure what's up with the CI build, but it looks to be a problem with that environment or something.

yanokwa:
Looks like a rebuild fixed it.

src/org/javarosa/core/io/Std.java
```diff
@@ -0,0 +1,32 @@
+package org.javarosa.core.io;
+
+import java.io.PrintStream;
+
+/**
+ * Used to break dependency directly on System.out and System.err.
+ */
+public class Std {
+
+    public static PrintStream out = System.out;
```
lognaturel: Would it be better to always force use of the setters by making the class vars private?

batkinson: > Do you have an approach for a quick sanity check?
So, after implementing a similar one myself, I realized that the console output in briefcase is split between a modest amount output from javarosa and similar, but far more verbose amount from similar `System.out` code in briefcase that prints warnings pulled from the jr API. Yours test is probably working, but just not easy to see with all the other console output.

batkinson: > Would it be better to always force use of the setters by making the class vars private?
It would be, but this it would prevent the similar usage pattern to System.out and err (they're public static final, but you set via native setOut/Err methods). We can fully encapsulate, but to do that the API will look like this:

```
Std.out().println()
```

To completely mimick the Java runtime classes we'd likely have to go native or modify the JRE. I figured this would probably just be a stop-gap until a proper logging rework can be done. If the method access isn't too uncomfortable, I can fully-encapsulate.

lognaturel: Ok, I'm fine with this being temporary and I agree with the argument for `Std.out` instead of `Std.out()`. @dcbriccetti can you please take a quick look -- don't want to make any changes that will surprise you since I know you're looking across the codebase.

I'll file an issue to rework logging.

lognaturel:
> I realized that the console output in briefcase is split between a modest amount output from javarosa and similar, but far more verbose amount from similar System.out code in briefcase that prints warnings pulled from the jr API. Yours test is probably working, but just not easy to see with all the other console output.

Sigh, I see it now. That's annoying. So I guess we can take this conversation back on the Briefcase side of things. I'd like to replace those printlns with `Log.info` calls but I'm having some trouble wrapping my head around how logging is even supposed to work. I wrote a note about that [here](https://github.com/opendatakit/briefcase/pull/143#pullrequestreview-49898184).

src/org/javarosa/core/io/Std.java
dcbriccetti: This overlaps files I have in pending changes in two branches, but that’s life.

Why not change this to use proper logging now, to avoid having to change all the calls again.

lognaturel: @batkinson Since it turns out JR printing isn't what's causing the bulk of the problems on the Briefcase side, maybe it isn't as much of a priority? @dcbriccetti would you like to "do it right"? I filed issue #152.

lognaturel:
> Why not change this to use proper logging now, to avoid having to change all the calls again.

@batkinson Since it turns out JR printing isn't what's causing the bulk of the problems on the Briefcase side, maybe it isn't as much of a priority? @dcbriccetti would you like to "do it right"? I filed issue #152.

batkinson:
>@batkinson Since it turns out JR printing isn't what's causing the bulk of the problems on the Briefcase side, maybe it isn't as much of a priority? @dcbriccetti would you like to "do it right"?

Yeah, it's not as high a priority, just mildly annoying you can't completely cleanup the console and control this. Also, I think I have a slightly different view on this from "avoid having to change the calls again". Now that you can use static analysis to find all these (the biggest pains being the stack dumps), the job is actually easier with this integrated.

The one exception to that is the external interface. That one you probably want to be something stable since you can't easily change the *clients*. The internal API usage is easy enough to change without much effort.

dcbriccetti:
I have no objection to merging this change now, and doing proper logging later, if there’s benefit in it now.

lognaturel:
Thanks both for input. Let's do this now, put in logging and then put `Std` on a deprecation path some time down the line.

PR #153, opened 2017-07-17
Issue 60: Allow http://openrosa.org/xforms namespace for meta

mdudzinski:
This work is based on #68 but it uses `namespacePrefix` field instead of `namespacePrefixesByURI`.

### DIFF
--- a/src/org/javarosa/xform/parse/XFormParser.java
+++ b/src/org/javarosa/xform/parse/XFormParser.java
@@ -334,15 +334,10 @@ public class XFormParser implements IXFormParserFunctions {
     private Map<String, String> buildNamespacesMap() {
         final Map<String, String> namespacePrefixesByURI = new HashMap<>();
 
-        for (int i = 0; i < _xmldoc.getChildCount(); i++) {
-            Object firstChild = _xmldoc.getChild(0);
-            if (firstChild instanceof Element) {
-                Element el = (Element) firstChild;
-                if (el.getName().equals("html")) {
-                    for (int j = 0; j < el.getNamespaceCount(); j++) {
-                        namespacePrefixesByURI.put(el.getNamespaceUri(j), el.getNamespacePrefix(j));
-                    }
-                }
+        Element el = _xmldoc.getRootElement();
+        if (el.getName().equals("html")) {
+            for (int i = 0; i < el.getNamespaceCount(); i++) {
+                namespacePrefixesByURI.put(el.getNamespaceUri(i), el.getNamespacePrefix(i));
             }
         }
 
--- a/src/org/javarosa/xform/parse/XFormParser.java
+++ b/src/org/javarosa/xform/parse/XFormParser.java
@@ -328,7 +328,7 @@ public class XFormParser implements IXFormParserFunctions {
         return _f;
     }
 
-    /** Extracts the namespaces from the element, if present, and creates a map of URI to prefix */
+    /** Extracts the namespaces from the given element and creates a map of URI to prefix */
     private static Map<String, String> buildNamespacesMap(Element el) {
         final Map<String, String> namespacePrefixesByURI = new HashMap<>();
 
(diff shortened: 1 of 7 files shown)


src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -330,15 +328,12 @@ public FormDef parse() throws IOException {
         return _f;
     }
 
-    /** Extracts the namespaces from the html element, if present, and creates a map of URI to prefix */
-    private Map<String, String> buildNamespacesMap() {
+    /** Extracts the namespaces from the element, if present, and creates a map of URI to prefix */
```
dcbriccetti: The “if present, ” in the comment no longer applies.

mdudzinski: @dcbriccetti that's right, thanks! I'll fix it.

test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -49,6 +77,50 @@ public void parsesMetaNamespaceForm() throws IOException {
         assertEquals("Number of error messages", 0, parseResult.errorMessages.size());
     }
 
+    @Test
+    public void serializeAndRestore_MetaNamespaceFormInstance() throws IOException {
```
dcbriccetti: Did you intend to use an “_”?

mdudzinski: Yes because I like the idea of having underscore in tests names because it's easier to read them and those test methods are never called by hand.
However all existing tests use camelCase so perhaps we shouldn't mix the conventions. 
At least in a single test class. I'll change this. 
Thanks!

lognaturel:
Thank you both very much @dcbriccetti and @mdudzinski! 

@dcbriccetti and I did a live review together in #javarosa-code. The only thing I was left wondering about was whether `serializeAndRestoreMetaNamespaceFormInstance` should additionally show that the namespace in the form def and the namespace in the filled out instance match. If you think that would be a helpful extra check, it can be added in a separate PR. 

And now that I think about it, I have an additional question about the test. I noticed that XML generated for the filled out instance has the `orx2` prefix defined on the `data` node when I did `xml.writeExternal(new DataOutputStream(System.out));`. As @mdudzinski commented before, Collect filled out instances define the prefix on the `meta` node. What accounts for that difference?

mdudzinski:
@lognaturel @dcbriccetti Very good point! I somehow thought namespaces are being exercised under the hood but they are not. I added missing assertions in #155.

Yes the serialization of the namespaces is a bit tricky. In kxml2, the prefixes -> URI map is kept in `org.kxml2.kdom.Element.prefixes` field. There can be multiple namespaces definitions (prefix->URI maps), each defined at different tree level just like in an xml file. When `KXmlSerializer` is serializing a `Document` instance and finds an `Element` instance that has an URI in a `namespace` field that hasn't been defined in the element's its ancestors `prefixes` field, then serializer provides own prefix value for this namespace and puts the definition (in xml)  in the element's tag.

Calling `loadNamespaces(_xmldoc.getRootElement(), fi);` in `private void parseDoc(Map<String, String> namespacePrefixesByUri)` loads the namespaces defined in the form root element (html tag) into the form instance root element (data tag). This ensure that prefixes are consistent across restored form instance and the form definition but in the form instance xml file, namespaces definition is put in the root tag.

So for the test case example, without the above code, the result of the form instance serialization was:
n0-namespace-prefix
with the above code:
orx2-namespace-prefix

There is one drawback though. This code will add all namespaces definitions to the form instance, even those not used by the instance. This can be fixed by filtering the prefixes map before attaching  it to the form instance. Let me know if that is a concern, I can implement this.

In general, the whole problem is caused by kxml2 implementation that doesn't handle namespace prefixes :(

I added the missing assertions and some comments for the above situation in #155.

PR #154, opened 2017-07-20
Provide the initial implementation of external secondary instances, Issue 16

dcbriccetti:
…ssue 16.


dcbriccetti:
@lognaturel, what’s the next action here?

lognaturel:
This is a complete implementation of CommCare JavaRosa's "slow" approach to external secondary instances, is that right?

Clayton wrote 
> If you are looking for stress tests I would also try to test using external instances with attributes, and context switches (IE : instance('a')/data/item[@id = instance('b')/data/id]/output to make sure those work correctly, as that's where the faults are likely to start showing themselves

Perhaps that's something to add here? Or maybe it's best left to another PR, I'm not sure. This should have no effect on any existing behavior, right?

I think now might be a good time to get @mdudzinski involved for a thorough review.

dcbriccetti:
>This is a complete implementation of CommCare JavaRosa's "slow" approach to external secondary instances, is that right?

Yes, at least as far as the functionality the parsesExternalSecondaryInstanceForm test exercises goes.

All existing tests still pass, and this should not alter existing functionality. Yes, a review from @mdudzinski would be great.

dcbriccetti:
>Perhaps that's something to add here? Or maybe it's best left to another PR, I'm not sure.

Because this change is so large, I’d like to see it merged to master even if there is more work to do on the feature.

mdudzinski:
@lognaturel @dcbriccetti I'll review this.

mdudzinski:
@dcbriccetti There is something odd with the commits `Manually merge “serialize and restore test improvement” from @mdudzinski`  is doubled but [the second commit](https://github.com/opendatakit/javarosa/pull/154/commits/d93811a06f2a3934b2ff487adf28aa2014b72cac) introduces changes that are not mine. I believe this should be fixed before merging so we have a clean commit tree.

src/org/javarosa/core/reference/ReferenceManager.java
```diff
@@ -49,7 +49,7 @@ private ReferenceManager() {
 	 * @return Singleton accessor to the global
 	 * ReferenceManager.
 	 */
-	public static ReferenceManager _() {
+	public static ReferenceManager instance() {
```
mdudzinski: Yeay finally :) '_' might not being supported in the future Java release warning was really annoying

src/org/javarosa/xml/util/ActionableInvalidStructureException.java
```diff
@@ -0,0 +1,37 @@
...
+import org.javarosa.core.util.NoLocalizedTextException;
+
+/**
+ * Invalid structure error that can _potentially_ be recovered from via
+ * advanced user intervention. Useful for notifying the user that the issue lies
+ * on the server.
+ *
+ * @author Phillip Mates (pmates@dimagi.com).
+ */
+public class ActionableInvalidStructureException extends InvalidStructureException {
```
mdudzinski: This class isn't used in the JavaRosa codebase at all. I guess it's used in dimagi's client but looks like we don't need it. At least yet?

src/org/javarosa/xml/util/UnfullfilledRequirementsException.java
```diff
@@ -0,0 +1,82 @@
+package org.javarosa.xml.util;
+
+/**
+ * @author ctsims
+ */
+public class UnfullfilledRequirementsException extends Exception {
```
mdudzinski: This exception doesn't seem to be thrown anywhere. Only caught in `ExternalDataInstance`'s contructor and declared in `ElementParser#parse()` method signature.

src/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -0,0 +1,50 @@
...
+        super(instanceId);
+        setName(instanceId);
+        try {
+            URI uri = new URI(srcLocation);
+            if ("jr".equals(uri.getScheme()) && "file".equals(uri.getHost())) {
+                String absolutePath = /* ToDo: find out how to get the actual location */
+                        System.getProperty("user.dir") + "/resources" + uri.getPath();
+                KXmlParser xmlParser = ElementParser.instantiateParser(new FileInputStream(absolutePath));
+                TreeElementParser treeElementParser = new TreeElementParser(xmlParser, 0, instanceId);
+                root = treeElementParser.parse();
```
mdudzinski: I/O operations in an object's constructor looks super strange to me. 
@dcbriccetti shouldn't we think on refactoring this? now or in the nearest future? Or we don't bother for now as more work from dimagi's is going to be merged?

dcbriccetti:
@mdudzinski Thanks for all the notes.

>There is something odd with the commits Manually merge “serialize and restore test improvement” from @mdudzinski is doubled but the second commit introduces changes that are not mine. I believe this should be fixed before merging so we have a clean commit tree.

In the process of merging your changes into the branch, I manually reapplied my changes to the files. Since you changed that file, I reapplied my changes to it. What can we do to make this better?

src/org/javarosa/xml/util/UnfullfilledRequirementsException.java
dcbriccetti: You’re right. Thanks.

src/org/javarosa/core/model/instance/ExternalDataInstance.java
dcbriccetti: At one point I had my autographed copy of Josh Bloch’s Effective Java Second Edition practically memorized. Then I left java for over 8 years (for Scala). Help me remember why this is a bad practice?

mdudzinski:
@dcbriccetti I think the message of the second commit should be different, something like "re-applying changes lost during the merge" to avoid confusion. But honestly I prefer re-basing as it produces much cleaner commit tree. Instead of merging somebody's work branch into your own, you just do a rebase against the master branch after somebody's work branch has been merged. After you do this, you need to update your branch with a force push. 
It works well only if you are the only one who works on your branch.

src/org/javarosa/core/model/instance/ExternalDataInstance.java
mdudzinski: @dcbriccetti I think it's a bad practice in a lot of languages. The biggest problem is that it may be hard to unit test this class in isolation. It is implicitly dependent on `TreeElementParser` so you cannot use a stubbed one. It also hardcodes the path so you cannot easily change it for testing or when using JavaRosa in a client app. 

I think it also breaks the single responsibility principle where a class that stands for a tree structure is also responsible for I/O, xml parsing and handling errors that come from this operations.

Another thing: I/O calls may be considered to have side-effects in similar way network calls does (communication with outside world and errors) although they are far less likely. However, consider running this code on an Android device when user ejects an sd-card. Or on an invalid external instance xml.

dcbriccetti: @mdudzinski, thanks for pointing out where this needs improvement. I was considering it a work in progress. As you saw there are several // ToDos in there, that I thought we’d resolve as we got a little further along. I suggest we make your recommended changes later, perhaps driven by testing and further development.

dcbriccetti:
>I think the message of the second commit should be different...

I agree with what you say. However, I’ve spent hours keeping this branch up to date and have a lot of other things to do. I don’t know how to go back and change a commit comment, but I have made a new branch from before that comment, and re-made the changes from that point. I can certainly understand that you don’t want your name tagged on a commit that you had nothing to do with. Can we let the rest of it slide this time?

dcbriccetti:
Replaced by https://github.com/opendatakit/javarosa/pull/157

src/org/javarosa/core/model/instance/ExternalDataInstance.java
mdudzinski: @dcbriccetti That's fine. It's an initial implementation after all:) just wanted to highlight the code that needs improving so we won't lose/forget about it as we move forward. Thanks!

mdudzinski:
@dcbriccetti No worries. It's not that I don't want my name tagged on this commit because it doesn't contain my work. It's just because I always try to pay attention to have clean commit tree. It makes dive in the history easier if you need to for some reasons and two commits with same message are always confusing. Like looking for when a bug, feature or unexpected behavior was introduced.

mdudzinski:
@lognaturel I did some reading and I'm impressed with what can be done with external instances. I haven't realized this. This code is a good starting point for further improvements. Not sure what is planned to be next but changing the magic csv support into an external instance approach seems reasonable.

Issue #60, opened 2017-06-02
Allow http://openrosa.org/xforms namespace for meta


lognaturel:
As shown in [ODK XForms spec examples](https://opendatakit.github.io/xforms-spec/), tags not in the W3C XForms spec should be in the `http://openrosa.org/xforms namespace`, commonly abbreviated to `orx`. In particular, `meta` and all children should really be `orx:meta`, etc.

Collect works fine with that but when I upload a form using the namespace to Aggregate, I get errors. For example, [audit-test.txt](https://github.com/opendatakit/javarosa/files/1048460/audit-test.txt) can be uploaded to Aggregate with no problems. However, when I try to upload [audit-test-orx.txt](https://github.com/opendatakit/javarosa/files/1048469/audit-test-orx.txt) I get the following error regarding `data/meta/instanceID` and `data/meta/audit`.
```
Form uploaded with warnings. There are value fields in the form that do not have <bind/> 
declarations or those <bind/> declarations do not have a type attribute that identifies the data type
of that field (e.g., boolean, int, decimal, date, dateTime, time, string, select1, select, barcode, 
geopoint or binary).
```

I haven't looked into the code at all but I suspect that there's a JavaRosa issue here. My best guess is that JavaRosa basically ignores the namespace in the instance block and then gets confused when there is one on the bind. The ideal would be for JR to accept either no namespace or `orx` for those tags. We might even consider providing a warning if the namespace isn't there.

dcbriccetti:
Validate gives more info:

```
Parsing form...
Title: "Audit test ORX"
XForm Parse Warning: <bind> defined for a node that doesn't exist [/data/orx:meta/orx:instanceID]. The node's name was probably changed and the bind should be updated. 
XForm Parse Warning: <bind> defined for a node that doesn't exist [/data/orx:meta/orx:audit]. The node's name was probably changed and the bind should be updated. 
```

dcbriccetti:
I simplified the input to this:
```
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms">
  <h:head>
    <h:title>Audit test ORX</h:title>
    <model>
      <instance>
        <data id="audit-test-orx">
[...]

PR #155, opened 2017-07-20
#60 serialize and restore test improvement

mdudzinski:
Added missing assertion reported by @lognaturel under #153.
Added a comment that explains namespaces serialization.


lognaturel:
Thank you!!

PR #157, opened 2017-07-28
External secondary instances 3

dcbriccetti:
Replaces https://github.com/opendatakit/javarosa/pull/154


mdudzinski:
Code is okay but I'm wondering how ed820c0 got into this PR :)

dcbriccetti:
>Code looks good but I'm wondering how ed820c0 got into this PR :)

I’ll try the rebase approach next time. I cherry picked a commit to make the branch current.

dcbriccetti:
Hi @lognaturel. Is there anything else to do before merging this?

lognaturel:
Just to make sure the status here is clear, we discussed in #javarosa-code and @dcbriccetti will build a jar and do one last sanity check with Collect using this version of the code so we can all sleep better at night.

dcbriccetti:
I’m glad we are sane and do sanity checks, because I missed something. Collect uses ReferenceManager’s ```_``` method, which I had renamed to ```instance```. I just added a deprecated “alias” ```_``` that calls instance, so no changes will be needed in Collect. *Now* we can sleep better at night.

lognaturel:
Thanks for doing that extra check and catching that issue.

@yanokwa how about doing a snapshot release that we can get in Collect after the 1.9.1 release? That will allow us to exercise these changes some more before they go out into the wild.

yanokwa:
2.5.0 snapshot is out!

dcbriccetti:
How great to get this large change in!

PR #160, opened 2017-07-31
Remove init() method from DateData

alxndrsn:


lognaturel:
@alxndrsn I believe this is a change just to improve readability but it would be good to get a confirmation. Could you come up with any conceivable reason as to why this `init` structure was used previously?

alxndrsn:
Yeah, it's to make the intention of the code clearer.

I can't think of a specific reason for using `init()` like this, although both `DateData.d` and `DateData.init` both being non-private does leave some fun opportunities for dangerous code.  Perhaps someone once took advantage of this and needed the `init()` method to fix things up?

It does rely on `DateUtils.roundDate()` _not_ mutating the date instance, which is currently the case and probably a reasonable assumption going forwards.

lognaturel:
Sorry about letting this sit! I agree it's an improvement but would like @dcbriccetti to also review since he's most deeply immersed in JavaRosa these days.

lognaturel:
(immersed like a fun day of swimming at the ocean, right??)

dcbriccetti:
I looked at it briefly some weeks back and couldn’t satisfy myself that the risk was low. The benefit doesn’t seem very high. I do agree with the change though, if it doesn’t break anything.

lognaturel:
Yeah, that's the way I'm feeling too. I don't see any totally obvious ramifications but I'm nervous when I don't understand why code was written in an unusual way. @alxndrsn can you think of a way to give us higher confidence that it's totally safe and/or a reason we're not seeing why this is important to do?

alxndrsn:
> can you think of a way to give us higher confidence that it's totally safe and/or a reason we're not seeing why this is important to do?

_If_ `init` (the boolean, rather than the method) is only accessed internally, and _if_ `d` is only accessed through `getValue()` or other methods rather than directly, this change is safe.

There may be many ways the current implementation could be used(/misused/abused) to do weird things.  This change is explicitly to prevent that.

I see two general approaches with code like this:

1. never change the weird/wrong stuff in case someone relies on something as if it were a feature
2. change the implementation to be easy to reason about, use and test, and bump the major version of the next release so people aren't surprised by breaking changes

If the plan for this lib leans towards **1**, then it might be helpful to add some project-level documentation stating similar, and in this specific case some class- or field-level documentation saying why we want to preserve the current implementation.

lognaturel:
Sounds like this will be a good candidate to go in with a series of such improvements and more thorough documentation. @dcbriccetti is working on documenting where the integration points are with core ODK tools so at least that's clear and then reducing the public interfaces as much as possible. I agree a version bump will make sense when that's done so let's group this change with that effort.

PR #161, opened 2017-07-31
Whitespace cleanup

alxndrsn:


dcbriccetti:
I see much more than whitespace changes. Is that intended?

alxndrsn:
I agree!

Issue #165, opened 2017-08-04
Use form or client language in month/day short text


lognaturel:
opendatakit/xforms-spec#122

If the form language definition uses the format described at XLSForm/xlsform.github.io/issues#73, try to use the language code for localization. If that fails or isn't available, use the client locale.

PR #166, opened 2017-08-05
Internationalize month and day values in DateUtils.format.

dcbriccetti:
This should get us partway on https://github.com/opendatakit/javarosa/issues/165, don’t you think, @lognaturel and @alxndrsn?

### DIFF
--- a/src/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/org/javarosa/core/model/utils/DateUtils.java
@@ -25,7 +25,9 @@ import java.util.ArrayList;
 import org.javarosa.core.services.locale.Localization;
 import org.javarosa.core.util.MathUtils;
 
+import org.joda.time.DateTime;
 import org.joda.time.LocalDateTime;
+import org.joda.time.format.DateTimeFormat;
 
 /**
  * Static utility methods for Dates in j2me
@@ -270,8 +272,7 @@ public class DateUtils {
 				} else if (c == 'n') {	//numeric month
 					sb.append(f.month);
 				} else if (c == 'b') {	//short text month
-					String[] months = new String[] {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
-					sb.append(months[f.month - 1]);
+                    sb.append(DateTimeFormat.forPattern("MMM").print(new DateTime(DateUtils.getDate(f))));
 				} else if (c == 'd') {	//0-padded day of month
 					sb.append(intPad(f.day, 2));
 				} else if (c == 'e') {	//day of month
@@ -287,8 +288,7 @@ public class DateUtils {
 				} else if (c == '3') {	//0-padded millisecond ticks (000-999)
 					sb.append(intPad(f.secTicks, 3));
 				} else if (c == 'a') {	//Three letter short text day
-					String[] dayNames = new String[] {"Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"};
-					sb.append(dayNames[f.dow - 1]);
+                    sb.append(DateTimeFormat.forPattern("EEE").print(new DateTime(DateUtils.getDate(f))));
 				} else if (c == 'Z' || c == 'A' || c == 'B') {
 					throw new RuntimeException("unsupported escape in date format string [%" + c + "]");
 				} else {
--- a/test/org/javarosa/core/model/utils/test/DateUtilsTests.java
+++ b/test/org/javarosa/core/model/utils/test/DateUtilsTests.java
@@ -32,311 +32,310 @@ import java.util.SimpleTimeZone;
 import java.util.TimeZone;
 
 public class DateUtilsTests extends TestCase {
-	
-	private static int NUM_TESTS = 8;
-	
-	Date currentTime;
-	Date minusOneHour;
-	
-	public DateUtilsTests(String name){
-		super(name);
-		System.out.println("Running " + this.getClass().getName() + " test: " + name + "...");
-	}
-	
-	public void setUp() throws Exception{
-		super.setUp();
-		
-		currentTime = new Date();
-		minusOneHour = new Date(new Date().getTime() - (1000*60));
-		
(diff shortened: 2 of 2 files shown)


lognaturel:
Yep! And it seems the locale should be correctly updated when Collect's user interface menu is used, right? If you could attach a quick test form that would be a nice additional sanity check. Looks like the only thing remaining is to try to use the form language if possible. This is using [this XLSForm feature](http://xlsform.org/#language) which gets converted to [this](https://opendatakit.github.io/xforms-spec/#languages) and should now follow [this format](https://github.com/XLSForm/xlsform.github.io/issues/73).

src/org/javarosa/core/model/utils/DateUtils.java
```diff
@@ -270,8 +272,7 @@ public static String format (DateFields f, String format) {
 				} else if (c == 'n') {	//numeric month
 					sb.append(f.month);
 				} else if (c == 'b') {	//short text month
-					String[] months = new String[] {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
-					sb.append(months[f.month - 1]);
+                    sb.append(DateTimeFormat.forPattern("MMM").print(new DateTime(DateUtils.getDate(f))));
```
alxndrsn: Something weird going on with the indentation here.

alxndrsn: I've just skimmed the JodaTime docs, and it refers to the "default timezone" a lot.  Does this mean the OS timezone, rather than defaulting to e.g. `UTC` for everyone?

test/org/javarosa/core/model/utils/test/DateUtilsTests.java
```diff
@@ -331,6 +334,40 @@ private void testCycle(Date in) {
 		}
 
 	}
+
+	public void testFormatting() {
```
alxndrsn: There's some weird indentation here too.

test/org/javarosa/core/model/utils/test/DateUtilsTests.java
```diff
@@ -331,6 +334,40 @@ private void testCycle(Date in) {
...
+                this.january = january;
+                this.sunday = sunday;
+            }
+
+            private String language;
+	        private String january;
+	        private String sunday;
+        }
+
+        LangJanSun langJanSuns[] = new LangJanSun[] {
```
alxndrsn: This would seem like a natural fit for an `enum`.

dcbriccetti: Tell me more? It’s an array of test data.

alxndrsn: e.g.
```
public void testFormatting() {
	enum LangJanSun {
		en("Jan", "Sun"),
		es("ene", "dom"),
		fr("janv.", "dim.");
		
		String january;
		String sunday;
		
		private LangJanSun(String language, String january, String sunday) {
			this.language = language;
			this.january = january;
			this.sunday = sunday;
		}
	}

	Locale savedLocale = Locale.getDefault();

	for (LangJanSun ljs : LangJanSun.values()) {
```

alxndrsn: I don't think this is particularly important, but it seemed like a relevant comment.

Personally I'd favour something like:
```
testFormatTranslations() {
	testFormatTranslation("en", "Sun Jan");
	testFormatTranslation("es", "dom ene");
	testFormatTranslation("fr", "dim. janv.");
}

testFormatTranslation(String language, String expectedJanSun) {
	Locale savedLocale = Locale.getDefault();

	
	Locale.setDefault(Locale.forLanguageTag(language));

	assertEquals(expectedJanSun, DateUtils.format(SUN_JAN_1_2017, "%a %b"));


	Locale.setDefault(savedLocale);
}
```

dcbriccetti: From the Java Tutorial:

> Enum Types
> 
> An enum type is a special data type that enables for a variable to be a set of predefined constants. The variable must be equal to one of the values that have been predefined for it. Common examples include compass directions (values of NORTH, SOUTH, EAST, and WEST) and the days of the week.

This is different. It’s a collection of test data, not an enumeration of all possible values of something.

alxndrsn: All the possible values of the class are predefined and constant so I'd say it follows the letter of the docs.  Either way, apologies if I caused offence.

dcbriccetti: >Personally I'd favour something like...
I like that approach, too.

src/org/javarosa/core/model/utils/DateUtils.java
dcbriccetti: That’s a tabs vs spaces problem. I’m reluctant to convert all the tabs to spaces because it would obscure the history (e.g., IDEA’s Show Annotation feature).

It means the timezone from the default locale, which I probably comes from the OS.

test/org/javarosa/core/model/utils/test/DateUtilsTests.java
dcbriccetti: I did convert all tabs to spaces in this file (perhaps before I thought of the history obscuration I mentioned above).

dcbriccetti:
>Looks like the only thing remaining is to try to use the form language if possible.

@lognaturel how about we merge this one and I’ll create another if we can set a locale based on the form language?

src/org/javarosa/core/model/utils/DateUtils.java
alxndrsn: Yeah, I'd be inclined to just be consistent with the rest of the file and use tabs.

lognaturel:
Verified with [format-date.xml.txt](https://github.com/opendatakit/javarosa/files/1258708/format-date.xml.txt). The app-wide language change works and the form illustrates the form-level language next step.

PR #177, opened 2017-08-24
Remove executable flag from source files

alxndrsn:
All these files were executable, but probably shouldn't be.


yanokwa:
@dcbriccetti Do you know if this will interfere with other PRs?

PR #178, opened 2017-08-29
Codestyle

grzesiek2010:

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -78,7 +78,7 @@
              value="WhitespaceAround: ''{0}'' is not preceded with whitespace."/>
         </module> -->
         <module name="OneStatementPerLine"/>
-        <!-- <module name="MultipleVariableDeclarations"/> -->
+        <module name="MultipleVariableDeclarations"/>
         <!-- <module name="ArrayTypeStyle"/> -->
         <!-- <module name="MissingSwitchDefault"/> -->
         <!-- <module name="FallThrough"/> -->
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -111,11 +111,11 @@
             <message key="name.invalidPattern"
              value="Member name ''{0}'' must match pattern ''{1}''."/>
         </module> -->
-        <!-- <module name="ParameterName">
+        <module name="ParameterName">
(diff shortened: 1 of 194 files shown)


lognaturel:
All of these changes are straight from the Google styleguide that we now follow across ODK so should be non-controversial. @grzesiek2010 Can you please confirm that you were able to let the IDE help you with all of these? If there were any that you had to do manually, we should know and pay extra attention because errors are more likely to have snuck in.

We'll pull the final 2.5.0 release on Sunday and then can merge this unless @dcbriccetti vetoes! I think we should also do a white space normalization pass before resuming development so we don't have to fight with inconsistent spacing anymore.

dcbriccetti:
No objection.

grzesiek2010:
@lognaturel ooops intellij offers to help with code improvements indeed but I was focused on those I can cover using standard checkstyle checks to avoid the same in the future (we can always click analyze the code -> ...). I thought you know it :smiley:

so yes everything is made by me and all I can confirm is the fact I was careful :smiley:

When it comes to intellij, there is no appropriate checkstyle check in most cases to solve problems this tool found, maybe we could find something in findbugs or pmd but I was focused on checkstyle only.

lognaturel:
I'm sorry, I don't quite follow! For `NeedBraces`, for example, did you run Analyze > Run Inspection By Name..., enter "Control flow without braces" and then click "Add braces to statement?" It sounds like no, you did it by running lint and then changing things by hand? It's true that there may not be an exact match for every `checkstyle` check but in my experience most are there. I think leaning on the IDE where possible when making this kind of change not only gives greater confidence but also makes it much faster. For example, it's possible to imagine adding a curly brace in the wrong place and changing the semantics.

lognaturel:
All that said, I'll review this as it is once the 2.5.0 release stabilizes. In particular, I'd like to see if #180 can be addressed (or confirmed to not be important).

lognaturel:
Thanks for doing this, @grzesiek2010. @dcbriccetti and I talked it over a bit in Slack and unfortunately we're just finding it too high risk to merge. With so many different kinds of changes in one PR and without the assurance that using a tool provides, there are too many subtle things that could go wrong.

The big one that was constantly causing trouble was whitespace so I sent in #185 to address that specifically. I think on this code base it would be fine to stick to making stylistic updates as they are needed in changesets such as #184.

PR #185, opened 2017-09-14
Tabs to spaces.

lognaturel:
Performed a replace on a tab character to "    " within IntelliJ. Ran `./gradlew test` to make sure those still run. Ran checkstyle to make sure all were converted.

I could have used something like `sed` to do the conversion but the approaches seemed equivalent.

This was done with tool assistance so I don't think there is risk. Also, using @dcbriccetti neat trick, [this](https://github.com/opendatakit/javarosa/pull/185/commits/d94861d576c84983b632e17e0842546b66db038f?w=1) shows all changes are whitespace changes.


lognaturel:
@dcbriccetti Oh, yes, forgot that multiple spaces get collapsed in HTML. I did mean 4 spaces.

Correct, this doesn't fix indentation, it just converts tabs to spaces.

PR #186, opened 2017-09-14
Fix ser/deserialization of forms with external data instances

dcbriccetti:
Fix ser/deserialization of forms with external data instances. Make some of @mdudzinski’s recommended changes from https://github.com/opendatakit/javarosa/pull/154#pullrequestreview-51806826.

I have tests for (de)serializing two types of forms with secondary instances.

It’s working more like commcare-core

Since there are no users of the feature yet, and we aren’t done, I would say no.


lognaturel:
What does `// ToDo find out how to get the actual location` refer to? It's the location of the external instance, right?

dcbriccetti:
Yes, where in the file system to find the file.

lognaturel:
👍 Other "stuff outside the form" uses `<odk-form-folder>/<form-name>-media/` as the root dir but in this case all external secondary instances will go in the same folder. It would be good for that folder to reside in the odk root folder. We can come back to it.

PR #189, opened 2017-09-15
Parse time using current date

grzesiek2010:
I've tested saving/reading form with time question

It's not perfect but we decide to use this simple solution which should work in most cases.

I think it's safe

### DIFF
--- a/src/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/org/javarosa/core/model/utils/DateUtils.java
@@ -328,7 +328,9 @@ public class DateUtils {
     }
 
     public static Date parseTime (String str) {
-        DateFields fields = new DateFields();
+        DateFields fields = getFields(new Date());
+        fields.second = 0;
+        fields.secTicks = 0;
         if (!parseTime(str, fields)) {
             return null;
         }


grzesiek2010:
It might look a bit strange because I kept this https://github.com/opendatakit/javarosa/pull/189/files#diff-18a7ab04391f14cd701fddea1a7b790bR336 
I need current date only during parsing here (a few lines above): https://github.com/opendatakit/javarosa/pull/189/files#diff-18a7ab04391f14cd701fddea1a7b790bR332
so I decided not to remove those lines and keep default 01-01-1970 after parsing.

dcbriccetti:
Sorry, I’ve lost the context for this. Is there an issue? Is this the save makes off by one hour problem?

lognaturel:
Yes, the intent is to interpret the time with the current date rather than with Jan 1, 1970 so that time doesn't change within a single timezone.

@grzesiek2010 Could you please generate an apk with this change so that @mmarciniak90 can verify all the different cases and confirm it's an improvement (save time in summer time, open form later but still in summer time; save time in summer time, open form in winter time, etc)?

grzesiek2010:
> @grzesiek2010 Could you please generate an apk with this change so that @mmarciniak90 can verify all the different cases and confirm it's an improvement (save time in summer time, open form later but still in summer time; save time in summer time, open form in winter time, etc)?

Of course. Here you go @mmarciniak90:
[time_test_apks.zip](https://github.com/opendatakit/javarosa/files/1332119/time_test_apks.zip)

f you have any questions let me know.

mmarciniak90:
I noticed unexpected behavior. @grzesiek2010 could you investigate this problem,
Use case:
1. User uses device with GMT+00.00 time zone
2. User adds answer to Time widget. eg. 08.00
3. User saved form and leaves it
4. User changes time zone to GMT+01.00
5. User opens saved form
6. User sees list of all question with 09.00 in Time widget
screenshot_20171006-090705
7. User opens Time Widget and sees 08.00
screenshot_20171006-090710
8. User clicks on 'Select time' button and sees 07.00
screenshot_20171006-090715

grzesiek2010:
@mmarciniak90 
It's was a bit tricky and hard to track down what's wrong but it's not related to my changes. 
The issue you noticed is already reported here: https://github.com/opendatakit/collect/issues/1309 
The same issue is visible if you use Collect 1.10.2. The reason why you are not able to notice the same issue using current master branch is because 
https://github.com/opendatakit/collect/pull/1492 has been merged recently and looks like it fixes this bug.

I prepared for you an apk with my fix using current master branch so it contains https://github.com/opendatakit/collect/pull/1492 and it should be ok.
[fix.zip](https://github.com/opendatakit/javarosa/files/1355447/fix.zip)

Please test everything again and look at https://github.com/opendatakit/collect/issues/1309 because as I said looks like we can close that issue.

mmarciniak90:
@grzesiek2010 You are right, everything is OK on 'collect-FIX2.apk'

Tested with success on Android 4.1, 4.2, 4.4, 5.1, 6.0, 7.0
Time is not changed when user saved form and opens it to edit.
Problem mentioned in previous comment also is not visible on new apk.

hooverlunch:
Great work you all! My heart was in my throat therefore second when reading this thread, But it looks like all is well. I will try to take a look at this later today. Thank you so much.

hooverlunch:
IT WORKS!!!!

Here is a marching band for you:

image

hooverlunch:
Aww no gif support.

lognaturel:
Yaaaaay, so happy we could at least make things better.

Next step is to document the current limitation clearly. opendatakit/docs#283

Issue #170, opened 2017-08-17
Evaluate performance of current external secondary instance implementation


lognaturel:
In #16, initial support for external secondary instances was added. This provides:
- A single way to query external data. [Data preloading](https://opendatakit.org/help/form-design/data-preloading/) exists in Collect to load external CSVs. However, it doesn't allow for querying with sophisticated expressions as requested [here](https://forum.opendatakit.org/t/pulldata-with-two-lookup-values/9435)
- A more XForms-like specification for external data that other members of the ecosystem such as Enketo can adopt
- The potential for "artificial instances" as described [in the initial OpenRosa spec](https://bitbucket.org/javarosa/javarosa/wiki/externalinstances) to query application data such as instances. This could enable things like [this use case](https://forum.opendatakit.org/t/a-dynamic-form-data-lookup-locally-without-server-synchronization/9382)

And perhaps more? Clearly it has some nice potential.

Before the current implementation can be used in any meaningful way, we need to know how performant it is. I suspect that it struggles once the external instances get big at all. We should try to get a sense of what the practical limits are. I'm not sure what the best way to do this is but as a start I might script a way to generate random XML data, generate files of different lengths, build a form that queries that data (e.g. gets the first element where child1 is "foo" and child2 is "bar"), try it out on a physical device and make a table to see how the performance degrades as more records are added (100, 1000, 10000, etc) and maybe as queries get more complex.

lognaturel:
@mdudzinski perhaps you could look into this? @dcbriccetti you probably have some good ideas as well.

lognaturel:
Had a chance to catch up with @mdudzinski and he will try to look at it early Sept and keep us updated here.

dcbriccetti:
I’m doing some experiments with large secondary instance files.

yanokwa:
Here is a form from our friends in Nigeria with a cascading list of states, wards, and LGAS. 

[nigeria-wards.xml.zip](https://github.com/opendatakit/javarosa/files/1332282/nigeria-wards.xml.zip)
[nigeria-wards.xlsx](https://github.com/opendatakit/javarosa/files/1332283/nigeria-wards.xlsx)

mdudzinski:
@dcbriccetti I've seen on Slack that you figured out a few bottlenecks and came up with a few ideas for improvements. Do you have some code already pushed to your fork? If so, could you point mi to the branch or a commit? I'd like to catch up and I find it hard to do it by only following the conversation on slack :)
[...]

PR #190, opened 2017-09-24
Time parsing of forms with external secondary instances

dcbriccetti:
Part of https://github.com/opendatakit/javarosa/issues/170

Write a “test” that times parsing of forms with increasingly larger external secondary instance files.

Closes #

### DIFF
--- a/resources/external-secondary-instance-large.xml
+++ b/resources/external-secondary-instance-large.xml
@@ -1,0 +1,20 @@
+<h:html xmlns="http://www.w3.org/2002/xforms"
+        xmlns:h="http://www.w3.org/1999/xhtml">
+
+    <h:head>
+        <h:title>Form with external secondary instance</h:title>
+        <model>
+            <instance>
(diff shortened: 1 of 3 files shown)


lognaturel:
@mdudzinski you'll want to take a look at this.

I think that this basically shows that loading the whole instance to memory doesn't work for more than a few thousand records even on a powerful device. Does that match your conclusions, @dcbriccetti?

dcbriccetti:
I’m working on a test to see if the results are the same for large “internal” secondary instances. They probably are. I’m surprised that the performance is polynomial and not linear. I’m inclined to investigate that. Is that worth some time? 

So far I’m testing on my Mid-2014 MacBook Pro with a SSD. I plan to test on an Android device too.

lognaturel:
That is puzzling. Investigation seems worth it to me! It could be that there are some relatively easy optimizations to make even without doing things like looking at introducing a database layer (which I think is inevitable).

PR #192, opened 2017-09-27
Improve parsing performance

dcbriccetti:
Part of issue https://github.com/opendatakit/javarosa/issues/170

Needs testing.

@mdudzinski, please make your testing improvements based on this branch. 

Closes #

All of the new or changed code has been tested except for parts of TreeElementChildrenList. @mdudzinski, you can run the tests with coverage (you’re using IDEA?) to see the spots.

There may be better solutions, but this change makes huge performance improvements.

With JavaRosa it’s hard to be extremely confident about changes of this magnitude.


src/org/javarosa/xform/parse/XmlTextConsolidator.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Determines whether to use the simpler, much slower, child deletion method, or the more complicated
+     * one, which performs well even with tens of thousands of deletions or children.
+     *
+     * @param numDeletions the number of deletions to be done
+     * @param numChildren  the number of children
+     * @return whether to use the high volume approach
+     */
+    private static boolean highVolumeApproachNeeded(int numDeletions, int numChildren) {
```
mdudzinski: Why do we want to support both child deletion approaches? It seems like we can keep just `rebuildChildrenAndAttributes` if it's *always* faster than `removeDeletedChildren`. On the other hand, we may want to keep it as an explanation why we need the complex `rebuildChildrenAndAttributes`.

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -388,55 +387,7 @@ public static Document getXMLDocument(Reader reader, CacheTable<String> stringCa
...
-                    e.addChild(Element.TEXT, stringCache.intern(accumulate));
-                }
-            }
-            for(int i = e.getChildCount() - 1; i >= 0 ; i-- ){
-                if(toRemove[i]) {
-                    e.removeChild(i);
-                }
-            }
-        }
+        XmlTextConsolidator.consolidateText(stringCache, doc.getRootElement());
```
mdudzinski: @dcbriccetti I've been wondering why do we need `stringCache` and `InterningKXmlParser` but I've never had a chance/time to investigate this properly. 
Do you know what is this for and under what condition is it used?
I'm just hoping that this may be coming from "the dark J2ME" days and we could get rid of this at some point :)

src/org/javarosa/xform/parse/XmlTextConsolidator.java
dcbriccetti: The new approach performs better only for forms with many children.

src/org/javarosa/xform/parse/XFormParser.java
dcbriccetti: I wondered the same, but didn’t look into it.

mdudzinski:
@dcbriccetti I'm working on improving test coverage of the parsing and I'm wondering if I should base it on your branch. The improvements you did are significant. Are you planning to do much more? I noticed that the CircleCI build failed but it seems to be a random fail. Or tests for the large instance parsing take too much memory :)

If you're through for now with the JR optimizations then I could go ahead and continue digging into your changes and work on the tests.

lognaturel:
I think the next step is for @mdudzinski and I to do a careful review, merge when that's satisfied and then integrating into Collect so devs can start using the changes. @dcbriccetti let us know when you're ready for that.

I think @mdudzinski then strengthening the test suite so we can have even more confidence in the changes before the next release is a fantastic idea.

test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -134,7 +134,7 @@ private void timeParsing(LargeInstanceFileGenerator lfg, Path largeDataFilename,
         NumberFormat nf = NumberFormat.getNumberInstance();
         List<String> results = new ArrayList<>(); // Collect and display at end
         results.add("Children\tSeconds");
-        for (double powerOfTen = 3; powerOfTen <= 6.0; powerOfTen += 0.1) {
+        for (double powerOfTen = 3; powerOfTen <= 4.0; powerOfTen += 0.1) {  // Raise this upper limit to really measure
```
mdudzinski: Shouldn't we document rough memory requirements? Seems like we know something.

mdudzinski:
@dcbriccetti @lognaturel  I did some investigations and `stringCache`, `InterningKXmlParser` and related code is not used internally by JavaRosa [nor any other odk project](https://github.com/search?utf8=%E2%9C%93&q=org%3Aopendatakit+XFormParserFactory&type=Code) but in theory it can by used by a third-party client. 
There are two "entry points" that calls this code:
* `XFormParser#getXMLDocument(Reader reader, CacheTable<String> stringCache)`
* `XFormParserFactory#XFormParserFactory(CacheTable<String> stringCache)`

And of course `InterningKXmlParser` itself. 
According to authors comment, it was added to:

> (...) to provide a custom string cache to be used during parsing, which should be helpful
> in conserving memories in environments where there might be
> multiple parsed forms in memory at the same time.

I know we are reluctant to change the public API so I am not not sure if we should continue discussing getting rid of this code along with the optimization work or leave it for now and get back to this in the future.
On the other hand, the less paths/branches in the code then it is easier to reach 100% code coverage. And maintain it.

[Here ](https://github.com/mdudzinski/javarosa/commit/8dbad97929addcddf2d62ea01337388c9744909f) you can find a commit deleting this code based on this branch.

dcbriccetti:
I’d love to pull out the code, as you did in your branch above. But @yanokwa and @lognaturel should decide if they accept the risk of this breaking things for someone with memory– and CPU-limited hardware.

test/org/javarosa/xform/parse/XFormParserTest.java
mdudzinski: @dcbriccetti Yeah but not 4 GB I guess. So it's seems like JVM on CircleCI container could've pre-allocated too much and the process was killed. This would mean it's matter of CI config and we don't need this change. I guess. Anyway, I doesn't seem to be super important right now.

dcbriccetti: This makes me think that that performance measurement code doesn’t belong in the tests. It certainly doesn’t need to run each time. Maybe I’ll find another place for it.

lognaturel: Agreed they don't need to run every time.

mdudzinski:
@lognaturel Yes that works for me! Thanks

PR #194, opened 2017-10-02
Use StringBuilder instead of StringBuffer

alxndrsn:
Foloow-up to https://github.com/opendatakit/javarosa/pull/193.

`StringBuffer` is synchronized, so for performance reasons `StringBuilder` should be
used in preference unless thread safety is required.
See https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuffer.html

Any external code previously using `PrefixTreeNode.render(StringBuffer)` will
need updating to use `PrefixTreeNode.render(StringBuilder)`.

# Risks

If any external code exists using `PrefixTreeNode.render(StringBuffer)`, and it actually depends on the thread-safety of the `StringBuffer`, this cannot be merged.  That seems unlikely.

### DIFF
--- a/src/org/javarosa/core/util/PrefixTreeNode.java
+++ b/src/org/javarosa/core/util/PrefixTreeNode.java
@@ -73,10 +73,19 @@ public class PrefixTreeNode {
     }
 
     public String render() {
-        StringBuffer temp = new StringBuffer();
+        StringBuilder temp = new StringBuilder();
         return render(temp);
     }
 
+    public String render(StringBuilder sb) {
+        if (parent != null) {
+            parent.render(sb);
+        }
+        sb.append(this.prefix);
+        return sb.toString();
+    }
+
+    @Deprecated
     public String render(StringBuffer buffer) {
         if(parent != null){
             parent.render(buffer);


dcbriccetti:
Unless we’ve demonstrated through measurement that the existing code is causing poor performance, I’m not convinced it’s worth changing the public API. Anyone calling render now would have to change their code, right?

mdudzinski:
@dcbriccetti @alxndrsn Why not create a new public method that uses `StringBuilder` and deprecate the old one?

alxndrsn:
@mdudzinski obviously a better approach, thanks!

PR updated 🙂

lognaturel:
Go team! 🎉

dcbriccetti:
I don’t like this solution since it duplicates code, and for extremely little benefit. I recommend:

1. Undo the change
2. See if you can change the type of the formal parameter to `AbstractStringBuilder`.

src/org/javarosa/core/util/PrefixTreeNode.java
```diff
@@ -73,10 +73,19 @@ public String toString () {
     }
 
     public String render() {
-        StringBuffer temp = new StringBuffer();
+        StringBuilder temp = new StringBuilder();
         return render(temp);
     }
 
+    public String render(StringBuilder sb) {
+        if (parent != null) {
```
dcbriccetti: This duplicates code. I think this change should be undone, and that we should leave this code alone.

PR #195, opened 2017-10-06
Implement round with a single argument (the second defaulting to 0).

### LINKED ISSUE #173: Make second parameter for round optional

is provided

dcbriccetti:
Closes #173 

I added a new test.

It’s simple and effective.
No.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -200,7 +200,7 @@ public class XPathFuncExpr extends XPathExpression {
                 places = 0;
             } else {
                 assertArgsCount(name, args, 2);
-                places = toNumeric(argVals[1]).intValue();
+                places = toNumeric(argVals[1]).intValue(); // non-standard Excel-style round(value,decimal place)
             }
             return round(toNumeric(argVals[0]), places);
         } else if (name.equals("string")) {
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -194,9 +194,15 @@ public class XPathFuncExpr extends XPathExpression {
         } else if (name.equals("int")) { //non-standard
             assertArgsCount(name, args, 1);
             return toInt(argVals[0]);
-        } else if (name.equals("round")) { // non-standard Excel-style round(value,decimal place)
-            assertArgsCount(name, args, 2);
-            return round(toNumeric(argVals[0]), toNumeric(argVals[1]).intValue());
+        } else if (name.equals("round")) {
+            final int places;
+            if (args.length == 1) {
+                places = 0;
+            } else {
+                assertArgsCount(name, args, 2);
+                places = toNumeric(argVals[1]).intValue();
+            }
+            return round(toNumeric(argVals[0]), places);
         } else if (name.equals("string")) {
             assertArgsCount(name, args, 1);
             return toString(argVals[0]);
--- a/test/org/javarosa/xpath/test/XPathEvalTest.java
+++ b/test/org/javarosa/xpath/test/XPathEvalTest.java
@@ -318,6 +318,10 @@ public class XPathEvalTest extends TestCase {
 
         logTestCategory("math functions");
         testEval("abs(-3.5)", 3.5);
+        // round, with a single argument
+        testEval("round('14.29123456789')", 14.0);
+        testEval("round('14.6')",           15.0);
+        // round, with a two arguments
         testEval("round('14.29123456789', 0)", 14.0);
         testEval("round('14.29123456789', 1)", 14.3);
         testEval("round('14.29123456789', 1.5)", 14.3);


xmlwerks:
This is the first time I've looked at this repo, but, seems you made quick work of this issue. The only thing I'm curious about is that I don't see the function used in any of the XForms tests, but perhaps they are more about de/serializing.

dcbriccetti:
Hi Hank. You saw that I added round with a single argument to 
`test/org/javarosa/xpath/test/XPathEvalTest.java`, right?

xmlwerks:
Yes, I saw the XPath tests, compiled and ran them (not that the CI doesn't but...) So, maybe a separate question, but not sure what the XForms tests do. I was looking to see the functions used in forms, a kind of integrated test. I'm not saying that is necessary.

dcbriccetti:
XFormParserTest is to test XFormParser, especially for some new features we have added since April.

lognaturel:
I think unit tests are sufficient for this one!

But, of course, nothing is ever totally simple. From https://www.w3.org/TR/xpath/#function-round:
```
The round function returns the number that is closest to the argument and that is an integer. If there are two such numbers, then the one that is closest to positive infinity is returned. If the argument is NaN, then NaN is returned. If the argument is positive infinity, then positive infinity is returned. If the argument is negative infinity, then negative infinity is returned. If the argument is positive zero, then positive zero is returned. If the argument is negative zero, then negative zero is returned. If the argument is less than zero, but greater than or equal to -0.5, then negative zero is returned.
```

First question is whether the results should be integers? Or we just leave the `.0`?

To be super explicit about the cases mentioned, I added
```
        testEval("round('14.5')", 15.0);
        testEval("round('NaN')", NaN);
        testEval("round('-NaN')", -NaN);
        testEval("round('0')", 0.0);
        testEval("round('-0')", -0.0);
        testEval("round('-0.5')", -0.0);
```
The first one fails and I think the rounding strategy should be changed.

That's easy enough but then I don't know what to do with the 2-param variant. Should it also round up to match? But that would change existing forms' behavior!

dcbriccetti:
Oh boy.

xmlwerks:
Sigh.  The ODK specification does state that it deviates from XPath 1.0 because the second (precision) parameter is optional: https://opendatakit.github.io/xforms-spec/#xpath-functions . However the issue reported was that an exception:

> unable to deploy: org.javarosa.xpath.XPathUnhandledException[The problem was located in calculate expression for ${calculation} XPath evaluation: cannot handle function ‘round’ requires 2 arguments. Only 1 provided.]

The problem is that the parameter was NOT optional -- that's clear from the diff. 

It seems better to support the 3.0 specification as far as reasonable (don't attempt to support exponential notation). The 3.0 specification does provide a behavior for the function with and without the precision. It's a bit more complicated because the precision can be negative:   https://www.w3.org/TR/xpath-functions-30/#func-round . One difference between round() and int() would be that it does not truncate to Java's Long.MAX_VALUE or Long.MIN_VALUE . Before anyone starts coding that, what say ye?

tiritea:
Presumably 3.0 was designed to be backward compat (mostly?) with 1.0, which is basically the crux of the problem I hit [my custom ODK client only supports XPath 1.0  presently...]. So my vote would be go with the 3.0 spec (and thereby knockoff another ODK 'deviation'...). That way, if something is still broken, we can then all blame W3C instead! :-)

xmlwerks:
Yes I consulted Stackoverflow. To follow the XPath spec, and pass the tests, we have to do something like the following:

`int method = (number < 0.0) ? BigDecimal.ROUND_HALF_DOWN : BigDecimal.ROUND_HALF_UP;
                return (new BigDecimal
                        (Double.toString(number))
                        .setScale(numDecimals, method))
                        .doubleValue();`

lognaturel:
Thanks for reasoning through this @xmlwerks @tiritea! 

JavaRosa and the ODK XForms spec predate XPath 3.0. My initial sense is that the 2-parameter function should maintain its current behavior and that changes to existing rounding behavior if deemed necessary should probably happen in a major version change. Seems it would be confusing for that behavior to change all of a sudden.

Since the 1-parameter function will be entirely new, it should follow XPath 1.0. But of course it's also confusing to have different behavior between the 1-parameter function and the 2-parameter function.

It seems there are two less than ideal options: change the behavior of the 2-parameter function or have different strategies for xxx.5 between the two `round` variants.

Does anyone have a strong feeling about this?

lognaturel:
@xmlwerks It looks like your #201 takes the first approach and changes the 2-parameter function, right?

Really, that's the better way to go for the long term. I'm just not sure how we communicate the change in the short term. It probably wouldn't affect a lot of people but it would be strange to have submissions one week doing 1.15 -> 1.1 and the next week doing 1.15 -> 1.2.

iowerx:
Hi @lognaturel . Sorry for confusion caused by my account switch. Yes, PR 201 supersedes a PR I created under @xmlwerks account. It adds another test for comma separated decimals, and a bit of a hack to handle them (string replace). I agree this represents a change in functionality so best to deploy with a release and accompanying notes. In the mean time, the documentation needs to be updated since the second parameter is not optional: https://opendatakit.github.io/xforms-spec/#xpath-functions . In the mean time, If a release branch is created, this could be merged there, or @tiritea can use it to get his forms working again. Cheers.

tiritea:
> ...Does anyone have a strong feeling about this?

Not strong, but my feeling is that there doesn't seem to be compelling reason for ODK to deviate from the well-defined XPath 1.0 or XPath 3.0 standards when it comes to rounding numbers in numeric XPath expressions [sic].

> ...one week doing 1.15 -> 1.1 and the next week doing 1.15 -> 1.2.

The fact nobody has previously complained about 1.15 being rounded *down* to 1.1, is probably indicative that nobody cared... till now. :-)

lognaturel:
> my feeling is that there doesn't seem to be compelling reason for ODK to deviate from the well-defined XPath 1.0 or XPath 3.0 standards
Agreed. And I know you've brought up some other deviations and mistakes in the documentation that I haven't had a chance to respond to yet. I think those can directly be filed as issues or better yet submitted as pull requests. I'm guessing any deviation is an accident of history more than a deliberate choice.

> The fact nobody has previously complained about 1.15 being rounded down to 1.1, is probably indicative that nobody cared... till now. :-)
Good point, that does seem really unusual.

So let's close this and if anyone has a chance to review #201 that would be much appreciated.

iowerx:
I hope that #201 does indeed follow the XPath 3.0 standard in terms of rounding. The documentation goes to lengths to explain that it is "counter-inuitive":
https://www.w3.org/TR/xpath-functions-30/#func-round
The 1.0 spec is less precise regarding halves, but negative numbers round towards zero.
It the same method as Apache Commons Math:
http://commons.apache.org/proper/commons-math/index.html
In any case, it's been fun, thanks for the feedback.

PR #196, opened 2017-10-09
Revert "Use StringBuilder instead of StringBuffer"

lognaturel:
Reverts opendatakit/javarosa#194 following [this discussion](https://opendatakit.slack.com/archives/C389J6X0W/p1507085720000248).

The commits don't look right but that's ok because they'll be squashed. The diff does look right.

### DIFF
--- a/src/org/javarosa/core/util/PrefixTreeNode.java
+++ b/src/org/javarosa/core/util/PrefixTreeNode.java
@@ -73,19 +73,10 @@ public class PrefixTreeNode {
     }
 
     public String render() {
-        StringBuilder temp = new StringBuilder();
+        StringBuffer temp = new StringBuffer();
         return render(temp);
     }
 
-    public String render(StringBuilder sb) {
-        if (parent != null) {
-            parent.render(sb);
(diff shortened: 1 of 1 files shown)


src/org/javarosa/core/util/PrefixTreeNode.java
```diff
@@ -73,19 +73,10 @@ public String toString () {
     }
 
     public String render() {
-        StringBuilder temp = new StringBuilder();
+        StringBuffer temp = new StringBuffer();
```
dcbriccetti: This could stay, but either way is fine with me.

lognaturel: Wouldn't that require adding the new `render` variant? Or do you mean using a `StringBuilder` in the zero-argument variant and never calling `render(StringBuffer buffer)`?

dcbriccetti: Oh, you’re right of course. Disregard, please. I like your change as is.

PR #198, opened 2017-10-10
Deprecate getXMLDocument(Reader reader, CacheTable<String> stringCache).

dcbriccetti:
Closes #

All tests still pass.

This is safe. We considered completely removing the code.

No


lognaturel:
For posterity: this was almost certainly added just for CommCare. No ODK tool has ever used it. More discussion at https://github.com/opendatakit/javarosa/pull/192#issuecomment-334178240

PR #200, opened 2017-10-12
Issue #173 - Make second parameter for round optional.

### LINKED ISSUE #173: Make second parameter for round optional

xmlwerks:
Closes #173

Passes existing and new test cases, per 3.0 specification
I hope this is the best possible, but other solutions were not.
Possible I18N decimal separator incompatibility, since it no longer forces separator to "."


xmlwerks:
@dcbriccetti After created the PR, I had some doubts about decimal separator handling. But it passed the existing test cases and those requested by @lognaturel Anyway, this was merely supposed to be an easy intro to contributing... Accept, reject, reuse, it's up to you.

PR #201, opened 2017-10-17
Issue #173 - Make second parameter for round optional.

### LINKED ISSUE #173: Make second parameter for round optional

iowerx:
Closes #173 

Passes old and new unit tests.
Regarding locale specific number formats, yes, however,  
it's complicated to handle comma separated numbers via Locale class, so string methods used.
To handle the XPath 3.0 specification, only ROUND_HALF_XXX was considered because
it specifies that negative numbers round toward zero, i.e., up.
It needs to be tested in French and other locales with comma decimal separators.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -44,13 +44,11 @@ import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.math.BigDecimal;
-import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.regex.Pattern;
 
 import static java.lang.Double.NaN;
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -194,9 +194,16 @@ public class XPathFuncExpr extends XPathExpression {
         } else if (name.equals("int")) { //non-standard
             assertArgsCount(name, args, 1);
             return toInt(argVals[0]);
-        } else if (name.equals("round")) { // non-standard Excel-style round(value,decimal place)
-            assertArgsCount(name, args, 2);
-            return round(toNumeric(argVals[0]), toNumeric(argVals[1]).intValue());
+        } else if (name.equals("round")) { // Proximate XPath 3.0 and Excel-style round(value,decimal place)
+            final int places;
+            if (args.length == 1) {
+                places = 0;
+             } else {
+                assertArgsCount(name, args, 2);
+                places = toNumeric(argVals[1]).intValue();
+            }
+            // Some locales use ',' instead of '.' as a decimal separator, but not supported by Decimal class.
+            return round(toNumeric(argVals[0].toString().replace(',','.')), places);
         } else if (name.equals("string")) {
             assertArgsCount(name, args, 1);
             return toString(argVals[0]);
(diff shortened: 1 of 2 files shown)


src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -194,9 +194,16 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
         } else if (name.equals("int")) { //non-standard
             assertArgsCount(name, args, 1);
             return toInt(argVals[0]);
-        } else if (name.equals("round")) { // non-standard Excel-style round(value,decimal place)
-            assertArgsCount(name, args, 2);
-            return round(toNumeric(argVals[0]), toNumeric(argVals[1]).intValue());
+        } else if (name.equals("round")) { // Proximate XPath 3.0 and Excel-style round(value,decimal place)
```
iowerx: Removed the assertion requiring two arguments. Sans a second argument, decimal places will be zero (0). Nevertheless does assert that there cannot be more than two (2) arguments. Handles numbers with comma (,) separated decimal place through string method replace.

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1045,14 +1052,22 @@ private static Double round(double number, int numDecimals) {
...
+        // Rounding with positive decimals
         if (numDecimals >= 0) {
-            // Some locales use ',' instead of '.' as a decimal separator. Since Double.valueOf
-            // doesn't handle comma as the decimal separator, we must ensure a dot will be used by
-            // forcing the US format locale.
-            final NumberFormat nf = NumberFormat.getNumberInstance(Locale.US);
-            nf.setMaximumFractionDigits(numDecimals);
-            nf.setGroupingUsed(false);
-            return Double.valueOf(nf.format(number));
+            try {
```
iowerx: Per the XPath 3.0 specification, rounds up, rounds half up, and rounds negative numbers to zero (mathematically, also up, I suppose).

test/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -331,9 +335,20 @@ public void doTests () {
         testEval("round('12345.12345', -2)", 12300.0);
         testEval("round('12350.12345', -2)", 12400.0);
         testEval("round('12345.12345', -3)", 12000.0);
+        // XPath specification tests
         testEval("round('1 div 0', 0)", NaN);
-        // Todo: get round('12345.15', 1) and round('-12345.15', 1) working on Java 8
```
iowerx: AFAIK, this test is contrary to the XPath 3.0 specification, which rounds up, see above.

iowerx:
Hi @dcbriccetti and @lognaturel et al. I canceled the PR under xmlwerks because millennials laugh at xml, and I will be contributing under this account. The one thing I have not tested is running under locales with comma decimal separators. Use, refuse, repurpose as you will. Cheers.

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -194,9 +194,16 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
...
+        } else if (name.equals("round")) { // Proximate XPath 3.0 and Excel-style round(value,decimal place)
+            final int places;
+            if (args.length == 1) {
+                places = 0;
+             } else {
+                assertArgsCount(name, args, 2);
+                places = toNumeric(argVals[1]).intValue();
+            }
+            // Some locales use ',' instead of '.' as a decimal separator, but not supported by Decimal class.
+            return round(toNumeric(argVals[0].toString().replace(',','.')), places);
```
dcbriccetti: Did you consider doing the replace inside toNumeric?

test/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -318,6 +318,10 @@ public void doTests () {
 
         logTestCategory("math functions");
         testEval("abs(-3.5)", 3.5);
+        // round, with a single argument
+        testEval("round('14.29123456789')", 14.0);
+        testEval("round('14.6')",           15.0);
+        // round, with a two arguments
```
dcbriccetti: “a two” does not compute

iowerx: Live or Die by your editor. Thanks.

lognaturel:
Thanks so much for taking this on, @iowerx! It's great to be just that much more standards-compliant.

If you haven't found your way there yet, ODK devs hang out at http://slack.opendatakit.org and the whole community is at https://forum.opendatakit.org/.

src/org/javarosa/xpath/expr/XPathFuncExpr.java
dcbriccetti: This code has a bug. It bypasses the unpack that toNumeric does.

iowerx: I missed your comment. Thanks for the catch.

Issue #202, opened 2017-10-19
Cannot parse 'eIMCI by D-Tree' form and a form with template repeat


mdudzinski:
#### Software versions 
JavaRosa v2.6.0-SNAPSHOT, Mac OS 10.12.6

#### Problem description
It is no longer possible to parse some of the forms after performance improvements (#192). Example of forms long with unit tests are available [here]
(https://github.com/mdudzinski/javarosa/commits/form-parsing-code-coverage) (the latest 2 commits)

* eIMCI by D-Tree

> java.lang.NullPointerException
> 	at org.javarosa.xform.parse.XFormParser.allChildrenHaveTheSameName(XFormParser.java:1869)
> 	at org.javarosa.xform.parse.XFormParser.buildInstanceStructure(XFormParser.java:1836)
> 	at org.javarosa.xform.parse.XFormParser.buildInstanceStructure(XFormParser.java:1839)
> 	at org.javarosa.xform.parse.FormInstanceParser.parseInstance(FormInstanceParser.java:60)
> 	at org.javarosa.xform.parse.XFormParser.parseDoc(XFormParser.java:445)
> 	at org.javarosa.xform.parse.XFormParser.parse(XFormParser.java:326)
> 	at org.javarosa.xform.parse.XFormParserTest.parse(XFormParserTest.java:302)
> 	at org.javarosa.xform.parse.XFormParserTest.parseIMCIbyDTreeForm(XFormParserTest.java:279)

* simple form with template repeat

> java.lang.IllegalStateException: Unexpected multiplicity mismatch
> 
> 	at org.javarosa.core.model.condition.EvaluationContext.expandReferenceAccumulator(EvaluationContext.java:266)
> 	at org.javarosa.core.model.condition.EvaluationContext.expandReference(EvaluationContext.java:217)
> 	at org.javarosa.xform.parse.FormInstanceParser.flagRepeatables(FormInstanceParser.java:114)
> 	at org.javarosa.xform.parse.FormInstanceParser.processRepeats(FormInstanceParser.java:105)
> 	at org.javarosa.xform.parse.FormInstanceParser.parseInstance(FormInstanceParser.java:81)
> 	at org.javarosa.xform.parse.XFormParser.parseDoc(XFormParser.java:445)
> 	at org.javarosa.xform.parse.XFormParser.parse(XFormParser.java:326)
> 	at org.javarosa.xform.parse.XFormParserTest.parse(XFormParserTest.java:302)
[...]

PR #203, opened 2017-10-19
Solve the NPE in Issue #202

dcbriccetti:
Cherry pick mdudzinski’s test forms and new tests for Issue #202. (I’m surprised git is attributing the changes to me. Maybe I should have done a merge or rebase.)
Fix the NPE by changing allChildrenHaveTheSameName to allChildrenAreElementsAndHaveTheSameName, and add several tests for it.

Shall we keep these two forms as part of our test suite?

Several new tests all pass, and the form loads in Collect.

It fixes the problem with no impact to other parts of the code. See https://github.com/opendatakit/javarosa/issues/202 for discussion.

“No”

### DIFF
--- a/resources/eIMCI-by-D-Tree.xml
+++ b/resources/eIMCI-by-D-Tree.xml
@@ -1,0 +1,4895 @@
+
+ <!--
+      eIMCI
+      Created by Dimagi based on eIMCI from Mtwara study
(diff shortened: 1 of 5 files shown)


dcbriccetti:
Ready for review.

lognaturel:
Should we use the full eIMCI form or a small one that just illustrates the comment issue?

PR #204, opened 2017-10-20
Same as Solve the NPE in Issue #202, but rebased

dcbriccetti:
I rebased on @mdudzinski’s changes to undo that the cherry picking I did before, in https://github.com/opendatakit/javarosa/pull/203, put my name on his changes.

### DIFF
--- a/test/org/javarosa/xform/parse/ChildProcessingTest.java
+++ b/test/org/javarosa/xform/parse/ChildProcessingTest.java
@@ -29,7 +29,7 @@ public class ChildProcessingTest {
         assertFalse(XFormParser.allChildrenAreElementsAndHaveTheSameName(el));
     }
 
-    @Test public void WorksWithNoChildren() {
+    @Test public void worksWithNoChildren() {
         Element el = new Element();
         assertFalse(XFormParser.allChildrenAreElementsAndHaveTheSameName(el));
     }


test/org/javarosa/xform/parse/ChildProcessingTest.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Test public void worksWithTwoNotMatchingChildren() {
+        Element el = new Element();
+        el.addChild(ELEMENT, el.createElement(null, "Child Name 1"));
+        el.addChild(ELEMENT, el.createElement(null, "Child Name 2"));
+        assertFalse(XFormParser.allChildrenAreElementsAndHaveTheSameName(el));
+    }
+
+    @Test public void WorksWithNoChildren() {
```
mdudzinski: Should start with lowercase.

lognaturel:
#203 has been merged but I can also merge this one to get the right authorship in there (though @dcbriccetti will appear to have previously made the same changes @mdudzinski made).

Good catch on the case issue, @mdudzinski! Could you please fix that real quick, @dcbriccetti?

dcbriccetti:
Hmm, why a conflict? @lognaturel, can you force that in or whatever the needful is?

mdudzinski:
@dcbriccetti @lognaturel  Shouldn't such typos be caught by checkstyle?

lognaturel:
@mdudzinski Yes but that rule must not be turned on yet. My guess is that the code base is pretty inconsistent so it may take some work to turn on. I think there's a way to turn on checks only for commits after a certain point in time and perhaps we should look into that.

Sorry, I'm not sure what to do with this! I can fix the merge conflict and then merge with a merge commit but that would make the history really strange.

dcbriccetti:
OK, let’s abandon this. I’ll fix the WorksWithNoChildren in the next PR in this area.

lognaturel:
🙏

dcbriccetti:
A slight load off.

lognaturel:
Every little bit counts.

PR #208, opened 2017-10-27
XForm parsing code coverage

mdudzinski:
I'm still working on some more tests for the parsing codebase but you can review what's so far.

Nothing. Those are only tests

N/A
Nope

### DIFF
--- a/resources/form-with-setvalue-action.xml
+++ b/resources/form-with-setvalue-action.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml">
+    <h:head>
+        <h:title>SetValue action</h:title>
+        <model>
+            <instance>
+                <data id="setvalue-action">
+                    <text/>
+                </data>
+            </instance>
+            <bind id="text-data-binding" nodeset="/data/text" type="string"/>
+            <setvalue bind="text-data-binding" event="xforms-ready" >Test Value</setvalue>
+        </model>
+    </h:head>
+    <h:body />
+</h:html>
--- a/test/org/javarosa/xform/parse/XFormParserTest.java
+++ b/test/org/javarosa/xform/parse/XFormParserTest.java
@@ -311,6 +311,27 @@ public class XFormParserTest {
         parse(r("body-before-model.xml"));
     }
 
+    @Test
+    public void parseFormWithTwoModels() throws IOException {
+        // Given & When
+        ParseResult parseResult = parse(r("two-models.xml"));
+
+        // Then
+        FormDef formDef = parseResult.formDef;
+        assertEquals(formDef.getTitle(), "Two Models");
+        assertEquals("Number of error messages", 1, parseResult.errorMessages.size());
+        assertEquals("Multiple models not supported. Ignoring subsequent models.", parseResult.errorMessages.get(0));
+        String firstModelInstanceId =
+                (String) formDef
+                        .getMainInstance()
+                        .getRoot()
+                        .getAttribute(null, "id")
+                        .getValue()
+                        .getValue();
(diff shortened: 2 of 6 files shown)


mdudzinski:
@lognaturel @dcbriccetti 
I want to cover:
* parsing the <setvalue> action (in progress)
* `FormInstanceParser#createMissingTemplates` & `removeInvalidTemplates` however I'm still not sure how to call out those lines and if this code has any value for ODK. Let me know if you know anything more :)

lognaturel:
@mdudzinski Is this ready for review?

mdudzinski:
@lognaturel Yes, sorry I didn't explicitly state that. 

It improves the coverage however there are still code flow paths that are not tested. The problem is in complexity and lack of documentation. I simply weren't sure how to cal out some of the code and I decided it will be better if we'll be adding tests as we go and learn more about JavaRosa and its codebase. I also think some of the code may become a way more testable in isolation - instead of just calling the parser against xml file -  after some refactoring in the future.

resources/form-with-setvalue-action.xml
```diff
@@ -0,0 +1,16 @@
...
+    <h:head>
+        <h:title>SetValue action</h:title>
+        <model>
+            <instance>
+                <data id="setvalue-action">
+                    <text/>
+                </data>
+            </instance>
+            <bind id="text-data-binding" nodeset="/data/text" type="string"/>
+            <setvalue bind="text-data-binding" event="xforms-ready" >Test Value</setvalue>
```
lognaturel: I did not even know `setvalue` was implemented.

test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -300,6 +301,16 @@ public void serializeAndRestoreMetaNamespaceFormInstance() throws IOException {
         assertEquals("/data/text", submissionProfile.getRef().getReference().toString());
     }
 
+    /**
+     * Simple tests that documents assumption that the model has to come before the body tag.
+     * According to the comment above {@link XFormParser#parseModel(Element)} method,
+     * this is not mandated by the specs but has been implemented this way to keep parsing simpler.
+     */
```
dcbriccetti: It’s good to have this implementation requirement documented, but I don’t see how this test helps. If we change the implementation to remove the restriction this test doesn’t fail. If someone in the field makes a form that violates the restriction, this test won’t catch it.

test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -331,6 +333,25 @@ public void parseFormWithTwoModels() throws IOException {
...
+        // Given & When
+        ParseResult parseResult = parse(r("form-with-setvalue-action.xml"));
+        FormDef formDef = parseResult.formDef;
+
+        // dispatch 'xforms-ready' action (Action.EVENT_XFORMS_READY)
+        formDef.initialize(true, new InstanceInitializationFactory());
+
+        // Then
+        assertEquals(formDef.getTitle(), "SetValue action");
+        assertEquals("Number of error messages", 0, parseResult.errorMessages.size());
```
dcbriccetti: Just a suggestion: Extract method -> assertNoParseErrors
Or maybe assertNumParseErrors(int).

test/org/javarosa/xform/parse/XFormParserTest.java
mdudzinski: @dcbriccetti Not sure if I follow. Once the restriction is removed, it should start failing because it expects a `RuntimeException` to be thrown during parsing by this code
`@Test(expected = RuntimeException.class)`.

test/org/javarosa/xform/parse/XFormParserTest.java
mdudzinski: I like this suggestion. I went with `assertNoParseErrors` as the IDE was whining that the expected number of errors is always 0 -_-

PR #212, opened 2017-10-31
Add count non empty func

### LINKED ISSUE #163: Add count-non-empty XForms 1.0 function
As described in opendatakit/xforms-spec#127 along with some tests.

This one should be approachable for a new contributor. The place to start looking is in [XPathFuncExpr](https://github.com/opendatakit/javarosa/blob/37effe59710b272c1d57f775eba422ff8474bed6/src/org/javarosa/xpath/expr/XPathFuncExpr.java). In particular, there are other examples that take nodesets as parameters. Join us in the #javarosa-code channel of [the Slack](slack.opendatakit.org) with any questions! 

Let's get this in the August release so pyxform can get the update. If no one picks it up in the next couple of weeks, @dcbriccetti it might be a nice palate cleanser between some of the more open-ended, tricky things you're looking at.

danielrees18:
Closes #163

Currently hung up on getting a unit test for this method. Looking for direction

It mirrors closely the method "count(object)" which uses the `.size()` method on the Nodeset to return the count. I considered not adding the method `.nonEmptySize()` to the nodeset, but I needed the `.toString()` value of the private `nodes` attribute to determine if the node was empty or not.

There are currently no tests to this code. Working to get that in.

I was attempting to add a test to `test/org/javarosa/xpath/test/XPathEvalTest.java` but i'm not familiar enough with Javarosa to actually know what to actually pass as an argument.

### DIFF
--- a/src/org/javarosa/xpath/XPathNodeset.java
+++ b/src/org/javarosa/xpath/XPathNodeset.java
@@ -19,7 +19,7 @@ import java.util.List;
  * instance.
  *
  * 2) A nodeset which wasn't able to reference into any known model (generally a reference which is
- * written in error). In this state, the size of the nodeset can be evaluated, but the acual reference
+ * written in error). In this state, the size of the nodeset can be evaluated, but the actual reference
  * cannot be returned, since it doesn't have any semantic value.
  *
  * (2) may be a deviation from normal XPath. This should be evaluated in the future.
@@ -130,6 +130,25 @@ public class XPathNodeset {
             return 0;
         }
         return nodes.size();
+    }
+
+    /**
+     * @return The number of nodes that are not empty
+     */
+    public int getNonEmptySize() {
+        if (nodes == null) {
+            return 0;
+        }
+
+        int count = 0;
+        for (TreeReference node : nodes) {
+            String value = node.toString();
+            if (value != null && !value.isEmpty()) {
+                ++count;
+            }
+        }
+
+        return count;
     }
 
     public TreeReference getRefAt (int i) {
--- a/src/org/javarosa/xpath/XPathNodeset.java
+++ b/src/org/javarosa/xpath/XPathNodeset.java
@@ -1,6 +1,7 @@
 package org.javarosa.xpath;
 
 import org.javarosa.core.model.condition.EvaluationContext;
+import org.javarosa.core.model.instance.AbstractTreeElement;
 import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.xpath.expr.XPathPathExpr;
@@ -142,8 +143,8 @@ public class XPathNodeset {
 
         int count = 0;
         for (TreeReference node : nodes) {
-            String value = node.toString();
-            if (value != null && !value.isEmpty()) {
+            AbstractTreeElement element = ec.getMainInstance().resolveReference(node);
+            if (element.getNumChildren() > 0 || element.getValue() != null) {
                 ++count;
             }
         }
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -95,7 +95,7 @@ public class XPathFuncExpr extends XPathExpression {
         if (o instanceof XPathFuncExpr) {
             XPathFuncExpr x = (XPathFuncExpr)o;
 
-            //Shortcuts for very easily comprable values
+            //Shortcuts for very easily comparable values
             //We also only return "True" for methods we expect to return the same thing. This is not good
             //practice in Java, since o.equals(o) will return false. We should evaluate that differently.
             //Dec 8, 2011 - Added "uuid", since we should never assume one uuid equals another
@@ -305,6 +305,9 @@ public class XPathFuncExpr extends XPathExpression {
         } else if (name.equals("count")) {
             assertArgsCount(name, args, 1);
             return count(argVals[0]);
+        } else if (name.equals("count-non-empty")) {
+            assertArgsCount(name, args, 1);
+            return countNonEmpty(argVals[0]);
         } else if (name.equals("sum")) {
             assertArgsCount(name, args, 1);
             if (argVals[0] instanceof XPathNodeset) {
@@ -1022,6 +1025,21 @@ public class XPathFuncExpr extends XPathExpression {
             throw new XPathTypeMismatchException("not a nodeset");
         }
     }
+
+    /**
+     * A node is considered non-empty if it is convertible into a string with a greater-than-zero length.
+     *
+     * @param o NodeSet to evaluate. Throws if not a NodeSet
+     * @return the number of non-empty nodes in argument node-set.
+     */
+    private int countNonEmpty (Object o) {
+        if (o instanceof XPathNodeset) {
+            return ((XPathNodeset) o).getNonEmptySize();
+        }
+
+        throw new XPathTypeMismatchException("not a nodeset");
+    }
+
 
     /**
      * sum the values in a nodeset; each element is coerced to a numeric value
--- a/test/org/javarosa/xpath/test/XPathEvalTest.java
+++ b/test/org/javarosa/xpath/test/XPathEvalTest.java
@@ -25,7 +25,6 @@ import java.util.Locale;
 import junit.framework.Test;
 import junit.framework.TestCase;
 import junit.framework.TestSuite;
-
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IFunctionHandler;
 import org.javarosa.core.model.data.IntegerData;
@@ -139,7 +138,7 @@ public class XPathEvalTest extends TestCase {
         FormInstance countNonEmptyInstance = createCountNonEmptyTestInstance();
 
         logTestCategory("counting");
-        testEval("count(/data)", countNonEmptyInstance, null, 1.0);
+        testEval("count(/data/path)", countNonEmptyInstance, null, 5.0);
         testEval("count-non-empty(/data/path)", countNonEmptyInstance, null, 3);
 
         logTestCategory("unsupporteds");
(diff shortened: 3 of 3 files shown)


src/org/javarosa/xpath/XPathNodeset.java
```diff
@@ -132,6 +132,25 @@ public int size () {
...
+     */
+    public int nonEmptySize() {
+        if(nodes == null) {
+            return 0;
+        }
+
+        int count = 0;
+        for (TreeReference node : nodes) {
+            String value = node.toString();
+            if (value != null && value.length() > 0) {
```
dcbriccetti: ! value.isEmpty() may be better

src/org/javarosa/xpath/XPathNodeset.java
```diff
@@ -132,6 +132,25 @@ public int size () {
...
+                count += 1;
```
dcbriccetti: ++count may be cleaner

src/org/javarosa/xpath/XPathNodeset.java
```diff
@@ -132,6 +132,25 @@ public int size () {
         return nodes.size();
     }
 
+    /**
+     * @return The number of nodes that are not empty
+     */
+    public int nonEmptySize() {
+        if(nodes == null) {
```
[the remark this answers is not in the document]
dcbriccetti: I also prefer `getNonEmptyCount`

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1018,6 +1021,21 @@ public static Double count (Object o) {
...
     /**
+     * A node is considered non-empty if it is convertible into a string with a greater-than zero length.
+     *
+     * @param o NodeSet to evaluate. Throws if not a NodeSet
+     * @return the number of non-empty nodes in argument node-set.
+     */
+    public static Double countNonEmpty (Object o) {
+        if (o instanceof XPathNodeset) {
+            return (double) ((XPathNodeset) o).nonEmptySize();
+        } else {
```
dcbriccetti: else after a return is not needed (nor any more clear, I would claim)

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1018,6 +1021,21 @@ public static Double count (Object o) {
     }
 
     /**
+     * A node is considered non-empty if it is convertible into a string with a greater-than zero length.
+     *
+     * @param o NodeSet to evaluate. Throws if not a NodeSet
+     * @return the number of non-empty nodes in argument node-set.
+     */
+    public static Double countNonEmpty (Object o) {
```
dcbriccetti: This should return `int`.  It’s not the concern of this method that its caller must return a reference variable. A discrete count should be an integer.

lognaturel: Yes, both great points. This one should be private.

dcbriccetti:
>I'd like to suggest we merge this and file a separate issue for testing functions that operate on node sets. @dcbriccetti @mdudzinski @iowerx what do you think?

I’m just looking at how this might be tested. I see there’s a form here: https://github.com/enketo/enketo-xpathjs/blob/master/test/doc.xml
And tests here: https://github.com/enketo/enketo-xpathjs/blob/master/test/spec/functions-openrosa.spec.js#L757-L785

I’d be inclined to see that this is demonstrated to be working before committing it, but @lognaturel perhaps you’re thinking it would be OK because nobody will be using it yet.

lognaturel:
> you’re thinking it would be OK because nobody will be using it yet.
Yes, exactly. Then we can iterate on a reasonable approach to testing all of those related functions as one chunk of work rather than tying it with this implementation. We should make sure the tests are in before the next release.

dcbriccetti:
OK, I have no objection, but (as you are probably already planning), I would like to see the changes you and I just asked for made first. I’m gonna spend a few minutes now seeing if I can run a test based on the Enketo tests and XML.

dcbriccetti:
It looks like Martijn added a div to his test data for this feature, in this commit:
https://github.com/enketo/enketo-xpathjs/commit/03fe3865473668a756b8de8c998fbe10a943bd2c
```
<div id="FunctionCountNonEmpty">
    <div>-5</div>
    <div>-15</div>
    <div></div>
    <p>
        <div></div>
        <div><!--comment--></div>
        <span></span>
        <span>
        </span>
    </p>
    <p></p>
</div>
```
I don’t see how to add that to our style of forms. I get errors like
```
XForm Parse Warning: Unrecognized element [div]. Ignoring and processing children...
    Problem found at nodeset: /html/body/div
    With element <div id="FunctionCountNonEmpty"><div>
```
@mdudzinski, maybe you have some ideas about testing this? Here’s what I tried. https://github.com/dcbriccetti/javarosa/commit/09d462e0dd120c927fbb925d2e6e44664087926d  I’ll suspend my efforts for now.

src/org/javarosa/xpath/XPathNodeset.java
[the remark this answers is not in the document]
danielrees18: Sounds good. I was following the pattern used in [`public int size()`](https://github.com/opendatakit/javarosa/pull/212/files#diff-bd0d119fa03312aee66f9ab889f532f7R128). Should that method be renamed to `getSize()` I'm also fine leaving it alone, I just want to be consistent

src/org/javarosa/xpath/expr/XPathFuncExpr.java
danielrees18: Sounds good, again though I wanted to be consistent with the method above: `public static Double count(Object)`. Should I change that method to be `private int` or create an issue as a follow up PR to make these changes so the two methods that provide counting are consistent?

src/org/javarosa/xpath/expr/XPathFuncExpr.java
danielrees18: Usually that's the pattern I prefer as well. Must have been late when I made this change 😄

src/org/javarosa/xpath/XPathNodeset.java
danielrees18: Just to verify, you're expecting `if (value != null && !value.isEmpty())` correct?

mdudzinski:
@dcbriccetti @lognaturel 
I don't think HTML tags like `div` are supported by ODK JavaRosa fork. `enketo-xpathjs` is a `Cross-browser OpenRosa-extended XPath 1.0 implementation in JavaScript`. Surely it has access to DOM so it can parse HTML tags. Perhaps we could do similar tests but with tags that are supported. Seems like the whole magic is in `documentEvaluate` function :)

I'll take a look into how to test such XPath functions in `org.javarosa.xpath.test.XPathEvalTest`.

mdudzinski:
@dcbriccetti @danielrees18 Please check d4e877d6. Seems like a form instance should be passed with some nodes to test such functions.

danielrees18:
@lognaturel @dcbriccetti An empty node is returning the string value of "/". Should the `countNonEmpty` assume that a value of "/" is empty?

Also thanks @mdudzinski for pointing out that test. I'm able to successfully get a count of 1 or 0 but cant figure out how to add more nodes to the tree so I can have 5 nodes, 2 empty and return a count of 3.

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1018,6 +1021,21 @@ public static Double count (Object o) {
     }
 
     /**
+     * A node is considered non-empty if it is convertible into a string with a greater-than zero length.
```
dcbriccetti: greater-than-zero or greater than zero

dcbriccetti:
>I'm able to successfully get a count of 1 or 0 but cant figure out how to add more nodes to the tree so I can have 5 nodes, 2 empty and return a count of 3.

Just guessing, but:

screen shot 2017-11-03 at 1 20 41 pm

mdudzinski:
@danielrees18 Dave - @dcbriccetti - is right about how to prepare a test data for an xpath function testing in `XPathEvalTest`. If you want a full working example, please check #220.

danielrees18:
@mdudzinski thanks, i'll update this PR tonight. Sorry for the delay

danielrees18:
Thanks for the help on this. I've updated the PR to reflect the suggestion that Dave pointed out for testing the countNonEmpty function.

My only question is that this test doesn't actually test that count excludes empty nodes, or maybe i'm just not understanding how the Nodeset works.

In the test, i add 3 nodes "/data/path" then the test checks `count-non-empty(/data/path)` which returns 3, as expected. But `count(/data/path)` will also return 3 because none of these nodes are empty. Is it possible to add a node to `/data/path` that is empty?

mdudzinski:
@danielrees18 I think it's because the implementation of `cound-non-empty` xpath function is wrong. Not sure why I didn't see it before but testing `TreeReference#toString()` against null doesn't seem to make sense. 
A node is empty if it either doesn't have any child or a value. To check that, you need to access the main instance. Please check 293573bb2f1d97af027a8ea41c991e19e4fd9216 with my implementation proposition. The test behaves now as you expected.

mdudzinski:
@danielrees18 
One more thing. I think we should care about the commit tree to be clean and readable. Which means commits like "last one" should be avoided. I believe you should squash all commits with imports fixes and merge into a last relevant comment ("Passing happy tests" I guess).

Commit messages could be improved as well. I think https://chris.beams.io/posts/git-commit/ is a good reference. I know it's hard to follow such restrict rules but we should try. It'll make our lives easier in the future :)

mdudzinski:
@danielrees18 Oh and after you fix the problems, you'll need to rebase your branch against the master to resolve the conflicts. 

You can do it also by merging master into yours but rebasing gives a way cleaner commits tree.

danielrees18:
@mdudzinski 

Thanks for all of the help on this PR

* Added changes from your commit, `count` now returns 5 while ` countNonEmpty` is returning 3 as expected
* resolved conflicts between this PR and master
* Squashed commits to only include relevant messages

lognaturel:
Thank you so much, @danielrees18! Welcome to the select group of JavaRosa experts. 😊

Issue #218, opened 2017-11-07
The repeat group with additionall group inside is skipped when I navigate backward


grzesiek2010:
#### Software versions 
JavaRosa v2.6.0

#### Problem description
The issue has been originally reported in [the Collect repository](https://github.com/opendatakit/collect/issues/1582) but after my investigation, I'm pretty sure it's on the Javarosa side.

The problem occurs when I want to use a repeat group with an additional group inside. That group can't display all questions on one screen (in that case everything looks ok). Then when I want to return to questions in such a group those questions are skipped and I'm moved to the question which is before the repeat group.

#### Steps to reproduce the problem
1. Use the attached form
2. Navigate Foward when you are on the first question
3. Add one group.
4. Navigate Foward when you are on the first question in the group
5. Navigate Foward when you are on the second question in the group
6. Don't add another group
7. Navigate Backward

#### Expected behavior
I should be moved to the last question in the repeat group.

#### Other information 
This form is a good example of the issue:
[rp1.xml.txt](https://github.com/opendatakit/javarosa/files/1449497/rp1.xml.txt)

Using the attached form I investigated [indexes the Javarosa](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/form/api/FormEntryController.java#L200) uses when I navigate in the form and here is a small example:

```
0, // Text - navigate forward
1_0, // Prompt new repeat - click on the add button
1_0, 0,
1_0, 0, 0, // Number - navigate forward
1_0, 0, 1, // Barcode - navigate forward
1_1 	// Prompt new repeat - click on the don't add button
-1 	// End of form - navigate backward
1_1 	// this is skipped automatically what is ok
	// QUESTION IS SKIPPED !!!
	// QUESTION IS SKIPPED !!!
1_0, 0,
1_0, 
0, 
```

Everything we do on the Collect side is in [stepToPreviousScreenEvent()](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/logic/FormController.java#L534) method. Looking at this part of code using the same form I can see that when I navigate backward (step 7):

[...]

PR #219, opened 2017-11-09
#213 Fix recalculations when deleting a repeat group

### LINKED ISSUE #213: Only a value in one group is recalculated after removing group.
#### Software versions 
JavaRosa v2.6.0

#### Problem description
I have a form with repeat group. The value of the first question is calculated `calculate="position(..)"`
If I remove one group, only the value in the next one is recalculated.

1. Open the attached form.
2. Add three groups
3. Go back to the first group and delete it.

#### Expected behavior
All values should be recalculated I think. But if I go to the second question it's still 3 like before.

#### Other information 
[RepeatGroupWithPosition.xml.txt](https://github.com/opendatakit/javarosa/files/1430192/RepeatGroupWithPosition.xml.txt)

mdudzinski:
Closes #213

So far only manual testing. Unit tests are in progress.

It seems to be most 'natural' to me. However there were other approaches considered. More details in the conversation under #213.
I think a really slow-down may be acceptable as `Safe2014DagImpl` seems to focous on being bugs-free. There are other implementations - like `Fast2014DagImpl` - which seem to be focused on the performance. Even at the expense of some bugs or not being as accurate as the safe implementations.

Well, when deleting a repeat, the default evaluation logic (`Safe2014DagImpl`) goes over all remaining repeats and fires all recalculations that were depended on the remaining repeat groups (their position) or their children. It might have a performance impact if there are a lot of repeats with depended calculations. I haven't tested it yet though.

### DIFF
--- a/src/org/javarosa/core/model/Safe2014DagImpl.java
+++ b/src/org/javarosa/core/model/Safe2014DagImpl.java
@@ -61,8 +61,9 @@ public class Safe2014DagImpl extends LatestDagBase {
          EvaluationContext evalContext, TreeReference deleteRef,
          TreeElement parentElement, TreeElement deletedElement) {
 
-      // After a repeat group has been deleted, the following repeat groups position has changed.
-      // Evaluate triggerables which depend on the position of the following repeats or on their children.
+      /* After a repeat group has been deleted, the following repeat groups position has changed.
+         Evaluate triggerables which depend on the repeat group reference. Directly or indirectly.
+       */
       final String repeatGroupName = deletedElement.getName();
       for (int i = deletedElement.getMultiplicity(); i < parentElement.getChildMultiplicity(repeatGroupName); i++) {
          final TreeElement repeatGroup = parentElement.getChild(repeatGroupName, i);
@@ -72,8 +73,17 @@ public class Safe2014DagImpl extends LatestDagBase {
                  new HashSet<QuickTriggerable>(0));
          publishSummary("Deleted", repeatGroup.getRef(), alreadyEvaluated);
 
-         evaluateChildrenTriggerables(mainInstance, evalContext,
-                 repeatGroup, false, alreadyEvaluated);
+         if (repeatGroup.getRef().equals(deleteRef)) {
+            /* Evaluate the children triggerables only once, for the deleted repeat group.
+               Only children of the deleted repeat group have actually changed (they're gone) and thus calculations depend
+               on the must be re-evaluated, the following repeat groups have been shifted along with their children.
+               If there are calculations - regardless if inside the repeat or outside - that depend on the following
+               repeat group positions, they will fired cascade by the above code anyway.
+               Unit test for this scenario:
+               Safe2014DagImplTest#deleteThirdRepeatGroup_evaluatesTriggerables_indirectlyDependentOnTheRepeatGroupsNumber
+             */
+            evaluateChildrenTriggerables(mainInstance, evalContext, repeatGroup, false, alreadyEvaluated);
+         }
       }
    }
 
--- a/src/org/javarosa/core/model/Safe2014DagImpl.java
+++ b/src/org/javarosa/core/model/Safe2014DagImpl.java
@@ -61,13 +61,20 @@ public class Safe2014DagImpl extends LatestDagBase {
          EvaluationContext evalContext, TreeReference deleteRef,
          TreeElement parentElement, TreeElement deletedElement) {
 
-      Set<QuickTriggerable> alreadyEvaluated = triggerTriggerables(
-            mainInstance, evalContext, deleteRef,
-            new HashSet<QuickTriggerable>(0));
-      publishSummary("Deleted", deleteRef, alreadyEvaluated);
+      // After a repeat group has been deleted, the following repeat groups position has changed.
+      // Evaluate triggerables which depend on the position of the following repeats or on their children.
+      final String repeatGroupName = deletedElement.getName();
+      for (int i = deletedElement.getMultiplicity(); i < parentElement.getChildMultiplicity(repeatGroupName); i++) {
+         final TreeElement repeatGroup = parentElement.getChild(repeatGroupName, i);
 
-      evaluateChildrenTriggerables(mainInstance, evalContext,
-            deletedElement, false, alreadyEvaluated);
+         Set<QuickTriggerable> alreadyEvaluated = triggerTriggerables(
+                 mainInstance, evalContext, repeatGroup.getRef(),
+                 new HashSet<QuickTriggerable>(0));
+         publishSummary("Deleted", repeatGroup.getRef(), alreadyEvaluated);
+
+         evaluateChildrenTriggerables(mainInstance, evalContext,
+                 repeatGroup, false, alreadyEvaluated);
+      }
    }
 
    @Override
--- a/test/org/javarosa/core/model/Safe2014DagImplTest.java
+++ b/test/org/javarosa/core/model/Safe2014DagImplTest.java
@@ -82,6 +82,263 @@ public class Safe2014DagImplTest {
         }
     }
 
+    @Test
+    public void deleteSecondRepeatGroup_evaluatesTriggerables_dependentOnTheParentPosition() throws Exception {
+        // Given
+        final FormDef formDef =
+                parse(r("repeat-group-with-children-calculations-dependent-on-the-parent.xml")).formDef;
+
+        assertIDagImplUnderTest(formDef);
+
+        formDef.initialize(false, new InstanceInitializationFactory()); // trigger all calculations
+        formDef.setEventNotifier(eventNotifier); // it's important to set the test event notifier now to avoid storing events from the above initialization
+
+        final FormInstance mainInstance = formDef.getMainInstance();
+
+        final TreeElement elementToBeDeleted = mainInstance.getRoot().getChildAt(2);
+        final TreeReference elementToBeDeletedRef = elementToBeDeleted.getRef();
+
+        // Index pointing to the second repeat group
+        final FormIndex indexToBeDeleted = new FormIndex(0, 1, elementToBeDeletedRef);
+
+        // When
+
+        // Safe2014DagImplTest.deleteRepeatGroup is called by the below method
+        formDef.deleteRepeat(indexToBeDeleted);
+
+        // Then
+        final List<TreeElement> repeats = mainInstance.getRoot().getChildrenWithName("houseM");
+
+        // check the values based on the position of the parents
+        assertThat(repeats.get(0).getChildAt(0).getValue().getDisplayText(), equalTo("1"));
+        assertThat(repeats.get(0).getChildAt(2).getValue().getDisplayText(), equalTo("A1"));
+        assertThat(repeats.get(1).getChildAt(0).getValue().getDisplayText(), equalTo("2"));
+        assertThat(repeats.get(1).getChildAt(2).getValue().getDisplayText(), equalTo("C2"));
+        assertThat(repeats.get(2).getChildAt(0).getValue().getDisplayText(), equalTo("3"));
+        assertThat(repeats.get(2).getChildAt(2).getValue().getDisplayText(), equalTo("D3"));
+        assertThat(repeats.get(3).getChildAt(0).getValue().getDisplayText(), equalTo("4"));
+        assertThat(repeats.get(3).getChildAt(2).getValue().getDisplayText(), equalTo("E4"));
+
+        // check that correct calculations were triggered
+        final String[] expectedMessages = {
+                "Processing 'Recalculate' for no [2_1] (2.0)",
+                "Processing 'Recalculate' for name_and_no [2_1] (C2)",
+                "Processing 'Deleted: houseM [2]: 2 triggerables were fired.' for ",
+                "Processing 'Deleted: no [2_1]: 0 triggerables were fired.' for ",
+                "Processing 'Deleted: name [2_1]: 0 triggerables were fired.' for ",
+                "Processing 'Deleted: name_and_no [2_1]: 2 triggerables were fired.' for ",
+                "Processing 'Recalculate' for no [3_1] (3.0)",
+                "Processing 'Recalculate' for name_and_no [3_1] (D3)",
+                "Processing 'Deleted: houseM [3]: 2 triggerables were fired.' for ",
+                "Processing 'Recalculate' for no [4_1] (4.0)",
+                "Processing 'Recalculate' for name_and_no [4_1] (E4)",
+                "Processing 'Deleted: houseM [4]: 2 triggerables were fired.' for ",
+        };
+
+        assertThat(dagEvents.size(), equalTo(expectedMessages.length));
+
+        int messageIndex = 0;
+        for (String expectedMessage : expectedMessages) {
+            assertThat(dagEvents.get(messageIndex++).getDisplayMessage(), equalTo(expectedMessage));
+        }
+    }
+
+    @Test
+    public void deleteSecondRepeatGroup_doesNotEvaluateTriggerables_notDependentOnTheParentPosition() throws Exception {
+        // Given
+        final FormDef formDef =
+                parse(r("repeat-group-with-children-calculations-not-dependent-on-the-parent.xml")).formDef;
+
+        assertIDagImplUnderTest(formDef);
+
+        formDef.initialize(false, new InstanceInitializationFactory()); // trigger all calculations
+        formDef.setEventNotifier(eventNotifier); // it's important to set the test event notifier now to avoid storing events from the above initialization
+
+        final FormInstance mainInstance = formDef.getMainInstance();
+
+        final TreeElement elementToBeDeleted = mainInstance.getRoot().getChildAt(2);
+        final TreeReference elementToBeDeletedRef = elementToBeDeleted.getRef();
+
+        // Index pointing to the second repeat group
+        final FormIndex indexToBeDeleted = new FormIndex(0, 1, elementToBeDeletedRef);
+
+        // When
+
+        // Safe2014DagImplTest.deleteRepeatGroup is called by the below method
+        formDef.deleteRepeat(indexToBeDeleted);
+
+        // Then
+        final List<TreeElement> repeats = mainInstance.getRoot().getChildrenWithName("houseM");
+
+        // check the values based on the position of the parents
+        assertThat(repeats.get(0).getChildAt(0).getValue().getDisplayText(), equalTo("1"));
+        assertThat(repeats.get(0).getChildAt(2).getValue().getDisplayText(), equalTo("AX"));
+        assertThat(repeats.get(1).getChildAt(0).getValue().getDisplayText(), equalTo("2"));
+        assertThat(repeats.get(1).getChildAt(2).getValue().getDisplayText(), equalTo("CX"));
+        assertThat(repeats.get(2).getChildAt(0).getValue().getDisplayText(), equalTo("3"));
+        assertThat(repeats.get(2).getChildAt(2).getValue().getDisplayText(), equalTo("DX"));
+        assertThat(repeats.get(3).getChildAt(0).getValue().getDisplayText(), equalTo("4"));
+        assertThat(repeats.get(3).getChildAt(2).getValue().getDisplayText(), equalTo("EX"));
+
+        // check that correct calculations were triggered
+        final String[] expectedMessages = {
+                "Processing 'Recalculate' for no [2_1] (2.0)",
+                "Processing 'Deleted: houseM [2]: 1 triggerables were fired.' for ",
+                "Processing 'Deleted: no [2_1]: 1 triggerables were fired.' for ",
+                "Processing 'Recalculate' for name_concat [2_1] (CX)",
+                "Processing 'Deleted: name [2_1]: 1 triggerables were fired.' for ",
+                "Processing 'Deleted: name_concat [2_1]: 1 triggerables were fired.' for ",
+                "Processing 'Recalculate' for no [3_1] (3.0)",
+                "Processing 'Deleted: houseM [3]: 1 triggerables were fired.' for ",
+                "Processing 'Recalculate' for no [4_1] (4.0)",
+                "Processing 'Deleted: houseM [4]: 1 triggerables were fired.' for "
+        };
+
+        assertThat(dagEvents.size(), equalTo(expectedMessages.length));
+
+        int messageIndex = 0;
+        for (String expectedMessage : expectedMessages) {
+            assertThat(dagEvents.get(messageIndex++).getDisplayMessage(), equalTo(expectedMessage));
+        }
+    }
+
+    @Test
+    public void deleteThirdRepeatGroup_evaluatesTriggerables_dependentOnTheRepeatGroupsNumber() throws Exception {
+        // Given
+        final FormDef formDef =
+                parse(r("calculation-dependent-on-the-repeat-groups-number.xml")).formDef;
+
+        assertIDagImplUnderTest(formDef);
+
+        formDef.initialize(false, new InstanceInitializationFactory()); // trigger all calculations
+        formDef.setEventNotifier(eventNotifier); // it's important to set the test event notifier now to avoid storing events from the above initialization
+
+        final FormInstance mainInstance = formDef.getMainInstance();
+
+        final TreeElement elementToBeDeleted = mainInstance.getRoot().getChildAt(2);
+        final TreeReference elementToBeDeletedRef = elementToBeDeleted.getRef();
+
+        // Index pointing to the second repeat group
+        final FormIndex indexToBeDeleted = new FormIndex(0, 2, elementToBeDeletedRef);
+
+        // When
+
+        TreeElement summaryNode = mainInstance.getRoot().getChildrenWithName("summary").get(0);
+        assertThat(summaryNode.getValue().getDisplayText(), equalTo("55")); // check the calculation result for 10 repeat groups
+
+        // Safe2014DagImplTest.deleteRepeatGroup is called by the below method
+        formDef.deleteRepeat(indexToBeDeleted);
+
+        // Then
+        final List<TreeElement> repeats = mainInstance.getRoot().getChildrenWithName("houseM");
+
+        // check the values based on the position of the parents
+        assertThat(repeats.get(0).getChildAt(0).getValue().getDisplayText(), equalTo("1"));
+        assertThat(repeats.get(1).getChildAt(0).getValue().getDisplayText(), equalTo("2"));
+        assertThat(repeats.get(2).getChildAt(0).getValue().getDisplayText(), equalTo("3"));
+        assertThat(repeats.get(3).getChildAt(0).getValue().getDisplayText(), equalTo("4"));
+        assertThat(repeats.get(4).getChildAt(0).getValue().getDisplayText(), equalTo("5"));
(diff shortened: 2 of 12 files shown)


mdudzinski:
@ln 
This one can be reviewed and/or tested.
I think that I figured out a better way for testing triggerables evaluation. It doesn't use API classes like `FormEntryController` and `FormEntryModel`. Instead, the test parsse a form with main instance that has test data and calls `Safe2014DagImpl#deleteRepeatGroup` in an isolation.

mdudzinski:
@ln @dcbriccetti This one also includes a proposition for managing test forms. We'll keep adding them to test various new or existing functionalities and having all of them in a same directory may become cumbersome at some point. So I propose keeping test forms in the same packages (inside resources directory) as the test classes. 
The only drawback I can see is a need to put the full path to the test form when writing a test. But I think the test form parser could be improved so it could infer the path to a given form basing on the package name of the test class.

lognaturel:
@mdudzinski and I discussed in Slack. This will have performance implications on instances that have lots of repeats. Would be good to get a sense of how significant the impact is.

dcbriccetti:
>Would be good to get a sense of how significant the impact is.

Shall I wait for that before reviewing?

lognaturel:
I think a quick pass with first impressions would be helpful.

test/org/javarosa/test/utils/FormParserHelper.java
```diff
@@ -0,0 +1,47 @@
+package org.javarosa.test.utils;
```
dcbriccetti: I like this change, but is there a reason you moved the classes out of the xform.parse package? I think they belong there, rather than in a new “utils” package.

test/org/javarosa/test/utils/ResourcePathHelper.java
```diff
@@ -0,0 +1,11 @@
+package org.javarosa.test.utils;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class ResourcePathHelper {
+    /** Makes a Path for a resource file */
+    public static Path r(String filename) {
```
dcbriccetti: I really like moving the forms within the package under resources, but why not go “all the way to town”?

```
/** Makes a Path for a resource file */
public static Path r(String filename) {
    return Paths.get("resources", filename);
}

/**
 * Makes a Path for a resource file in the package identified by <code>classForPackage</code>
 * @param classForPackage the class whose path is to be used to locate the file in the resources directory
 * @param filename the name of the file in the subdirectory of the resources directory
 * @return a Path
 */
public static Path r(Class<?> classForPackage, String filename) {
    return Paths.get("resources", classForPackage.getPackage().getName().replace('.', '/'), filename);
}
```

Call like this:
`r(getClass(), "repeat-group-with-position-calculation.xml")`

test/org/javarosa/core/model/Safe2014DagImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+        // Then
+        final List<TreeElement> repeats = mainInstance.getRoot().getChildrenWithName("houseM");
+
+        // check the values based on the position of the parents
+        assertThat(repeats.get(0).getChildAt(0).getValue().getDisplayText(), equalTo("1"));
+        assertThat(repeats.get(1).getChildAt(0).getValue().getDisplayText(), equalTo("2"));
+        assertThat(repeats.get(2).getChildAt(0).getValue().getDisplayText(), equalTo("3"));
+        assertThat(repeats.get(3).getChildAt(0).getValue().getDisplayText(), equalTo("4"));
+
+        // check that calculations have not been triggered for the repeat group prior to the deleted one
```
dcbriccetti: Would you consider this, to improve readability and maintainability?

```
// check that calculations have not been triggered for the repeat group prior to the deleted one
final String[] expectedMessages = {
        "Processing 'Recalculate' for no [2_1] (2.0)",
        "Processing 'Deleted: houseM [2]: 1 triggerables were fired.' for ",
        "Processing 'Deleted: no [2_1]: 1 triggerables were fired.' for ",
        "Processing 'Recalculate' for no [3_1] (3.0)",
        "Processing 'Deleted: houseM [3]: 1 triggerables were fired.' for ",
        "Processing 'Deleted: no [3_1]: 1 triggerables were fired.' for ",
        "Processing 'Recalculate' for no [4_1] (4.0)",
        "Processing 'Deleted: houseM [4]: 1 triggerables were fired.' for ",
        "Processing 'Deleted: no [4_1]: 1 triggerables were fired.' for "
};
int messageIndex = 0;
for (String expectedMessage : expectedMessages) {
    assertThat(dagEvents.get(messageIndex++).getDisplayMessage(), equalTo(expectedMessage));
}
```

test/org/javarosa/core/model/Safe2014DagImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+           At position 0 there is a template TreeElement (multiplicity == -2)
+           At position 1 is a repeat group with an already populated populated
+           At positions 2 - 4 there are repeats with values supposed to be populated after deleteRepeatGroup is called
+        */
+        final TreeElement deletedElement = mainInstance.getRoot().getChildAt(2);
+        final TreeReference dummyDeletedRef = null;
+
+        // When
+        dag.deleteRepeatGroup(mainInstance, formDef.getEvaluationContext(), dummyDeletedRef, parentElement, deletedElement);
```
dcbriccetti: I’m wondering if it would be safer, more realistic, and/or simpler to call formdef.deleteRepeat(FormIndex index) rather than getting the dag and calling deleteRepeatGroup directly.

test/org/javarosa/core/model/Safe2014DagImplTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Test
+    public void deleteRepeatGroup_evaluatesTriggerables_dependedOnAllRepeatGroupSiblings() throws Exception {
+        // Given
+        final FormDef formDef =
+                parse(r("org/javarosa/core/model/repeat-group-with-position-calculation.xml")).formDef;
+
+        formDef.setEventNotifier(eventNotifier);
+        IDag dag = getIDagImplUnderTest(formDef);
+        assertThat(dag, instanceOf(Safe2014DagImpl.class));
```
dcbriccetti: getIDagImplUnderTest casts to this, so you’ll crash before failing this assertion.

dcbriccetti:
Hey, this is complex stuff! Hats off to you, @mdudzinski. I made a few review comments, and I’ll continue later today or tomorrow.

test/org/javarosa/test/utils/ResourcePathHelper.java
dcbriccetti: I also like `r(this, ...)` because it’s shorter, but some might be bothered (myself included) that far more than needed was passed as an argument.

mdudzinski: @dcbriccetti Actually I thought about this but wanted to use `Thread.currentThread().getStackTrace()` to automatically infer the package of the calling class. I'm fine with both approaches.

test/org/javarosa/core/model/Safe2014DagImplTest.java
mdudzinski: @dcbriccetti I love this! Will use this approach.

test/org/javarosa/core/model/Safe2014DagImplTest.java
mdudzinski: @dcbriccetti Well, I mostly wanted to test dag in isolation but it should be possible to test it via `formdef.deleteRepeat(FormIndex index)`. 
Since I'd like to avoid using API classes like `FormEntryModel` and/or `FormEntryController`, the `FormIndex index` would need to be constructed from scratch. After my time spent on #218 (where I was analyzing FormIndex inc/dec codebase) , it shouldn't be a problem now.

test/org/javarosa/core/model/Safe2014DagImplTest.java
mdudzinski: @dcbriccetti Nice catch! Thanks

test/org/javarosa/test/utils/FormParserHelper.java
mdudzinski: @dcbriccetti Since this is code is now used by tests outside of the`xform.parse` package, I wanted to put it in a common place, not related to any particular feature of the production code (like xform parsing).

test/org/javarosa/test/utils/ResourcePathHelper.java
dcbriccetti: >getStackTrace

Oh, that’s the best! It hadn’t occurred to me.

mdudzinski:
@lognaturel 
I tested the performance (on an AVD) impact on the following form:
[delete-repeat-recalculate.xml.txt](https://github.com/opendatakit/javarosa/files/1495912/delete-repeat-recalculate.xml.txt)
It has 100 repeats with the following calculations:
```
            <bind calculate="position(..)" nodeset="/data/household/index" type="int"/>
            <bind calculate="log(pow(position(..), 5))" nodeset="/data/household/somecalculation" type="decimal"/>
```
and there was no performance impact. Or at least it wasn't noticeable. 

To get more reliable test results, I asked @mmarciniak90 to test this form on an actual device (Thanks Marzena!).
She tested this on Android 4.4 and 4.1 and confirmed there were no noticeable performance impact as well.

I'm now working on fixing/implementing suggestions and concerns from the review. It should be available for the review again soon.

test/org/javarosa/core/model/Safe2014DagImplTest.java
mdudzinski: @dcbriccetti I followed your suggestion and now the test looks way better. Thanks!

test/org/javarosa/core/model/Safe2014DagImplTest.java
mdudzinski: @dcbriccetti Fixed

mdudzinski:
@dcbriccetti @lognaturel I updated this PR, please review it when you get a chance. Thanks!

test/org/javarosa/test/utils/ResourcePathHelper.java
mdudzinski: @dcbriccetti Done 😃

src/org/javarosa/core/model/Safe2014DagImpl.java
```diff
@@ -61,13 +61,20 @@ public void deleteRepeatGroup(FormInstance mainInstance,
...
 
-      Set<QuickTriggerable> alreadyEvaluated = triggerTriggerables(
-            mainInstance, evalContext, deleteRef,
-            new HashSet<QuickTriggerable>(0));
-      publishSummary("Deleted", deleteRef, alreadyEvaluated);
-
-      evaluateChildrenTriggerables(mainInstance, evalContext,
-            deletedElement, false, alreadyEvaluated);
+      // After a repeat group has been deleted, the following repeat groups position has changed.
+      // Evaluate triggerables which depend on the position of the following repeats or on their children.
```
lognaturel: This makes it sound like it only evaluates certain triggerables. Maybe it should say something more like "evaluate triggerables for all following repeats and their children?" Let me know if I'm just misunderstanding the wording.

mdudzinski: @lognaturel  At the first glance it looks like it evaluates only directly related triggerables like `position(..)` but `Safe2014DagImpl` is super smart and will also fire all indirectly related ones. For example:
`<bind calculate="concat(/rgwp/houseM/name, /rgwp/houseM/no)" nodeset="/rgwp/houseM/namecalc" type="string"/>`
This one uses `/rgwp/houseM/no` value which explicitly depends on the position. The dag implementation will fire the concatenation calculation after the position calculation for the `/rgwp/houseM/no` node. 

On the other hand, the following 
`<bind calculate="concat(/rgwp/houseM/name, 'some string value')" nodeset="/rgwp/houseM/namecalc" type="string"/>`
won't be fired because the concatenation doesn't rely at all on the position within the parent. 
This is pretty amazing I think.

Anyway, both "evaluate triggerables for all following repeats and their children" and " evaluate triggerables which depend on the position of the following repeats or on their children" may be confusing. Perhaps we should document the above with a new unit test scenario?

lognaturel: That's cool! So we ended up with your "ideal" solution after all.

Adding a test and doing your best to clarify the comment both sound good. Perhaps in the comment you could link to the relevant tests.

test/org/javarosa/test/utils/ResourcePathHelper.java
```diff
@@ -0,0 +1,35 @@
...
+            return Paths.get("resources", filename);
+        }
+    }
+
+    /**
+     * If the class that called {@link ResourcePathHelper#r(String)} is {@link org.javarosa.core.model.Safe2014DagImpl}
+     * then this method will return "org/javarosa/core/model"
+     * @return Caller test class package as a path
+     */
+    private static String inferResourceFileParentPath() {
```
dcbriccetti: I love this, and would not have thought of it. How about making it more resilient, though, by removing the assumption that the caller will always be exactly two level up. Something like this, perhaps?

```
    private static String inferResourceFileParentPath() {
        final StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
        int callerStackIndex = 2; // 0 is getStackTrace, 1 is this method, 2 is immediate caller
        while (callerStackIndex < stackTrace.length &&
                stackTrace[callerStackIndex].getClassName().equals(ResourcePathHelper.class.getName())) {
            ++callerStackIndex;
        }
        final StackTraceElement callingClass = stackTrace[callerStackIndex];
        final String callerClassName = callingClass.getClassName();
        return callerClassName
                .substring(0, callerClassName.lastIndexOf(".")) // strip the class name
                .replaceAll("\\.", "\\" + File.separator); // change all '.' to '/' ('\' on Windows)
    }
```

test/org/javarosa/test/utils/ResourcePathHelper.java
```diff
@@ -0,0 +1,35 @@
...
+        final StackTraceElement callingClass =  Thread.currentThread().getStackTrace()[CALLER_CLASS_INDEX_IN_STACKTRACE];
+        final String callerPackage = callingClass.getClassName();
+        return callerPackage
+                .substring(0, callerPackage.lastIndexOf(".")) // strip the class name
+                .replaceAll("\\.", "\\" + File.separator); // change all '.' to '/' ('\' on Windows)
```
dcbriccetti: `"\\" + File.separator` gives me `"\/"` (on Mac). Is that better than `"/"`? The second argument to replaceAll is a simple string. But, why not avoid regexes altogether and use `replace(".", File.separator)`?

test/org/javarosa/test/utils/FormParserHelper.java
dcbriccetti: I’m not sure I understand. If the classes are related to parsing, it makes sense to have them in a `parse` package, even if they are used from classes outside the package.

dcbriccetti:
>I tested the performance (on an AVD) impact on the following form: delete-repeat-recalculate.xml.txt

Might it be useful to make a test that does many deletes from that form, so we can time it? Or are we confident that performance is acceptable? Might there be actual forms with many thousands of elements, such as there is with the Nigeria Wards form?

test/org/javarosa/test/utils/ResourcePathHelper.java
mdudzinski: @dcbriccetti Why not! I'll take this, thanks!

test/org/javarosa/test/utils/ResourcePathHelper.java
mdudzinski: @dcbriccetti it just worked on my mac so I didn't check the output. I'll take a look at your suggestion. Thanks!

mdudzinski:
> Might it be useful to make a test that does many deletes from that form, so we can time it? Or are we confident that performance is acceptable? Might there be actual forms with many thousands of elements, such as there is with the Nigeria Wards form?

There is no point in timing many deletes because you can't do this in batch. What matters is a time of a single deletion when there is a lot of following repeats with calculations to be recomputed. 
And it's important that not every calculation will be recomputed, only those that actually depend on the parent position. More details can be found [here](https://github.com/opendatakit/javarosa/pull/219#discussion_r153533480).

Of course it's possible to create a form with way more complex computation depending on the parent position than we tested [here](https://github.com/opendatakit/javarosa/pull/219#issuecomment-346336293) and at some point a user will start seeing the performance impact but not sure if that would be a real life production scenario.

test/org/javarosa/test/utils/FormParserHelper.java
mdudzinski: @dcbriccetti Hmm I just treated this class as a helper class and wanted to keep it outside of packages related to the actual production code. On second thought, it actually tests the form parsing codebase so it *is* related to the production code. I'll move it back to the `xform.parse` package. Thanks!

test/org/javarosa/test/utils/ResourcePathHelper.java
mdudzinski: @dcbriccetti But anyway I forgot about `replace` method which doesn't use regexes. I always try to avoid them whenever possible 😃 
As  Jamie Zawinski said:
>Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

test/org/javarosa/test/utils/ResourcePathHelper.java
mdudzinski: @dcbriccetti done!

test/org/javarosa/test/utils/ResourcePathHelper.java
mdudzinski: @dcbriccetti done!

test/org/javarosa/test/utils/FormParserHelper.java
mdudzinski: @dcbriccetti Done!

mdudzinski:
Updated PR according to the latest suggestions from the code review but I'm still about to add new tests and improve a bit the comments around the changed method.

src/org/javarosa/core/model/Safe2014DagImpl.java
mdudzinski: @lognaturel Ayay I must've made a mistake yesterday while debugging this! It's smart but not as smart as I thought. It doesn't re-evaluate same triggerable when a repeat is deleted but it will re-evaluate any other triggerable inside of the repeat. 
So all I said for
`<bind calculate="concat(/rgwp/houseM/name, /rgwp/houseM/no)" nodeset="/rgwp/houseM/namecalc" type="string"/>` 
is correct but it is not for
`<bind calculate="concat(/rgwp/houseM/name, 'some string value')" nodeset="/rgwp/houseM/namecalc" type="string"/>`.

And I know why this happens. DAG goes over each child of each repeat and looks for triggerables that are dependent on the child reference. In this scenario, it finds `calculate="concat(/rgwp/houseM/name, 'some string value')" `. This should be improved because this triggerable depends on `/rgwp/houseM/name` which doesn't depends on a calculation related to the parent reference - `/rgwp/houseM`. 
I can think about fixing it now or raise an issue for this with some information I already have.

Sorry for the confusion. It came out when I created a test for this. It shows how desperate JavaRosa needs tests.
So no ideal solution for now 😞

mdudzinski: @lognaturel if fixing the above behavior isn't trivial, I can test some way more complex calculations in a repeat and see if there is still no impact on the performance.

lognaturel:
Great that the tests are helping uncover interesting things! I think creating an issue describing the triggerables that are reevaluated but that shouldn't need to be is a good idea. Seems adding in a timing test would be valuable as well. I think what @dcbriccetti is suggesting is to simulate multiple deletes to get an average timing value that is representative of reality. Would be good to try up to 1000 repeats.

mdudzinski:
@lognaturel Sure, I can add a timing test. Any idea how much complicated calculations should be in the test form?

Now I'm investigating calling ` evaluateChildrenTriggerables(mainInstance, evalContext,          repeatGroup, false, alreadyEvaluated)` for each following repeat. This triggers unnecessary re-calculations for some of my test forms but I'm wondering if there is a production scenario where this re-evaluation is *not* unnecessary.

lognaturel:
@mdudzinski I don't think the calculations need to be very complicated. My guess is that position is generally used only for display and not as part of larger expressions. Are you thinking about the complexity of other child/sibling calculations as well? Something like a few integer fields and a calculation of their sum seems like it would be representative.

mdudzinski:
@lognaturel Okay so I did a tweak to the fix and now only children of the actually deleted repeat have their triggerables evaluated. Turned out it's unnecessary to evaluate children triggs for all the following repeats. I added bunch of tests to cover this and other scenarios. In the last commit - still marked as wip - I added a timing test and honestly I don't like the results... The test form is `delete-repeat-group-with-calculations-timing-test.xml` and there are couple of bindings commented to not stress the CI 😛 

So where there are 1000 repeats and the following bindings:
```
<bind calculate="position(..)" nodeset="/rgwp/houseM/current_position" type="int"/>
<bind calculate="pow(/rgwp/houseM/current_position, 3)" nodeset="/rgwp/houseM/int1" type="int"/>
<bind calculate="log10(/rgwp/houseM/int1)" nodeset="/rgwp/houseM/double1" type="double"/>
<bind calculate="round(/rgwp/houseM/double1)" nodeset="/rgwp/houseM/int2" type="int"/>
<bind calculate="sum(/rgwp/houseM/current_position)" nodeset="/rgwp/summary" type="double"/>
```
deleting one repeat takes about 4.25 seconds. Since the test deletes all the repeats starting from the first one and the fix evaluates triggerables only for the following groups, this time should be decreasing with each iteration (because of less repeat group to iterate through).

Interesting is that calculations inside of the repeat (pow, log and round) have very little impact on this time and this test run against:
```
<bind calculate="position(..)" nodeset="/rgwp/houseM/current_position" type="int"/>
<bind calculate="sum(/rgwp/houseM/current_position)" nodeset="/rgwp/summary" type="double"/>
```
shows that deletion of a single repeat takes about 3.9 ~ 4 seconds. 

So the sum function is a bottleneck here. When you take it out and leave the remaining calculations, everything works quite smoothly (100 - 200 millis for a single deletion at the beginning of the loop).

When you reduce the amount of repeats to 500 and leave the sum, then deleting a repeat group at the beginning of the loop takes about 0.4 - 0.5 seconds. At the end it takes less than 0.001 of a second.

mdudzinski:
This also means that if you deleting a 800th repeat group will be way faster then deleting a second one.

test/org/javarosa/core/model/Safe2014DagImplTest.java
```diff
@@ -0,0 +1,396 @@
...
+        long startMs = System.currentTimeMillis();
+
+        for (int i = 0; i < numberOfRepeats; i++) {
+            long currentIterationStart = System.currentTimeMillis();
+            formDef.deleteRepeat(firstRepeatIndex);
+            long took = System.currentTimeMillis() - currentIterationStart;
+            String tookFormatted = LocalTime.fromMillisOfDay(took).toString();
+            System.out.println("Deletion of repeat #" + i + " took " + tookFormatted);
+        }
+
```
dcbriccetti: Just a thought. If you do something like this:
```
System.out.println("Repeat Group\nDeletion Time (ms)");
for (int i = 0; i < numberOfRepeats; i++) {
    long currentIterationStart = System.nanoTime();
    formDef.deleteRepeat(firstRepeatIndex);
    double tookMs = (System.nanoTime() - currentIterationStart) / 1000000D;
    System.out.printf("%d\t%.3f\n", i, tookMs);
}
```
Then we can do neat things like this:

deletion times

mdudzinski: @dcbriccetti Thanks, I'll fix the output to match this format.

@lognaturel 
If the deletion times are acceptable then I'd like to clean up the test so it can be merged. The number of iterations should be decreased to not slow down build times - what would say about lowering it to 200 elements? Deleting all repeats takes about 3.5 seconds in that case.

mdudzinski: @dcbriccetti @lognaturel I pushed the clean-up timing test. It uses format suggested by @dcbriccetti and tests deletion of 200 repeats.

lognaturel:
Cleaning things up and getting it ready for merging sounds great. I think we can get this into the next Collect release — code freeze Wednesday and release on Sunday or so.

test/org/javarosa/core/model/Safe2014DagImplTest.java
dcbriccetti: What we did for the timing “tests” in XFormParserTest was make them run very fast (a fraction of a second), but have a comment with instructions for anyone wanting to make them run longer, for actual timing purposes:
`        for (double powerOfTen = 3; powerOfTen <= 4.0; powerOfTen += 0.1) {  // Raise this upper limit to really measure
`

mdudzinski: @dcbriccetti Right, thanks! I added a comment and left 200 repeats. It runs fast and it's still quite a lot.

lognaturel:
Thanks for approaching this in such a thoughtful way, @mdudzinski! The test cases are really good and I can't think of others. These are really informative tests.

I understand that `sum` takes a lot less time when there are fewer repeats overall -- when you've deleted 400 out of 500 for example.

> This also means that if you deleting a 800th repeat group will be way faster then deleting a second one.

This makes it sound like if I have a total of 1000 repeat and delete one in either the 800th or 2nd position the time to compute the sum will be different. Is that really the case? If so, how? I wouldn't expect there to be a shortcut for that one -- seems it would just have to re-compute the whole thing. Similarly, I would expect that adding a new repeat (say, the 1001st) would lead to the `sum` outside the repeat being re-computed and taking just as much time as a delete in that case. Is that right?

Other than that, this looks ready to merge. We'll ask people to try this out in the beta. And we'll just be ready to revert or adjust in case of any problems in production.

lognaturel:
Well, I had to satisfy my curiosity -- 
```
        final FormIndex indexToDelete = new FormIndex(0, 3, mainInstance.getRoot().getChildAt(4).getRef());
        long currentIterationStart = System.nanoTime();
        formDef.deleteRepeat(indexToDelete);
        double tookMs = (System.nanoTime() - currentIterationStart) / 1000000D;
        System.out.printf("%d\t%.3f\n", 3, tookMs);

        final FormIndex indexToDelete2 = new FormIndex(0, 497, mainInstance.getRoot().getChildAt(4).getRef());
        currentIterationStart = System.nanoTime();
        formDef.deleteRepeat(indexToDelete2);
        tookMs = (System.nanoTime() - currentIterationStart) / 1000000D;
        System.out.printf("%d\t%.3f\n", 497, tookMs);
```
Results:
```
3	787.517
497	1.992
```

So I think what you meant is that things speed up as more repeats are deleted.

And what about adding? `createNewRepeat` runs very fast. Does it not result in `sum` being evaluated?

mdudzinski:
@lognaturel It does but only once for each new repeat. So evaluating calculations which are dependent on a repeat group and target nodes outside of the repeat group should be improved. My initial idea is to run them once after all "inner" calculations are evaluated for all following repeat groups. I can raise an issue for this with description of my initial idea once this one is merged.

lognaturel:
Thank you, that makes sense. Please do file that issue!

mdudzinski:
@ln Done #229 !

PR #220, opened 2017-11-10
Fix 'indexed-repeat' error when index is blank

### LINKED ISSUE #214: Indexed-repeat error when index is blank
#### Software versions 
JavaRosa v2.6.0

#### Problem description
Form and error posted at https://forum.opendatakit.org/t/indexed-repeat-error-for-previously-defined-integer-as-the-third-argument/10442

Read the form above in Collect. The error shows up when the second household member is added.

#### Expected behavior
No error, this should be a valid form.

#### Other information 
The problem seems to be that `index1` initially has no value. I was able to get this to work by defining a default of `1`.

mdudzinski:
Closes #214 

Tested with ODK Collect. A unit test is on the way!

That was the simplest solution. It has one minor drawback - if index is blank, it sets its value to one ('1'). So till the the question referenced by the index isn't answered, it'll keep using the first repeat group for the calculation. 
I think other approaches would require a ridiculous amount of work and refactoring. 

I can't see any risk but let's wait for the unit test.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -946,12 +946,14 @@ public class XPathFuncExpr extends XPathExpression {
             // process index (if valid)
             int groupIdx = toInt(args[idxargi].eval(model, ec)).intValue();
             if (groupIdx <= 0) {
-                // ignore invalid indexes (primarily happens during validation)
+                /*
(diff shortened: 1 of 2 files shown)


mdudzinski:
@lognaturel Added a unit test, it's ready to be reviewed.

lognaturel:
Thanks, @mdudzinski! 

A somewhat strange side effect is that it's now fine to put in index -59384 and that will just give the first name. That seems much better than a crash, certainly.

PR #222, opened 2017-11-16
Index incrementation and decrementation code clean up

mdudzinski:
It doesn't close anything but is related to #218. I was analyzing the codebase related to this issue and got tired of reading it. 
I got rid of the unnecessary verbosity.

Unit tests pass and I also did some quick manual tests

N/A

Nope. It's just a refactoring.

### DIFF
--- a/src/org/javarosa/core/model/FormDef.java
+++ b/src/org/javarosa/core/model/FormDef.java
@@ -1293,12 +1293,6 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
             int ix = indexes.get(i);
             int mult = multiplicities.get(i);
 
-            // ----begin unclear why this is here... side effects???
-            // TODO: ... No words. Just fix it.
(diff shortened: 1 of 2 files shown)


src/org/javarosa/core/model/FormDef.java
```diff
@@ -1283,22 +1283,16 @@ public void collapseIndex(FormIndex index, List<Integer> indexes, List<Integer>
...
 
-        List<IFormElement> curElements = new ArrayList<IFormElement>();
+        List<IFormElement> curElements = new ArrayList<>();
         curElements.addAll(elements);
 
         for (int i = indexes.size() - 1; i >= 0; i--) {
             int ix = indexes.get(i);
             int mult = multiplicities.get(i);
 
-            // ----begin unclear why this is here... side effects???
```
lognaturel: I did a quick double check that these all do appear like simple getters with no side effects.

PR #224, opened 2017-11-17
#218 The repeat group with additional group inside is skipped when I navigate backward

### LINKED ISSUE #218: The repeat group with additionall group inside is skipped when I navigate backward

mdudzinski:
Closes #218

This PR is based on #222 and only the last commit - d7bef68 - is relevant. 
I was digging a lot in the index decrementation code and I found out that something is really odd in the `setRepeatNextMultiplicity(List<IFormElement> elements, List<Integer> multiplicities)` method if a `<group>` node doesn't have the `nodeset` or `ref` attribute. Although it's not mandatory for `<group>` nodes (according to the comment in the code and the [specs](https://opendatakit.github.io/xforms-spec/#groups) to have such attributes, a group with such attribute is considering to be a `logical group`. I think it can be assumed that groups which have representation in the instance, should be `logical` and therefore have either `ref` or `nodeset` attribute. If a reference attribute is present, the decrementation code works as expected. 

Manual tests so far. Unit will be added once this approach is accepted.

This is less invasive solution. I believe there are is no chance to break something. 

Shouldn't be because it only sets the `binding` value of a `GroupDef` instance if the `nodeset` attribute is present in the related `<group>` element during a form parsing.


lognaturel:
I did a rebase and force push to see just the relevant commit. I was a little overzealous and should have let you do that -- forgive me!

Thinking through it...

lognaturel:
Aliasing `nodeset` to `ref` as this done seems good.

This still won't work with the original form @grzesiek2010 posted, right? That form defines the group in the instance but doesn't have a `ref` (or `nodeset`). That doesn't seem like an acceptable combination and should be fixed on the Build end of things, right?

mdudzinski:
@lognaturel no worries :) fixed

Yes, you're right. A group in instance should have either `ref` or `nodeset` attribute in the corresponding element in the body. For now, the `nodeset` attribute on the `<group>` node is completely ignored by the javarosa parser so only `ref` can be used. This change allows to use either of those two. 

I'll also  add a unit test for this change tomorrow.

mdudzinski:
@lognaturel Test added but this branch will have to be rebased once #219 is merged so the review can be postponed.

mdudzinski:
@lognaturel I updated this PR.

PR #228, opened 2017-12-11
PrefixTree improvements and performance instrumentation

dcbriccetti:
This contains a set of IDEA-recommended changes and other small, safe changes, and instrumentation to PrefixTree and related files, to improve the code and help us understand how PrefixTree is used, and its costs.

Please see this for background: https://forum.opendatakit.org/t/collect-large-form-performance/10841

Automatic tests, and lots of manual testing.

These are simple changes.

No.

### DIFF
--- a/src/org/javarosa/core/services/locale/Localizer.java
+++ b/src/org/javarosa/core/services/locale/Localizer.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.javarosa.core.io.Std;
+import org.javarosa.core.util.CodeTimer;
 import org.javarosa.core.util.NoLocalizedTextException;
 import org.javarosa.core.util.OrderedMap;
 import org.javarosa.core.util.PrefixTree;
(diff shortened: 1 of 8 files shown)


lognaturel:
Looks good to me. 

Only thing I'm wondering about is where that timing feedback should go. Ideally it would be logged at level INFO but of course there's no true logging system. As is, it will show up on the console of dependent systems like Briefcase or Collect. Even though this is temporary, we know it's annoying. Even just using the `Std` class introduced by @batkinson instead of `System` would be an improvement so that dependent systems do have the option of redirecting.

dcbriccetti:
We could turn it all off by default. I don’t know if it has value in production.

In a coming-soon branch, I’ve created a new class that does use the Std.

```
/** Helps with timing and operation and logging the time */
public class CodeTimer {
    private final long startTime = System.nanoTime();
    private final String operation;

    /**
     * Creates a CodeTimer for the specified operation
     * @param operation the name of the operation, such as “parsing form”
     */
    public CodeTimer(String operation) {
        this.operation = operation;
    }

    /** Calculates and logs the time since this object was constructed. */
    public void logDone() {
        logDone(Std.out);
    }

    /**
     * Calculates and logs the time since this object was constructed.
     * @param stream the PrintStream onto which to log the message
     */
    public void logDone(PrintStream stream) {
        stream.printf("%s finished in %.3f ms\n", operation, (System.nanoTime() - startTime) / 1e6);
    }
}
```

lognaturel:
Agreed there isn't much use in production. Do you then want to take the timing code out of this PR and merge just the code improvements? Then the timing can be kept in a branch that isn't merged to master. 

If you think it's useful to have in master I think that's not a big deal.

dcbriccetti:
I turned it off by default, but we can enable it from tests with a method call.

lognaturel:
🙏

Issue #229, opened 2017-12-13
Improve the performance of a repeat group deletion


mdudzinski:
#### Software versions 
The latest JavaRosa with fix for #213.

#### Problem description
If a repeat group is deleted then all dependent triggerables will be evaluated for each of the following repeats after the deleted one. This behavior was added in #213 / #219. 
If at least one of the dependent calculations targets a node outside of the repeat group and takes a repeat group children nodeset as an argument, then running such calculation for each of the following repeats can cause performance problems if there are a lot of the following repeats to process. More details can be found in the conversation under #219.
Example of such functions:
* [sum](https://opendatakit.github.io/xforms-spec/#fn:sum)
* [concat](https://opendatakit.github.io/xforms-spec/#fn:concat)

#### Steps to reproduce the problem
The easiest way is to launch `Safe2014DagImplTest#deleteRepeatGroupWithCalculationsTimingTest` with an increased value of `final int numberOfRepeats` constant. It start being significant for values above `500`.

#### Expected behavior
Deletion of a repeat group should be a matter of millisecond regardless of the related calculation and total amount of the repeats.

#### Other information 
It can be either fixed by running calculations outside of the repeat group only once - after all "inner" calculations for the following repeats are evaluated - or by fixing - if feasible - the bottleneck in `org.javarosa.core.model.instance.DataInstance.resolveReference` / `org.javarosa.xpath.expr.XPathPathExpr.eval` methods.
<img width="990" alt="zrzut ekranu 2017-12-13 o 14 12 21" src="https://user-images.githubusercontent.com/1262799/33940689-42c45c72-e010-11e7-8879-329a1495b706.png">
<img width="982" alt="zrzut ekranu 2017-12-13 o 14 25 36" src="https://user-images.githubusercontent.com/1262799/33941073-9a6c620c-e011-11e7-888a-cae00d7617b2.png">
<img width="1070" alt="zrzut ekranu 2017-12-13 o 14 25 57" src="https://user-images.githubusercontent.com/1262799/33941081-9e10aa3a-e011-11e7-9395-7708d83bd4a9.png">

mdudzinski:
@lognaturel I spent some more time on this and prepared a PoC of a simple approach that doesn't a lot of changes or/and refactoring -> #230

dcbriccetti:
[...]

PR #230, opened 2017-12-13
WIP: Fix a repeat group deletion performance Proof of Concept

mdudzinski:
Just a proof of concept of how #229 could be fixed. 

This approach leverages existing interfaces of Dag methods for triggerable evaluation. It only needs to expose - for readonly - a one field from `Triggerable` class.

Event assertions have been commented off. They changed and updating them takes a bit of time :) But assertions for actual evaluation results pass. 

The code is dirty but shows the approach - postpone evaluation for all calculations targeting nodes outside of the repeat. I assume this shouldn't introduce any regression because it's how it actually worked before #219  ...

### DIFF
--- a/src/org/javarosa/core/model/Safe2014DagImpl.java
+++ b/src/org/javarosa/core/model/Safe2014DagImpl.java
@@ -56,6 +56,15 @@ public class Safe2014DagImpl extends LatestDagBase {
       return conditionRepeatTargetIndex.get(repeatRef.genericize());
    }
 
+   private void collectExcluded(TreeReference trigger, QuickTriggerable cascade, Set<QuickTriggerable> exclude) {
+      if (!trigger.isParentOf(cascade.t.getContextRef(), true)) {
+         exclude.add(cascade);
+      }
+      for (QuickTriggerable nextCascade : cascade.t.getImmediateCascades()) {
+         collectExcluded(trigger, nextCascade, exclude);
+      }
+   }
+
    @Override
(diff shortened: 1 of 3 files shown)


dcbriccetti:
This looks promising. I didn’t compare the results with those before this change. Is there much improvement? Is the largest factor the TreeElement.getChild slowness I mentioned in https://github.com/opendatakit/javarosa/issues/229 ?

As I was studying the code I made some changes to make it easier for me to understand. Feel free to use or ignore. I’ve seen a lot of code in my day that reassigns values to identifiers (often increasing the difficulty of understanding), so except where the scope is just a few lines, I really benefit from seeing `final`.

screen shot 2017-12-24 at 12 59 45 pm

mdudzinski:
@dcbriccetti I'm always for final identifiers so I'll definitely use it! I try to do such refactoring as well when I've got time.

Regarding the performance - yes there is a huge improvement.
For the `deleteRepeatGroupWithCalculationsTimingTest()` and 1000 of repeats it now starts with about 140 milliseconds when the original approach starts with about 4 seconds. So this is a huge and should be enough but `TreeElement.getChild` thing can be improved anyway as other parts of code may benefit from it.

lognaturel:
That's a significant speed increase, @mdudzinski! 👏 

The recursive approach seems like a good one. For defensive purposes, how about returning a clone of the context ref instead of the object itself? Otherwise this seems like a good thing to clean up and integrate.

lognaturel:
Replaced by #583. Thanks again, @mdudzinski!

PR #235, opened 2017-12-28
WIP: Improve the performance optimizations for TreeElement children for more cases

### LINKED ISSUE #234: Improve the performance optimizations for TreeElement children for more cases
https://github.com/opendatakit/javarosa/commit/09034d63849eb0a25b7259aa5144a6de09b0495a increased the speed of TreeElement child handling under a special case, where “all children have the same name, and none have a special multiplicity value (< 0)”. 

Improve performance for more child patterns, such as the case (https://github.com/opendatakit/javarosa/issues/229#issuecomment-353801900) where the groups of same-named children are preceded by an element with multiplicity of TreeReference.INDEX_TEMPLATE (-2), and followed by siblings with different names.

dcbriccetti:
### Work in Progress

Partially implement an indexing scheme to improve the performance optimizations for TreeElement children for more cases (https://github.com/opendatakit/javarosa/issues/234).

Closes #234

### DIFF
--- a/src/org/javarosa/core/model/FormDef.java
+++ b/src/org/javarosa/core/model/FormDef.java
@@ -465,20 +465,7 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
         TreeReference parentRef = deleteRef.getParentRef();
         TreeElement parentElement = mainInstance.resolveReference(parentRef);
 
-        int childMult = deleteElement.getMult();
-        parentElement.removeChild(deleteElement);
-
-        // update multiplicities of other child nodes
-        for (int i = 0; i < parentElement.getNumChildren(); i++) {
-            TreeElement child = parentElement.getChildAt(i);
-            if (child.getName().equals(deleteElement.getName()) && child.getMult() > childMult) {
-                child.setMult(child.getMult() - 1);
-
-                // When a group changes it multiplicity, then all caches of its
-                // children must clear.
-                child.clearChildrenCaches();
-            }
-        }
+        parentElement.removeChildAndAdjustSiblingMults(deleteElement);
 
         dagImpl.deleteRepeatGroup(getMainInstance(), getEvaluationContext(), deleteRef, parentElement, deleteElement);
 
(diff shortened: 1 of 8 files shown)


dcbriccetti:
@mdudzinski Hey, maybe take a quick look for any major flaws with this scheme? Or let me know what I should document better to help acquaint readers with it? Thanks.

lognaturel:
@dcbriccetti I believe this is targeting removal of repeats, and currently I think that's all it speeds up, is that right? Would `NameToNameIndex` be able to speed any other cases up in the future?

dcbriccetti:
Hi @lognaturel!
>speed any other cases up

Are there cases with many child TreeElements where the volume isn’t from repeats? We could use NameToNameIndex (maybe with a more general name) to directly access the child element being sought, instead of the current sequential search.

lognaturel:
> Are there cases with many child TreeElements where the volume isn’t from repeats?

Would a case where a form has multiple groups for structure benefit from this kind of optimization? For example, I have 40 questions in a "Demographics" group, 30 questions in a "XXX" group, etc.

Does it/can it help for things other than removal?

How significant is the speed improvement? I'm wondering because this does add some complexity so the payoff has to be worthwhile!

lognaturel:
PS Happy New Year, @dcbriccetti 😄🎊

dcbriccetti:
It improves reads, insertions, and deletions from O(n²) to O(n log n). 

It operates at the parent TreeElement node level (any TreeElement with children). So in the case you described, the Demographics group TreeElement could be found quickly from its parent, and its children could be found quickly by name.

I was thinking of making some diagrams.

lognaturel:
Ah! Ok, that sounds like it could be significant. Yes, a diagram would help, especially a little bit saying what exactly is O(n²) currently. Evidently I haven't done a deep enough dive.

lognaturel:
This is now really out of sync with `master`. My quick analysis suggests it would help with performance when repeats are involved but it comes with a significant complexity increase so I don't think it's worth it.

PR #238, opened 2018-01-17
bug #236: add attributes to appropriate whitelists

### LINKED ISSUE #236: Warning displayed when rows attribute is used on an input node.
#### Problem description
Warning displayed when `rows` attribute is used on an `input` node.

Add the `rows` attribute to a text field:
```
        <input ref="/data/comments" rows="2">
          <label>Comments</label>
        </input>
```
This is recognized and used by Collect and perhaps other clients but the following warning is shown using Validate:
```
Problem found at nodeset: 
    With element <input ref="/data/comments" rows="2"><label>
```

#### Expected behavior
No warning.

#### Other information 
I guess we should specifically allow `rows`. Another option would be to allow any attribute but that might be less helpful.

daveycrockett:
- `saveIncomplete` and `requiredMsg` added to allowed bind node attribs
- `rows` added to allowed input attributes

Closes #236 

Added a unit test that parses an example form with the additional attributes

Follows existing conventions

None that I can see!

### DIFF
--- a/test/org/javarosa/xform/parse/XFormParserTest.java
+++ b/test/org/javarosa/xform/parse/XFormParserTest.java
@@ -386,6 +386,11 @@ public class XFormParserTest {
         assertThat(groupElement.getBind(), is(expectedXPathReference));
     }
 
+    @Test public void parseWithWhitelistedAttributes() throws IOException {
+      ParseResult parseResult = parse(r("whitelisted-attributes.xml"));
+      assertThat(parseResult.errorMessages.size(), is(0));
+    }
+
     private TreeElement findDepthFirst(TreeElement parent, String name) {
         int len = parent.getNumChildren();
         for (int i = 0; i < len; ++i) {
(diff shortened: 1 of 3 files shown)


test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -386,6 +386,11 @@ public void parseFormWithSetValueAction() throws IOException {
         assertThat(groupElement.getBind(), is(expectedXPathReference));
     }
 
+    @Test public void parseWithWhitelistedAttributes() throws IOException {
```
dcbriccetti: When I created this class last April,

https://github.com/dcbriccetti/javarosa/commit/ab7517e984a34bedf82ddfe15507d7e4dc3381f2

I established a method naming convention based on the indicative mood (“parses…”, rather than “parse…”), in order to express what each method tests. I carried on like this for months until @mdudzinski joined in and started littering the code with imperative mood method names. :wink: I think he is not a native English speaker so he deserves some leeway. Now, @daveycrockett, you must choose which of us you wish to align with. :grin:

daveycrockett:
Eek! I tend to follow "when in Rome," but this was subtle enough that I didn't notice the other method names! That's fixed (with a commit message also using the imperative aspect, just to be safe).

lognaturel:
Thanks for doing this @daveycrockett!

PR #241, opened 2018-01-23
Remove PrefixTree

dcbriccetti:
Hélène summarized this nicely here: https://github.com/opendatakit/javarosa/pull/241#issuecomment-360935062

And we’ve discussed it here:
See https://forum.opendatakit.org/t/collect-large-form-performance/10841/22?u=dcbriccetti

PrefixTree stores Strings in PrefixTreeNodes. Replacing PrefixTrees with List<String> and replacing PrefixTreeNodes with Strings was a simple operation. All tests pass. Downloading, loading from XML and loading from the cache all appear to work correctly.

It removes code and lowers complexity. No.

No.


dcbriccetti:
comparison of ram use with and without prefixtree

lognaturel:
Let me try to summarize!
- `PrefixTree` was added to reduce memory usage for forms that have a lot of translations that start with the same text. It takes a long time to build up on form load.
- Removing `PrefixTree` for the eIMCI and Nigeria Wards forms improves load time significantly (at least 20x) with at worst ~13% increase in memory usage.
- The kind of form removing `PrefixTree` would significantly impact is one with many choices that all start with the same text. That doesn't seem terribly common.

The decision we need to make now is whether to ship this change directly after a beta period or to hide it behind a feature toggle and gather feedback over a longer period. I think we've pretty clearly identified the kinds of forms this would impact negatively and they don't seem a likely case. We can clearly identify the type of form that needs testing and ask users to try them during the beta. If someone does report that forms they used to be able to load can't be loaded anymore we can decide what to do. For example, we might want to do a reverse toggle for "low-memory mode" or something like it.

We can leave this open for comment for a few days and if there are no objections, merge.

lognaturel:
I am satisfied that this is safe enough. In the beta notes we will specifically say that people using forms with many choices on old phones should give the update a try.

PR #247, opened 2018-02-16
Issue 164 add digest function

### LINKED ISSUE #164: Add digest function
opendatakit/xforms-spec#103 (use [this spec](https://www.w3.org/TR/xforms/#fn-digest))

ggalmazor:
Closes #164 

This PR adds a minimal implementation of the `digest()` function, defined in the [XForms 1.1 specs](https://www.w3.org/TR/xforms/#fn-digest).

- It supports the required MD5, SHA-1, and SHA-256 hashing algorithms
- It supports hexadecimal and base64 hash encoding methods

Create new test cases for the `digest()` function in `XPathEvalTest`, including specific examples from the XForms 1.1 specs.

I didn't create failure tests (passing an unsupported hashing algorithm or encoding method).

This is a straightforward implementation, using basic Java7 features.

None that I'm aware of.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -470,7 +470,7 @@ public class XPathFuncExpr extends XPathExpression {
             try {
                 digest = MessageDigest.getInstance((String) argVals[1]);
             } catch (NoSuchAlgorithmException e) {
-                throw new XPathTypeMismatchException("The function digest received a non supported cryptographic hashing algorithm '"+ argVals[0] +"'. Valid values are MD5, SHA-1, SHA-256");
+                throw new XPathTypeMismatchException("The function digest received a non supported cryptographic hashing algorithm '" + argVals[1] + "'. Valid values are MD5, SHA-1, SHA-256");
             }
 
             // Use UTF-8 encoding when reading data to be hashed by default
@@ -980,7 +980,7 @@ public class XPathFuncExpr extends XPathExpression {
                  */
                 groupIdx = 1;
             }
-            
+
             contextRef.setMultiplicity(groupRef.size()-1, groupIdx-1);
         }
 
@@ -1388,7 +1388,7 @@ public class XPathFuncExpr extends XPathExpression {
             for(HashEncodingMethod candidate : values())
                 if (candidate.value.equals(value))
                     return candidate;
-            throw new XPathTypeMismatchException("Unsupported hash encoding algorithm \"\". Supported values are: \"hex\", \"base64\"");
+            throw new XPathTypeMismatchException("Unsupported hash encoding algorithm \"" + value + "\". Supported values are: \"hex\", \"base64\"");
         }
 
         abstract String encode(byte[] bytes);
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -16,10 +16,6 @@
 
 package org.javarosa.xpath.expr;
 
-import java.io.UnsupportedEncodingException;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import javax.xml.bind.DatatypeConverter;
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IFunctionHandler;
 import org.javarosa.core.model.condition.pivot.UnpivotableExpressionException;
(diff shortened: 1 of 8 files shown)


ggalmazor:
By the way, I don't know who usually does revisions... I've added @dcbriccetti and @yanokwa but feel free to ignore :)

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -461,6 +465,23 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
             }
 
             return GeoUtils.calculateAreaOfGPSPolygonOnEarthInSquareMeters(gpsCoordinatesList);
+        } else if (name.equals("digest") && (args.length == 3)) {
```
lognaturel: The third parameter should be optional and default to `base64`

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -461,6 +465,23 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
...
+            MessageDigest digest;
+            try {
+                digest = MessageDigest.getInstance((String) argVals[1]);
+            } catch (NoSuchAlgorithmException e) {
+                throw new XPathTypeMismatchException("The function digest received a non supported cryptographic hashing algorithm '"+ argVals[0] +"'. Valid values are MD5, SHA-1, SHA-256");
```
lognaturel: The error should use `argVals[1]`

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1335,4 +1356,37 @@ public Object pivot (DataInstance model, EvaluationContext evalContext, List<Obj
...
+
+        HashEncodingMethod(String value) {
+            this.value = value;
+        }
+
+        static HashEncodingMethod from(String value) {
+            for(HashEncodingMethod candidate : values())
+                if (candidate.value.equals(value))
+                    return candidate;
+            throw new XPathTypeMismatchException("Unsupported hash encoding algorithm \"\". Supported values are: \"hex\", \"base64\"");
```
lognaturel: Did you mean to put `value` in those parens?

ggalmazor:
Thanks, @lognaturel!

ggalmazor:
By the way, the hashing method I've used works well for small payloads. I haven't been able to find any reference on how much ceases to be small and starts to be big.

For big payloads, it's better to:

```java
MessageDigest md = MessageDigest.getInstance("MD5");
md.update(payload.getBytes());
byte[] digest = md.digest();
```

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1335,4 +1360,37 @@ public Object pivot (DataInstance model, EvaluationContext evalContext, List<Obj
...
     }
+
+    enum HashEncodingMethod {
+        HEX("hex") {
+            @Override
+            String encode(byte[] bytes) {
+                StringBuilder stringBuffer = new StringBuilder();
+                for (byte aResult : bytes)
+                    stringBuffer.append(Integer.toString((aResult & 0xff) + 0x100, 16).substring(1));
+                return stringBuffer.toString();
```
dcbriccetti: Did you consider using the DatatypeConverter for hex as well as base64? The tests pass with this:
`return DatatypeConverter.printHexBinary(bytes).toLowerCase();`

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1335,4 +1360,37 @@ public Object pivot (DataInstance model, EvaluationContext evalContext, List<Obj
...
+
+        HashEncodingMethod(String value) {
+            this.value = value;
+        }
+
+        static HashEncodingMethod from(String value) {
+            for(HashEncodingMethod candidate : values())
+                if (candidate.value.equals(value))
+                    return candidate;
+            throw new XPathTypeMismatchException("Unsupported hash encoding algorithm \"" + value + "\". Supported values are: \"hex\", \"base64\"");
```
dcbriccetti: Can you avoid reimplementing Enum.valueOf? Maybe like this?
```
try {
    return valueOf(value.toUpperCase());
} catch (IllegalArgumentException ex) {
    throw new XPathTypeMismatchException("Unsupported hash encoding algorithm \"" + value + "\". Supported values are: \"hex\", \"base64\"");
}
```

ggalmazor: The rationale was twofold:
- Hide the generation and error handling inside the type
- Avoid coupling between the enum name and the actual algorithm name

src/org/javarosa/xpath/expr/XPathFuncExpr.java
ggalmazor: Using a javax api felt a little hacky, but after your comment, it's probably ok :)

ggalmazor:
@dcbriccetti I've taken much from your branch. Thanks for your review!!

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -466,26 +462,10 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
...
-            }
-
-            // Hash encoding method defaults to base64 if no third arg is received
-            HashEncodingMethod encodingMethod = args.length == 3
-                ? HashEncodingMethod.from((String) argVals[2])
-                : HashEncodingMethod.BASE64;
-            return encodingMethod.encode(result);
+            return DigestAlgorithm.from((String) argVals[1]).digest(
+                (String) argVals[0],
+                args.length == 3 ? Encoding.from((String)argVals[2]) : Encoding.BASE64
```
dcbriccetti: Space after the cast

PR #249, opened 2018-02-17
Issue 36 constraint msg output

### LINKED ISSUE #36: Output doesn't work in a text used as a jr:constraintMsg
> Collect v1.5.1, on a Nexus 7 running Android 5.0.2
> 
> I created a form where calculate fields are used to create both a value used for a constraint and text to use in the constraint message. However, while the constraint itself seems to work correctly when I view the form on my tablet, the constraint message doesn't - instead of displaying my text from the calculate field, it displays e.g. "${0}". This same text works when I use it in a note instead. I've attached a sample form where this is reproduced.
> 
> Thanks,
> Laura

[constraintMessageError.xlsx](https://github.com/opendatakit/javarosa/files/919784/constraintMessageError.xlsx)

ggalmazor:
Closes #36

Thir PR changes `FormEntryPrompt` to mirror what's done in `FormEntryCaption.getQuestionText(String textID)` to substitute variables inside a constraint text.

TDD'ed the development with a new `OutputInComputedConstraintTextTest` test class that covers the change.

It is the simplest solution I've been able to come up with.

None that I know of.

### DIFF
--- a/test/org/javarosa/core/form/api/test/OutputInComputedConstraintTextTest.java
+++ b/test/org/javarosa/core/form/api/test/OutputInComputedConstraintTextTest.java
@@ -1,0 +1,86 @@
+package org.javarosa.core.form.api.test;
+
+import static org.junit.Assert.assertEquals;
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+import org.javarosa.core.PathConst;
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.FormIndex;
+import org.javarosa.core.model.IFormElement;
+import org.javarosa.core.model.QuestionDef;
+import org.javarosa.core.model.data.LongData;
+import org.javarosa.core.services.PrototypeManager;
+import org.javarosa.core.test.FormParseInit;
+import org.javarosa.form.api.FormEntryCaption;
+import org.javarosa.form.api.FormEntryController;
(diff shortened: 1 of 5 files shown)


test/org/javarosa/core/form/api/test/OutputInComputedConstraintTextTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private Map<String, FormIndex> formIndexesById = new HashMap<>();
+  private FormDef formDef;
+  private FormEntryController ctrl;
+  private FormEntryModel model;
+
+  @Before
+  public void setUp() {
+    FormParseInit fpi = new FormParseInit();
+    fpi.setFormToParse((new File(PathConst.getTestResourcePath(), "constraint-message-error.xml")).getAbsolutePath());
```
dcbriccetti: We decided some months back that new resource files used by tests would be placed into a subdirectory matching the package name. `resources/org/javarosa/core/form/api/test/constraint-message-error.xml` in this case.

You can use `r` to specify this path automatically.

`import static org.javarosa.test.utils.ResourcePathHelper.r;`
`fpi.setFormToParse(r("constraint-message-error.xml").toString());`

ggalmazor: Alright! Do you think the test should be in that package?

PR #257, opened 2018-02-19
Issue 256 Remove javax

### LINKED ISSUE #256: Remove use of javax api not included in Android
#### Software versions 
JavaRosa v1.8.0

#### Problem description
#247 uses some classes from `javax` that are not included in Android. Unfortunately that makes JR unusable so an alternative will need to be considered.

CC @ggalmazor @dcbriccetti

ggalmazor:
Closes #256 

This PR removes the dependency on javax components for encoding to hexadecimal and base64.

Ran the automated tests that were already covering the implementation (Refactoring, yay!)

There are tons of implementations of encoders for Java7. I've chosen the ones that were advertised as being fast and efficient.

None

### DIFF
--- a/src/org/javarosa/xpath/expr/Encoding.java
+++ b/src/org/javarosa/xpath/expr/Encoding.java
@@ -1,6 +1,6 @@
 package org.javarosa.xpath.expr;
 
-import javax.xml.bind.DatatypeConverter;
+import java.io.UnsupportedEncodingException;
 import org.javarosa.xpath.XPathUnsupportedException;
 
 /**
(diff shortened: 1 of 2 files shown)


src/org/javarosa/xpath/expr/Encoding.java
```diff
@@ -10,13 +10,52 @@
...
+      }
+      return sb.toString();
     }
   },
   BASE64("base64") {
     @Override
-    String encode(byte[] bytes) {
-      return DatatypeConverter.printBase64Binary(bytes);
+    String encode(byte[] sArr) {
+      // Copied from: https://github.com/brsanthu/migbase64/blob/master/src/main/java/com/migcomponents/migbase64/Base64.java
```
dcbriccetti: What a shame we have to include all this bit twiddling. I don’t have a better option.

ggalmazor: Eh :/

PR #258, opened 2018-02-21
Start and end labels were added for range question types

andualem:
This feature was required for PMA2020 project. This pull request is made in case similar change is required by anyone.

### DIFF
--- a/src/org/javarosa/core/model/RangeQuestion.java
+++ b/src/org/javarosa/core/model/RangeQuestion.java
@@ -8,6 +8,8 @@ public class RangeQuestion extends QuestionDef {
     private BigDecimal rangeStart;
     private BigDecimal rangeEnd;
     private BigDecimal rangeStep;
+    private String rangeStartLabel;
+    private String rangeEndLabel;
 
     public void setRangeStart(BigDecimal rangeStart) {
         this.rangeStart = rangeStart;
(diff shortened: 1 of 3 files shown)


lognaturel:
Thanks, @andualem! This will require approval at the spec level and corresponding changes to Collect. I'll mark as `needs discussion` pending those conversations.

src/org/javarosa/core/model/RangeQuestion.java
```diff
@@ -32,4 +34,20 @@ public void setRangeStep(BigDecimal rangeStep) {
     public BigDecimal getRangeStep() {
         return rangeStep;
     }
+
+    public String getRangeStartLabel() {
+        return rangeStartLabel;
```
grzesiek2010: Maybe it should return `rangeStart` if rangeStartLabel is not specified? It would be easier to use in Collect then.

src/org/javarosa/core/model/RangeQuestion.java
```diff
@@ -32,4 +34,20 @@ public void setRangeStep(BigDecimal rangeStep) {
...
+    }
+
+    public void setRangeStartLabel(String rangeStartLabel) {
+        this.rangeStartLabel = rangeStartLabel;
+    }
+
+    public String getRangeEndLabel() {
+        return rangeEndLabel;
```
grzesiek2010: Similar as above.

yanokwa:
@andualem We've let this sit for too long without a decision, so apologies for that. I'm closing this PR because it'll require spec changes and some user discussion. I think we do want to add something like this, so please file a feature request at https://forum.opendatakit.org/c/features so we can have that discussion.

yanokwa:
Feature request has been filed at https://forum.opendatakit.org/t/slider-widget-with-text-as-begin-and-end-label/17144/ for discussion.

PR #259, opened 2018-02-22
Issue 152 slf4j migration

### LINKED ISSUE #152: Implement logging strategy more useful to tools that use the lib
In #150, @batkinson described some of the problems with with printing to `System.out` and `System.err`. #151 gives clients of the library an opportunity to direct output to other streams. It would be even better to implement logging so that clients can pick the level of logging to see, where to the direct the logging and so on.

Some thought needs to be put into which logging library or bridge to use. I don't have a sophisticated enough understanding of the tradeoffs to decide. Briefcase uses the Apache Commons Logging [bridge](https://commons.apache.org/logging/). There would be some convenience in those matching, I think. Other options include straight `java.util.logging`, [SLF4J](https://www.slf4j.org). Maybe [LogBack](https://logback.qos.ch/) is the modern choice. @dcbriccetti, hoping you can help make an informed decision here.

ggalmazor:
Closes #152 

This PR adds SLF4J API to the project and Logback binding only for the tests.

All the code that was directly or indirectly using `System.out` and `System.err` has been migrated to SLF4J.

As a consequence, first, deprecation flags have been added to some public artifacts. Then any unused deprecated artifact has been removed. This is also discussed in #52:
- `org.javarosa.core.api.ILogger`
- `org.javarosa.core.io.Std`
- `org.javarosa.core.io.StreamsUtil`
- `org.javarosa.core.services.Logger`
- `org.javarosa.core.util.CodeTimer`
- `org.javarosa.xform.parse.XFormParserReporter`
  - There's a coupling issue with this class. It's used by `FormParserHelper` for logging **and** error handling:

    ```java
    public static ParseResult parse(Path formName) throws IOException {
        XFormParser parser = new XFormParser(new FileReader(formName.toString()));
        final List<String> errorMessages = new ArrayList<>();
        parser.attachReporter(new XFormParserReporter() {
            @Override
            public void warning(String type, String message, String xmlLocation) {
                errorMessages.add(message);
                super.warning(type, message, xmlLocation);
            }
    
            @Override
            public void error(String message) {
                errorMessages.add(message);
                super.error(message);
            }
        });
    
        return new ParseResult(parser.parse(), errorMessages);
    }
    ```

    This is the only special use of `XFormParserReporter` that requires a change on the API. The solution I propose is to add into `XFormParser` a pair of `onWarning` and `onError` methods:

    ```java
    public static ParseResult parse(Path formName) throws IOException {
        XFormParser parser = new XFormParser(new FileReader(formName.toString()));
        final List<String> errorMessages = new ArrayList<>();
        parser.onWarning(new XFormParser.WarningCallback() {
            @Override
            public void accept(String type, String message, String xmlLocation) {
                errorMessages.add(message);
            }
        });
        parser.onError(new XFormParser.ErrorCallback() {
            @Override
            public void accept(String message) {
                errorMessages.add(message);
            }
        });
        return new ParseResult(parser.parse(), errorMessages);
    }
    ```
Compiled project and ran automated tests.

I tried to do the minimum amount of changes to do a meaningful migration to slf4j

Removal of public artifacts has to be explicitly documented on any release note of a JR release that includes this PR

Any error introduced by migrating manual log message variable interpolation to SLF4J's `{}` placeholder syntax. I manually reviewed this three times but there's a risk here of losing information on logs or getting unexpected changes on current log lines.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
--- a/src/org/javarosa/core/api/ILogger.java
+++ b/src/org/javarosa/core/api/ILogger.java
@@ -15,38 +15,71 @@
  */
 
 /**
- * 
+ *
  */
 package org.javarosa.core.api;
 
+import java.io.IOException;
+import java.util.Date;
 import org.javarosa.core.log.IFullLogSerializer;
 import org.javarosa.core.log.StreamLogSerializer;
 
-import java.io.IOException;
-import java.util.Date;
-
 /**
+ * <b>Warning:</b> This class is unused and should remain that way. It will be removed in a future release.
+ *
  * IIncidentLogger's are used for instrumenting applications to identify usage
  * patterns, usability errors, and general trajectories through applications.
- * 
- * @author Clayton Sims
- * @date Apr 10, 2009 
  *
+ * @author Clayton Sims
+ * @deprecated Use {@link org.slf4j.LoggerFactory#getLogger(Class)} instead
  */
+@Deprecated
 public interface ILogger {
-
+    /**
+     * @deprecated Use {@link org.slf4j.Logger#info} instead
+     */
+    @Deprecated
     public void log(String type, String message, Date logDate);
 
+    /**
+     * @deprecated Use {@link org.slf4j.Logger} instead
+     */
+    @Deprecated
     public void clearLogs();
 
+    /**
+     * @deprecated Use {@link org.slf4j.Logger} instead
+     */
+    @Deprecated
     public <T> T serializeLogs(IFullLogSerializer<T> serializer);
 
+    /**
+     * @deprecated Use {@link org.slf4j.Logger} instead
+     */
+    @Deprecated
     public void serializeLogs(StreamLogSerializer serializer) throws IOException;
+
+    /**
+     * @deprecated Use {@link org.slf4j.Logger} instead
+     */
+    @Deprecated
     public void serializeLogs(StreamLogSerializer serializer, int limit) throws IOException;
 
+    /**
+     * @deprecated Use {@link org.slf4j.Logger} instead
+     */
+    @Deprecated
     public void panic();
 
+    /**
+     * @deprecated Use {@link org.slf4j.Logger} instead
+     */
+    @Deprecated
     public int logSize();
 
+    /**
+     * @deprecated Use {@link org.slf4j.Logger} instead
+     */
+    @Deprecated
     public void halt();
 }
--- a/src/org/javarosa/core/io/Std.java
+++ b/src/org/javarosa/core/io/Std.java
@@ -3,16 +3,24 @@ package org.javarosa.core.io;
 import java.io.PrintStream;
 
 /**
+ * <b>Warning:</b> This class is unused and should remain that way. It will be removed in a future release.
  * Used to break dependency directly on System.out and System.err.
+ *
+ * @deprecated Use {@link org.slf4j.LoggerFactory#getLogger(Class)} instead
  */
+@Deprecated
 public class Std {
 
+    @Deprecated
     public static PrintStream out = System.out;
+    @Deprecated
     public static PrintStream err = System.err;
 
     /**
      * Use this to replace stdout in case anything more than assignment is necessary in the future.
+     *
      * @param out
+     * @deprecated Use {@link org.slf4j.LoggerFactory#getLogger(Class)} instead
      */
     public static void setOut(PrintStream out) {
         Std.out = out;
@@ -20,12 +28,17 @@ public class Std {
 
     /**
      * Use this to replace stderr in case anything more than assignment is necessary in the future.
+     *
      * @param err
+     * @deprecated Use {@link org.slf4j.LoggerFactory#getLogger(Class)} instead
      */
     public static void setErr(PrintStream err) {
         Std.err = err;
     }
 
+    /**
+     * @deprecated Use {@link org.slf4j.Logger#error(String, Throwable)} instead
+     */
     public static void printStack(Throwable t) {
         t.printStackTrace(Std.err);
     }
--- a/src/org/javarosa/core/io/StreamsUtil.java
+++ b/src/org/javarosa/core/io/StreamsUtil.java
@@ -7,23 +7,22 @@ import java.io.OutputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * <b>Warning</b>: This class is unused and should remain that way. It will be removed in a future release.
+ *
+ * @deprecated
+ */
+@Deprecated
 public class StreamsUtil {
-    private static final Logger logger = LoggerFactory.getLogger(StreamsUtil.class);
-
     private StreamsUtil() {
-        // private constructor
     }
 
     /**
      *
      * Write everything from input stream to output stream, byte by byte then
      * close the streams
-     *
-     *
-     * @param in
-     * @param out
-     * @throws IOException
      */
+    @Deprecated
     public static void writeFromInputToOutput(InputStream in, OutputStream out, long[] tally) throws InputIOException, OutputIOException {
         //TODO: God this is naive
         int val;
@@ -47,10 +46,12 @@ public class StreamsUtil {
         }
     }
 
+    @Deprecated
     public static void writeFromInputToOutputSpecific(InputStream in, OutputStream out) throws InputIOException, OutputIOException {
         writeFromInputToOutput(in, out, null);
     }
 
+    @Deprecated
     public static void writeFromInputToOutput(InputStream in, OutputStream out) throws IOException {
         try {
             writeFromInputToOutput(in, out, null);
@@ -64,13 +65,9 @@ public class StreamsUtil {
     private static final int CHUNK_SIZE = 2048;
 
     /**
-     *
      * Write the byte array to the output stream
-     *
-     * @param bytes
-     * @param out
-     * @throws IOException
      */
+    @Deprecated
     public static void writeToOutput(byte[] bytes, OutputStream out, long[] tally) throws IOException {
         int offset = 0;
         int remain = bytes.length;
@@ -86,6 +83,7 @@ public class StreamsUtil {
         }
     }
 
+    @Deprecated
     public static void writeToOutput(byte[] bytes, OutputStream out) throws IOException {
         writeToOutput(bytes, out, null);
     }
@@ -97,15 +95,10 @@ public class StreamsUtil {
     }
 
     /**
-     *
      * Read bytes from an input stream into a byte array then close the input
      * stream
-     *
-     * @param in
-     * @param len
-     * @return
-     * @throws IOException
      */
+    @Deprecated
     public static byte[] readFromStream(InputStream in, int len)
             throws IOException {
 
@@ -152,11 +145,13 @@ public class StreamsUtil {
 
         IOException internal;
 
+        @Deprecated
         public DirectionalIOException(IOException internal) {
             super(internal.getMessage());
             this.internal = internal;
         }
 
+        @Deprecated
         public IOException getWrapped() {
             return internal;
         }
@@ -164,23 +159,21 @@ public class StreamsUtil {
         //TODO: Override all common methodss
     }
 
+    @Deprecated
     public class InputIOException extends DirectionalIOException{
-        /**
-         *
-         */
         private static final long serialVersionUID = 5939766950738216779L;
 
+        @Deprecated
         public InputIOException(IOException internal) {
             super(internal);
         }
     }
 
+    @Deprecated
     public class OutputIOException extends DirectionalIOException{
-        /**
-         *
-         */
         private static final long serialVersionUID = -1816322555490749440L;
 
+        @Deprecated
         public OutputIOException(IOException internal) {
             super(internal);
         }
--- a/src/org/javarosa/core/io/StreamsUtil.java
+++ b/src/org/javarosa/core/io/StreamsUtil.java
@@ -4,8 +4,6 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * <b>Warning</b>: This class is unused and should remain that way. It will be removed in a future release.
--- a/src/org/javarosa/core/io/StreamsUtil.java
+++ b/src/org/javarosa/core/io/StreamsUtil.java
@@ -4,8 +4,11 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class StreamsUtil {
+    private static final Logger logger = LoggerFactory.getLogger(StreamsUtil.class);
 
     private StreamsUtil() {
         // private constructor
@@ -156,10 +159,6 @@ public class StreamsUtil {
 
         public IOException getWrapped() {
             return internal;
-        }
-
-        public void printStackTrace()  {
-            Std.printStack(internal);
         }
 
         //TODO: Override all common methodss
--- a/src/org/javarosa/core/services/ProgramFlow.java
+++ b/src/org/javarosa/core/services/ProgramFlow.java
@@ -1,0 +1,32 @@
+package org.javarosa.core.services;
+
+import org.javarosa.core.log.FatalException;
+import org.slf4j.LoggerFactory;
+
+public final class ProgramFlow {
+    private static final org.slf4j.Logger logger = LoggerFactory.getLogger(ProgramFlow.class);
+
+    public static void die(String thread, Exception e) {
+        //log exception
+        logger.error("unhandled exception at top level", e);
+
+        //crash
+        final FatalException crashException = new FatalException("unhandled exception in " + thread, e);
+
+        //depending on how the code was invoked, a straight 'throw' won't always reliably crash the app
+        //throwing in a thread should work (at least on our nokias)
+        new Thread() {
+            public void run() {
+                throw crashException;
+            }
+        }.start();
+
+        //still do plain throw as a fallback
+        try {
+            Thread.sleep(3000);
+        } catch (InterruptedException ie) {
+        }
+        throw crashException;
+    }
+}
+
--- a/src/org/javarosa/core/services/PropertyManager.java
+++ b/src/org/javarosa/core/services/PropertyManager.java
@@ -20,13 +20,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.NoSuchElementException;
 
-import org.javarosa.core.io.Std;
 import org.javarosa.core.services.properties.IPropertyRules;
 import org.javarosa.core.services.properties.Property;
 import org.javarosa.core.services.storage.IStorageUtilityIndexed;
 import org.javarosa.core.services.storage.StorageFullException;
 import org.javarosa.core.services.storage.StorageManager;
 import org.javarosa.core.util.externalizable.Externalizable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * PropertyManager is a class that is used to set and retrieve name/value pairs
@@ -41,6 +42,7 @@ import org.javarosa.core.util.externalizable.Externalizable;
  *
  */
 public class PropertyManager implements IPropertyManager {
+    private static final Logger logger = LoggerFactory.getLogger(PropertyManager.class);
 
     ///// manage global property manager /////
 
@@ -104,7 +106,7 @@ public class PropertyManager implements IPropertyManager {
         }
         if(retVal == null) {
             //#if debug.output==verbose
-            Std.out.println("Warning: Singular property request failed for property " + propertyName);
+            logger.warn("Singular property request failed for property {}", propertyName);
             //#endif
         }
         return retVal;
@@ -172,7 +174,7 @@ public class PropertyManager implements IPropertyManager {
             }
             //#if debug.output==verbose
             else {
-                Std.out.println("Property Manager: Unable to write value (" + propertyValue + ") to " + propertyName);
+                logger.info("Property Manager: Unable to write value ({}) to {}", propertyValue, propertyName);
             }
             //#endif
         }
--- a/src/org/javarosa/core/services/PropertyManager.java
+++ b/src/org/javarosa/core/services/PropertyManager.java
@@ -105,9 +105,7 @@ public class PropertyManager implements IPropertyManager {
             }
         }
         if(retVal == null) {
-            //#if debug.output==verbose
             logger.warn("Singular property request failed for property {}", propertyName);
-            //#endif
         }
         return retVal;
     }
@@ -171,12 +169,9 @@ public class PropertyManager implements IPropertyManager {
             if(valid) {
                 writeValue(propertyName, propertyValue);
                 notifyChanges(propertyName);
-            }
-            //#if debug.output==verbose
-            else {
+            } else {
                 logger.info("Property Manager: Unable to write value ({}) to {}", propertyValue, propertyName);
             }
-            //#endif
         }
 
     }
--- a/src/org/javarosa/core/services/PropertyManager.java
+++ b/src/org/javarosa/core/services/PropertyManager.java
@@ -57,7 +57,7 @@ public class PropertyManager implements IPropertyManager {
            setPropertyManager(new PropertyManager());
     }
 
-    public static IPropertyManager _ () {
+    public static IPropertyManager __() {
            if (instance == null) {
                initDefaultPropertyManager();
            }
--- a/src/org/javarosa/core/services/locale/LocalizationUtils.java
+++ b/src/org/javarosa/core/services/locale/LocalizationUtils.java
@@ -3,18 +3,21 @@
  */
 package org.javarosa.core.services.locale;
 
-import org.javarosa.core.io.Std;
 import org.javarosa.core.util.OrderedMap;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * @author ctsims
  *
  */
 public class LocalizationUtils {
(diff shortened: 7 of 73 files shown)


ggalmazor:
I think I broke git history... Maybe I need to recreate it...

src/org/javarosa/core/io/Std.java
```diff
@@ -1,32 +0,0 @@
-package org.javarosa.core.io;
-
-import java.io.PrintStream;
-
-/**
- * Used to break dependency directly on System.out and System.err.
- */
-public class Std {
```
dcbriccetti: Did you coordinate with @batkinson, who created this in July?
https://github.com/dcbriccetti/javarosa/commit/f97c27dcf0fbcb39775bde54cd2b4143d5954765

ggalmazor: Not really... It was very useful to guide the refactors, though...

dcbriccetti: This will be a breaking change for him. @lognaturel, should we ask for Brent’s review of this change?

.editorconfig
```diff
@@ -0,0 +1,5 @@
+root = true
```
dcbriccetti: I’ve been of the view that “meta” files like this, and IDE-specific files, don’t belong in the version control system. Probably because they are not part of the product being developed.

I don’t feel strongly about it, though.

src/org/javarosa/core/services/ProgramFlow.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public static void die(String thread, Exception e) {
+        //log exception
+        logger.error("unhandled exception at top level", e);
+
+        //crash
+        final FatalException crashException = new FatalException("unhandled exception in " + thread, e);
+
+        //depending on how the code was invoked, a straight 'throw' won't always reliably crash the app
+        //throwing in a thread should work (at least on our nokias)
```
dcbriccetti: “at least on our nokias” 
Huh? 
Where did this comment come from?

src/org/javarosa/core/services/ProgramFlow.java
```diff
@@ -0,0 +1,32 @@
+package org.javarosa.core.services;
+
+import org.javarosa.core.log.FatalException;
+import org.slf4j.LoggerFactory;
+
+public final class ProgramFlow {
+    private static final org.slf4j.Logger logger = LoggerFactory.getLogger(ProgramFlow.class);
+
+    public static void die(String thread, Exception e) {
```
dcbriccetti: Maybe add a doc comment, explaining what this is all about—why it’s needed.

src/org/javarosa/core/services/PropertyManager.java
```diff
@@ -104,7 +106,7 @@ public String getSingularProperty(String propertyName) {
         }
         if(retVal == null) {
             //#if debug.output==verbose
-            Std.out.println("Warning: Singular property request failed for property " + propertyName);
+            logger.warn("Singular property request failed for property {}", propertyName);
```
dcbriccetti: Does this,
`//#if debug.output`
do anything? If not, let’s get rid of them as part of this change.

.editorconfig
ggalmazor: In my opinion, this is really necessary while there's no better alternative. I'd go even further and add a code style definition for IntelliJ. Ideally, IntelliJ would read our checkstyle rules and configure code style settings to produce complying code but, in the meantime, I think these are very useful.

These things not only help a lot to ease people into start collaborating. They also prevent silly merging conflicts.

I can clearly remember having spent more than one hour monkeying around the code style settings to make the checkstyle rules on Briefcase pass. It was a really frustrating experience that would have been prevented had I have the code style definition already in the repo...

src/org/javarosa/core/services/locale/LocalizationUtils.java
```diff
@@ -84,7 +87,7 @@ private static void parseAndAdd(OrderedMap<String, String> locale, String line,
                 if(line.trim().equals("")) {
                     //Empty Line
                 } else {
-                    Std.out.println("Invalid line (#" + curline + ") read: " + line);
+                    logger.info("Invalid line (#{}) read: {}", curline, line);
```
dcbriccetti: I assume you weren’t able to test every one of these changes where you added the `{}` placeholders (a change I like), so I’ll just mention the risk of breakage.

src/org/javarosa/core/services/ProgramFlow.java
ggalmazor: This is actually old code that I've only moved around... I couldn't know. 

I mean, I can interpret that there's some intent of ending "gracefully" normal execution flow, but not much more...

.editorconfig
dcbriccetti: OK.

src/org/javarosa/core/services/ProgramFlow.java
ggalmazor: ¯\_(ツ)_/¯

src/org/javarosa/core/services/PropertyManager.java
ggalmazor: I was really puzzled by these comments. It felt like c#'s regions...

src/org/javarosa/core/services/locale/LocalizationUtils.java
ggalmazor: Thanks for noticing :)

I reviewed all the changes trice, but some problem might have slipped... I'll update the issue description to reflect this.

src/org/javarosa/xpath/expr/XPathExpression.java
```diff
@@ -76,9 +76,10 @@ public Object pivot (DataInstance model, EvaluationContext evalContext, List<Obj
     int indent;
 
     private void printStr (String s) {
+        StringBuilder padding = new StringBuilder();
         for (int i = 0; i < 2 * indent; i++)
-            Std.out.print(" ");
-        Std.out.println(s);
+            padding.append(" ");
```
dcbriccetti: Here we lack Python’s `' ' * 2 * indent`.

test/org/javarosa/core/model/data/test/TimeDataTests.java
```diff
@@ -54,8 +57,6 @@ public static Test suite() {
         return aSuite;
     }
     public static String testMaster (int testID) {
-        //System.out.println("running " + testID);
```
dcbriccetti: Ah, someone else likes deleting commented-out code. Very nice.

src/org/javarosa/xpath/expr/XPathExpression.java
ggalmazor: I'm sure there is a more elegant way to do left padding...

test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -79,11 +80,6 @@
     private static final String ORX_2_NAMESPACE_PREFIX = "orx2";
     private static final String ORX_2_NAMESPACE_URI = "http://openrosa.org/xforms";
 
-    @Before
```
dcbriccetti: Why remove? These code timings are appropriate here.

lognaturel:
Agreed that we should wait to hear from @batkinson to merge. I'm guessing he uses `Std` mostly to silence the noise but we should confirm and make sure he's happy with the changes.

src/org/javarosa/core/services/ProgramFlow.java
dcbriccetti: Sorry, I thought it was new.

dcbriccetti:
Do we need to light the Bat(kinson)-Signal? 
https://en.wikipedia.org/wiki/Bat-Signal

@batkinson

src/org/javarosa/core/io/Std.java
lognaturel: @batkinson wasn't on the TSC call today so I'm guessing he has not had Internet access. :/

How about leaving the `Std` class in, marking it as deprecated and just not using it internally anymore? We could do this with other public classes removed. If we're removing public interfaces we probably should do a major release. By leaving them in we are sure clients can upgrade without changes (even though they may not get the expected logging anymore).

dcbriccetti: Sounds good to me.

ggalmazor: OK, I'll review this option and get back with feedback... Let's hold this PR in the meantime

ggalmazor:
@lognaturel, @dcbriccetti, @batkinson, I've reverted changes on the removed classes and added deprecation annotations and JavaDoc notices wherever I felt there was a clear alternative (in `Logger` it's hard to tell, honestly). 

Please review the latest pair of commits https://github.com/opendatakit/javarosa/pull/259/commits/03d0b84e8e00fc3f0aed10ce8b046770d81425e5 and https://github.com/opendatakit/javarosa/pull/259/commits/0db70cdff7c2fd5d58f0d4010a67d265087fde99

src/org/javarosa/core/api/ILogger.java
```diff
@@ -15,38 +15,69 @@
...
  * IIncidentLogger's are used for instrumenting applications to identify usage
  * patterns, usability errors, and general trajectories through applications.
- * 
- * @author Clayton Sims
- * @date Apr 10, 2009 
  *
+ * @author Clayton Sims
+ * @deprecated Use {@link org.slf4j.LoggerFactory#getLogger(Class)} instead
  */
+@Deprecated
```
dcbriccetti: If you deprecate the type, must you also deprecate all the members?

src/org/javarosa/core/util/CodeTimer.java
```diff
@@ -16,40 +16,59 @@
...
 import java.io.PrintStream;
+import org.javarosa.core.io.Std;
 
-/** Helps with timing an operation and logging the time */
+/**
+ * Helps with timing an operation and logging the time
+ *
+ * @deprecated Use {@link StopWatch} instead
+ */
+@Deprecated
```
dcbriccetti: Hey, where were you when I wrote this class that turns out to be redundant? :-)

src/org/javarosa/core/api/ILogger.java
ggalmazor: It's just a pretext to add the `@deprecated` JavaDoc so that anyone ctrl-clicking gets information about an alternative...

src/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -4,8 +4,11 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class StreamsUtil {
+    private static final Logger logger = LoggerFactory.getLogger(StreamsUtil.class);
```
lognaturel: Is this used?

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -186,16 +191,15 @@ private static void initProcessingRules() {
                     final List<String> passedThroughInputAtts = Collections.unmodifiableList(Arrays.asList(
                             "rows",
                             "query"
-                    ));
-
-                    p.parseControl((IFormElement) parent, e, Constants.CONTROL_INPUT, passedThroughInputAtts,
-                            passedThroughInputAtts);
+                    ));p.parseControl((IFormElement) parent, e, Constants.CONTROL_INPUT,passedThroughInputAtts,
```
lognaturel: Looks like some funny things happened with spacing in the merge!

test/org/javarosa/core/util/test/NumericEncodingTest.java
```diff
@@ -19,16 +19,18 @@
...
+import org.slf4j.LoggerFactory;
 
-public class NumericEncodingTest extends TestCase  {
+public class NumericEncodingTest extends TestCase {
+    private static final Logger logger = LoggerFactory.getLogger(NumericEncodingTest.class);
 
     public NumericEncodingTest(String name) {
         super(name);
-        System.out.println("Running " + this.getClass().getName() + " test: " + name + "...");
+        logger.info("Running {} test: {}...", this.getClass().getName(), name);
```
lognaturel: I'm wondering whether System.out might be the right interface for a test... I'm not sure! Might be odd to get it in a log file?

src/org/javarosa/core/io/StreamsUtil.java
ggalmazor: I see two `StreamsUtil` classes on `org.javarosa.core.util` and `org.javarosa.core.io` and none are used by JR. It's alarming how they're basically the same classes with slightly different implementations for handling streams. 

I think this classes can be safely removed:
- JR itself doesn't use them.
- No one should depend on a JR's implementation of stream handling functions. There are better and far more battle-tested alternatives like commons-io and anyone relying on these classes should migrate their code a.s.a.p.

src/org/javarosa/xform/parse/XFormParser.java
ggalmazor: I've fixed that, but that class could benefit of static imports a lot. That block could be:
```java
final List<String> passedThroughInputAtts = unmodifiableList(asList("rows","query"));
p.parseControl((IFormElement) parent, e, CONTROL_INPUT, passedThroughInputAtts, passedThroughInputAtts);
```
(less cluttered)

Shall I make that change too?

Also, why is that list `final`? It's used there only to pass it to the `p.parseControl()` method, so, so what if someone reassigns it?

test/org/javarosa/core/util/test/NumericEncodingTest.java
ggalmazor: I don't know... I guess that what we should really do is to migrate tests to junit4 and use proper test method names for such feedback messages.

Do you want me to revert the changes on tests?

lognaturel: If you think sending them to the logging makes sense, I'm fine with leaving it. Agreed migrating the tests to junit4 would be ideal. Can that be a quick, surface change or will it require going deeper into each test file?

src/org/javarosa/core/io/StreamsUtil.java
lognaturel: 👍 How about marking both as deprecated?

src/org/javarosa/xform/parse/XFormParser.java
dcbriccetti: I like static imports where they improve readability and shorten code.

Scala has val and var. We use val unless the identifier’s value is allowed to change (which is generally discouraged). val is like Java’s final. The trouble with final in java is that it’s a keyword, and adds noise. So I generally skip it if it’s immediately obvious to the reader that the value doesn’t change, e.g., when its scope is just a few lines. So, I wouldn’t object to removing this final.

src/org/javarosa/core/io/StreamsUtil.java
ggalmazor: I'm still concerned... I think we should break build if we detect deprecation warnings.

test/org/javarosa/core/util/test/NumericEncodingTest.java
ggalmazor: I think the migration to junit4 is pretty straightforward thanks to the structure we have followed so far.

One would have to go through the all the 22 test classes and:
- Remove the extension on `TestCase`
- Add a `@Before` annotation on `setUp` and remove the call on `super.setUp()`
- Add a `@Test` annotation on all methods listed on `testMaster()`.
- Remove `testMaster()` and `suite()`
- Remove the class constructor

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -188,36 +191,30 @@ private static void initProcessingRules() {
...
                     // reasons.
-                    final List<String> passedThroughInputAtts = Collections.unmodifiableList(Arrays.asList(
-                        "rows",
-                        "query"
-                    ));
-                    p.parseControl((IFormElement) parent, e, Constants.CONTROL_INPUT,
-                        passedThroughInputAtts,
-                        passedThroughInputAtts
-                    );
+                    List<String> passedThroughInputAtts = unmodifiableList(asList("rows", "query"));
```
dcbriccetti: I also agree with putting these two string literals side by side. I wonder what my threshold is for wanting them in a column (or columns).

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -188,36 +191,30 @@ private static void initProcessingRules() {
...
+                    p.parseControl((IFormElement) parent, e, CONTROL_INPUT, passedThroughInputAtts, passedThroughInputAtts);
                 }
             });
             put("range", new IElementHandler() {
                 @Override public void handle(XFormParser p, Element e, Object parent) {
-                    p.parseControl((IFormElement) parent, e, Constants.CONTROL_RANGE,
-                        Arrays.asList("start", "end", "step") // Prevent warning about unexpected attributes
+                    p.parseControl((IFormElement) parent, e, CONTROL_RANGE,
```
dcbriccetti: I like having these `Constants` statically imported.

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -82,6 +82,9 @@
 import java.util.Map;
 import java.util.Set;
 
+import static java.util.Arrays.*;
+import static java.util.Collections.*;
+import static org.javarosa.core.model.Constants.*;
```
dcbriccetti: I like the wildcard import for Constants (because there are so many), but somewhat dislike it here for Arrays and Collections.

src/org/javarosa/xform/parse/XFormParser.java
ggalmazor: Set it to 999 :D

src/org/javarosa/xform/parse/XFormParser.java
ggalmazor: Yeah... static imports on Collections are kind of nice, though...

test/org/javarosa/core/util/test/NumericEncodingTest.java
lognaturel: Sounds great. @ggalmazor you want to just do it (in a PR after this one)?

src/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -5,22 +5,20 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 
+/**
+ * @deprecated Please, don't use anything from this class **at all**. It will removed on the next release
```
lognaturel: I think the deprecation annotation makes the intent clear but if you want to add a comment, perhaps "This class is unused and should remain that way. It will be removed in a future release.

src/org/javarosa/xform/parse/XFormParser.java
dcbriccetti: Right! I forgot who I was talking to. 😀   Mr. “I probably have one of these:”

dell-u3415w-ultrawide-curved-monitor-review-front-full-2-720x720

test/org/javarosa/core/util/test/NumericEncodingTest.java
ggalmazor: OK!

src/org/javarosa/core/services/ProgramFlow.java
batkinson: Hah, I do. Javarosa was originally developed for use on Nokia feature phones (I believe part of the original CommCare, the original, not CommCareODK, which came well after) and later the more powerful smart phones. You've found evidence of that fact here. This is likely a *very* old comment.

test/org/javarosa/core/util/test/NumericEncodingTest.java
batkinson: So, I have to disagree with @lognaturel's original comment here. Using the logging framework in the tests is the better option IMO. The reason is that it gives you the flexibility to redirect output to a file by default (so you don't blast the console during automated testing), but enable it conditionally while running them during a test debugging session. Really, this is what these frameworks were invented for. Remember those commented `System.out`s you removed? Well, if you log them you never have to leave commented print statements... and... you can enable them at runtime without ever having to recompile the code.

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -480,7 +489,7 @@ This will lead to inconsistency between prefixes used in the form definition (bi
         logger.info(codeTimer.logLine("Creating FormDef from parsed XML"));
     }
 
-    private final Set<String> validElementNames = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
+    private final Set<String> validElementNames = unmodifiableSet(new HashSet<>(asList(
```
batkinson: What's the difference between the final here and the final removed above? This jarred me a little after reading the commit message. In general, I would have split the changes so the commit message explained the *why* and not the action taken. I would have limited the scope of the change to the semantically relevant pieces only if possible.

src/org/javarosa/xform/parse/XFormParser.java
ggalmazor: I'm happier since I discovered that shift+mousewheel scrolls horizontally :)

src/org/javarosa/core/api/ILogger.java
```diff
@@ -15,38 +15,69 @@
...
 
 /**
  * IIncidentLogger's are used for instrumenting applications to identify usage
  * patterns, usability errors, and general trajectories through applications.
- * 
- * @author Clayton Sims
- * @date Apr 10, 2009 
  *
+ * @author Clayton Sims
+ * @deprecated Use {@link org.slf4j.LoggerFactory#getLogger(Class)} instead
```
batkinson: So, I just want to maybe spread a little code-culture here. I would try as much as possible to keep the little edits like you have here in separate commits, clearly explained by their corresponding commit message. It makes reviewing and understanding the scope of commits *way* easier because as soon as you see something that doesn't match the commit message you immediately can question it. Basically, this makes review fatigue immediate and code quality usually suffers as a result. You shouldn't be afraid to do the cleanup, just organize it in incremental commits. Just in case, it's not clear what I'm talking about...

You simultaneously did the following:

- Minor code formatting (whitespace)
- Reorganized imports
- Eliminated the  `@date` javadoc tag 
- Added the `@Deprecated` annotation
- Added the `@deprecated` javadoc tag

You can still make all these changes, but by reorganizing them you can eliminate the need for questions like Dave's here by having the commit message explain exactly *why* you're doing them. It also makes it easy to see your progression by your commit log. Basically, my commit log would look like the above bullets, except I'd reword the last 3 descriptions to explain the reasoning, not the action (which you can see easily using git diff, so the commit message adds little value):

- Eliminated the  `@date` javadoc tag -> Eliminated low-value javadoc noise? (guessing at reason)
- Added the `@Deprecated` annotation -> Annotated recently retired logging api as deprecated
- Added the `@deprecated` javadoc tag -> Provided deprecation suggestions using javadoc tags

You could argue that the last two could be combined, and I agree if kept clean and explainable by:

- Annotated recently deprecated api with javadoc suggestions

ggalmazor: This comment is pure gold. Thanks, @batkinson!

batkinson:
So, this looks ok to me, with the caveat that it was difficult to review due due to scope and organization of the changes. I tried to leave enough clues for the reason `Std.err|out` bit was introduced by using the same approach I recommended above. You can get it using the git pickaxe:

```
$ git log -S'Std.err'
commit f97c27dcf0fbcb39775bde54cd2b4143d5954765
Author: Brent Atkinson <brent.atkinson@gmail.com>
Date:   Fri Jul 14 00:13:49 2017 +0100

    Allow client redirection of output sent to stderr and stdout (#151)
```

batkinson:
In case it's not 100% clear, by converting the print statements to use the logging framework, the need for the code-based redirection becomes less important. Clients can provide their own logging configuration that redirects the output as needed.

src/org/javarosa/xform/parse/XFormParser.java
ggalmazor: Thanks for your suggestions!

Regarding `final` here's my rationale:
- Use `final` on all class fields. Remove it if you really need to reassign and there is no alternative
- Avoid `final` elsewhere

I see value expressing that some member is immutable by hinting it with the `final` keyword but my biggest argument is that it forces you to assign a value on construction, which plays really nice when you want to design your state with value objects. I prefer to work with value objects because it's easier to reason about code with them and they often force you to think about  their validity, emptiness, etc., but that's another discussion.

I wouldn't use `final` elsewhere because they tend to hide bigger problems:
- If you use `final` to protect yourself of a too big/complex method, the real solution is to refactor the method.
- If you use `final` to prevent extending classes to override a method, the design on that class hierarchy is probably flawed.

If your methods are small and your class hierarchies are well designed, adding `final` probably makes code harder to read without adding much value to it, in my opinion.

src/org/javarosa/core/io/StreamsUtil.java
ggalmazor: OK, I'll add the comments on deprecated classes.

This will be my last allegation in favor of breaking the build if there are compilation warnings :) 

Remember that we're already breaking the build for code styling reasons... breaking it for using deprecated stuff makes all the sense to me. Also, what risks do you see, @lognaturel? Maybe I'm missing something...

Right now, there's no warnings at all (I've solved them all). It would be really easy to roll this into the build :)

ggalmazor:
Thanks, @batkinson for reviewing the PR and for your very helpful insights about how to improve my git skills!

Normally, now I'd rework this PR's git history to start applying all the good stuff from @batkinson's review, but since this PR has already been reviewed by three people, and it's quite big, I'm afraid of introducing errors while doing so and we would have to review everything again.

Is it OK if we leave this PR like it is, or do you think that I should rework the git history?

src/org/javarosa/xform/parse/XFormParser.java
dcbriccetti: >Avoid final elsewhere

You make good points, @ggalmazor. You will probably agree that in a codebase such as this, with some long and confusing methods, with variables in scope for tens of lines, that it’s nice to know (via `final`) that at least a variable’s value doesn’t change.

ggalmazor: I think they help to understand current code, but if a variable is bound to some meaningful contiguous set of lines, it’s probably better to extract them to a small method.

If the variable is used in many meaningful sets of lines, then extract many small methods and pass the variable as an arg.

If the variable is used with no apparent structure... Then, yeah, `final` brings some good safety restrictions but I think that code would need rework.

ggalmazor:
@lognaturel, @yanokwa, do you think we can merge this PR?

lognaturel:
Yes, thanks, @ggalmazor!

PR #260, opened 2018-02-22
Remove unused, home-written, MD5 code.

dcbriccetti:
I wrote the following:
```
            MessageDigest md = MessageDigest.getInstance("MD5");
            md.update(message.getBytes());
            return md.digest();
```
and verified that its output matches that of the home-written code I deleted.

No


src/org/javarosa/core/util/externalizable/PrototypeFactory.java
```diff
@@ -129,7 +130,7 @@ public static Object getInstance (Class c) {
 
     public static byte[] getClassHash (Class type) {
         byte[] hash = new byte[CLASS_HASH_SIZE];
-        byte[] md5 = MD5.hash(type.getName().getBytes()); //add support for a salt, in case of collision?
+        byte[] md5 = getMd5Hash(type.getName());
```
ggalmazor: I think you could get away reusing what we did for the `digest()` function with just a little change. 

I have pushed it to my fork. You can cherry pick if you want: https://github.com/ggalmazor/javarosa/commit/0b571a1c04969a803f3f87c9f084e32ba10dba89

ggalmazor: Also, `DigestAlgorithm` is tested! yay! :)

dcbriccetti:
Thanks for making that so convenient, @ggalmazor.

PR #261, opened 2018-02-23
Issue 226 Keep time information for format-date() function

### LINKED ISSUE #226: Time component of now() calculation not available
[now-time.xml.txt](https://github.com/opendatakit/javarosa/files/1483901/now-time.xml.txt)

When an XLSForm is converted using pyxform, all calculates are of type `string`. @mdudzinski describes [here](https://github.com/opendatakit/javarosa/issues/199#issuecomment-345254947) why this leads to the time component not showing up. This is the `now3` case in the form above. Perhaps we should have a special case for a type of string that uses datetime to guarantee no info is lost.

<strike>`format-date(now(), "%Y-%m-%e %H:%M:%S.%3")` doesn't show time either. I don't think this is related and don't have a good explanation. This is the `now1` case in the form above.</strike> addressed in #261

ggalmazor:
Closes #226 

This PR changes `format-date()` function to keep time information.

Wrote a new test case for this function

I think this is a simpler solution than the one proposed on the issue.

I don't think so.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -224,7 +224,7 @@ public class XPathFuncExpr extends XPathExpression {
             return boolStr(argVals[0]);
         } else if (name.equals("format-date")) {
             assertArgsCount(name, args, 2);
-            return dateStr(argVals[0], argVals[1], true);
+            return formatDateTime(argVals[0], argVals[1]);
         } else if (name.equals("abs")) { //XPath 3.0
             checkArity(name, 1, args.length);
             return Math.abs(toDouble(argVals[0]));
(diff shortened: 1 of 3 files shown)


lognaturel:
The primary issue with #336 is the one with `string` calculates. I think @mdudzinski's suggestion for that one is the way to go and that'd be good to do in a separate PR.

I got confused about `format-date` because of the documentation at https://opendatakit.github.io/xforms-spec/#fn:format-date  -- `format-date-time` isn't documented at all (but will be shortly -- https://github.com/opendatakit/xforms-spec/issues/20). 

All that said, I'm in favor of this change. It makes `format-date` able to properly handle both `date` and `dateTime` questions without any downside I can see. How about also removing the `preserveTime` parameter while you're here?

ggalmazor:
Alright! While I was there, I've added tests covering (almost) all possible cases of `XPathFuncExpr.toDate()` because it felt somewhat complex and it's used by other parts of JR.

lognaturel:
🎉

PR #264, opened 2018-02-27
#227 Add support for xforms-value-changed event

### LINKED ISSUE #227: Add support for xforms-value-changed event

(

)

mdudzinski:
Closes #227 

Manual testing + unit tests.

This approach leverages existing `setvalue` action related code. 
As a different approach, the existing Observer Pattern code (`FormElementStateListener`) could be considered. However it seems to me to be too low level since it operates on `TreeElement` instances.

Existing code in `SetValueAction` uses `FormDef#setValue` to update the value of a node. Since now `FormDef#setValue` fires the `xforms-value-changed` event, it's possible to create a form that has dependency cycles. For example the following code:
```
<bind nodeset="/data/cost" type="decimal" />
<bind nodeset="/data/cost2" type="decimal" />
<bind nodeset="/data/cost3" type="decimal" />
...
<input ref="/data/cost">
  <setvalue event="xforms-value-changed" ref="/data/cost2" value="pow(/data/cost, 2)" />
</input>
<input ref="/data/cost2">
  <setvalue event="xforms-value-changed" ref="/data/cost3" value="pow(/data/cost2, 2)" />
</input>
<input ref="/data/cost3">
  <setvalue event="xforms-value-changed" ref="/data/cost" value="pow(/data/cost3, 2)" />
</input>
```
will keep looping till each node has the [infinity value](https://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#POSITIVE_INFINITY). In other scenarios it may loop till `StackOverflowException` is thrown.

Thoughts on fixing this so far:
* use `FormDef#setAnswer` instead of `setValue` - this method however won't fire recalculation of triggerables so it may not be a feasible solution
* simply don't fire the event (but fire triggerables) if a value is updated from `SetValueAction` - would require some ugly code to handle it and wouldn't be consistent
* validate against dependency cycle and don't allow such forms

### DIFF
--- a/resources/org/javarosa/core/model/actions/nested-setvalue-action-with-repeats.xml
+++ b/resources/org/javarosa/core/model/actions/nested-setvalue-action-with-repeats.xml
@@ -1,0 +1,32 @@
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa">
+    <h:head>
+        <h:title>Nested setvalue action</h:title>
+        <model>
+            <instance>
+                <data id="data-id">
+                    <repeat>
+                        <cost/>
+                        <cost_timestamp/>
+                    </repeat>
+                    <repeat>
+                        <cost/>
+                        <cost_timestamp/>
+                    </repeat>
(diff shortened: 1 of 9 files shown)


mdudzinski:
Whoops I messed something up and this PR got auto-closed. Created a new one #265.
Added `[INVALID]` prefix to this one to highlight it's broken or something.

PR #265, opened 2018-02-27
#227 Add support for xforms-value-changed event

### LINKED ISSUE #227: Add support for xforms-value-changed event

mdudzinski:
Closes #227 

Manual testing + unit tests.

This approach leverages existing `setvalue` action related code. 
As a different approach, the existing Observer Pattern code (`FormElementStateListener`) could be considered. However it seems to me to be too low level since it operates on `TreeElement` instances.

Existing code in `SetValueAction` uses `FormDef#setValue` to update the value of a node. Since now `FormDef#setValue` fires the `xforms-value-changed` event, it's possible to create a form that has dependency cycles. For example the following code:
```
<bind nodeset="/data/cost" type="decimal" />
<bind nodeset="/data/cost2" type="decimal" />
<bind nodeset="/data/cost3" type="decimal" />
...
<input ref="/data/cost">
  <setvalue event="xforms-value-changed" ref="/data/cost2" value="pow(/data/cost, 2)" />
</input>
<input ref="/data/cost2">
  <setvalue event="xforms-value-changed" ref="/data/cost3" value="pow(/data/cost2, 2)" />
</input>
<input ref="/data/cost3">
  <setvalue event="xforms-value-changed" ref="/data/cost" value="pow(/data/cost3, 2)" />
</input>
```
will keep looping till each node has the [infinity value](https://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#POSITIVE_INFINITY). In other scenarios it may loop till `StackOverflowException` is thrown.

Thoughts on fixing this so far:
* use `FormDef#setAnswer` instead of `setValue` - this method however won't fire recalculation of triggerables so it may not be a feasible solution
* simply don't fire the event (but fire triggerables) if a value is updated from `SetValueAction` - would require some ugly code to handle it and wouldn't be consistent
* validate against dependency cycle and don't allow such forms

### DIFF
--- a/src/org/javarosa/core/model/FormDef.java
+++ b/src/org/javarosa/core/model/FormDef.java
@@ -376,12 +376,16 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
                          boolean midSurvey) {
         IAnswerData oldValue = node.getValue();
         IAnswerDataSerializer answerDataSerializer = new XFormAnswerDataSerializer();
+        boolean valueChanged = !objectEquals(answerDataSerializer.serializeAnswerData(oldValue),
+                answerDataSerializer.serializeAnswerData(data));
         if (midSurvey && dagImpl.shouldTrustPreviouslyCommittedAnswer()
-                && objectEquals(answerDataSerializer.serializeAnswerData(oldValue),
-                answerDataSerializer.serializeAnswerData(data))) {
+                && !valueChanged) {
             return;
         }
         setAnswer(data, node);
+        if (valueChanged) {
+            dispatchFormEvent(Action.EVENT_XFORMS_VALUE_CHANGED, ref);
+        }
         Collection<QuickTriggerable> qts = triggerTriggerables(ref, midSurvey);
         dagImpl.publishSummary("New value", ref, qts);
         // TODO: pre-populate fix-count repeats here?
@@ -1633,10 +1637,14 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
         this.eventListeners.put(event, actions);
     }
 
-    public void dispatchFormEvent(String event) {
+    public void dispatchFormEvent(String event, TreeReference context) {
         for (Action action : getEventListeners(event)) {
-            action.processAction(this, null);
+            action.processAction(this, context);
         }
+    }
+
+    public void dispatchFormEvent(String event) {
+        dispatchFormEvent(event, null);
     }
 
     public <X extends XFormExtension> X getExtension(Class<X> extension) {
--- a/src/org/javarosa/core/model/FormDef.java
+++ b/src/org/javarosa/core/model/FormDef.java
@@ -19,6 +19,7 @@ package org.javarosa.core.model;
 import org.javarosa.core.io.Std;
 import org.javarosa.core.log.WrappedException;
 import org.javarosa.core.model.IDag.EventNotifierAccessor;
+import org.javarosa.core.model.actions.NestedSetValueAction;
 import org.javarosa.core.model.condition.Condition;
 import org.javarosa.core.model.condition.Constraint;
 import org.javarosa.core.model.condition.EvaluationContext;
@@ -655,6 +656,37 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
      */
     public void reportDependencyCycles(XFormParserReporter reporter) {
         dagImpl.reportDependencyCycles(reporter);
+    }
+
+
+    public void reportDependencyCycles() {
+        final EventNotifierAccessor ia = new EventNotifierAccessor() {
+
+            @Override
+            public EventNotifier getEventNotifier() {
+                return FormDef.this.getEventNotifier();
+            }
+        };
+        final IDag dag = new Safe2014DagImpl(ia);
+
+        for (final String event : eventListeners.keySet()) {
+            for (final Action action : getEventListeners(event)) {
+                if (!(action instanceof NestedSetValueAction)) {
+                    // Ignore;
+                    continue;
+                }
+
+                final NestedSetValueAction nestedAction = (NestedSetValueAction) action;
+
+                if (nestedAction.getValue() == null) {
+                    // No XPath expression, ignore
+                    continue;
+                }
+
+//                dag.addTriggerable(new Recalculate(nestedAction.getValue(), nestedAction.getTarget()));
+            }
+        }
+//        for (final Action action : getEventListeners())
     }
 
     /**
--- a/src/org/javarosa/core/model/actions/SetValueAction.java
+++ b/src/org/javarosa/core/model/actions/SetValueAction.java
@@ -28,6 +28,7 @@ import org.javarosa.xpath.expr.XPathFuncExpr;
  */
 public class SetValueAction extends Action {
     private TreeReference target;
+    private TreeReference trigger;
     private XPathExpression value;
     private String explicitValue;
 
@@ -35,19 +36,26 @@ public class SetValueAction extends Action {
 
     }
 
-    public SetValueAction(TreeReference target, XPathExpression value) {
+    public SetValueAction(TreeReference target,  TreeReference trigger, XPathExpression value) {
         super("setvalue");
         this.target = target;
         this.value = value;
+        this.trigger = trigger;
     }
 
-    public SetValueAction(TreeReference target, String explicitValue) {
+    public SetValueAction(TreeReference target, TreeReference trigger, String explicitValue) {
         super("setvalue");
         this.target = target;
         this.explicitValue = explicitValue;
+        this.trigger = trigger;
     }
 
     public void processAction(FormDef model, TreeReference contextRef) {
+
+        // If the trigger was specified then check if the context ref is not null and actually equals the trigger.
+        if (trigger != null && (contextRef == null || !trigger.equals(contextRef.genericize()))) {
+            return;
+        }
 
         //Qualify the reference if necessary
         TreeReference qualifiedReference = contextRef == null ? target : target.contextualize(contextRef);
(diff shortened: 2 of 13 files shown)


src/org/javarosa/core/model/FormDef.java
```diff
@@ -376,12 +376,16 @@ public void setValue(IAnswerData data, TreeReference ref, TreeElement node,
...
+        boolean valueChanged = !objectEquals(answerDataSerializer.serializeAnswerData(oldValue),
+                answerDataSerializer.serializeAnswerData(data));
         if (midSurvey && dagImpl.shouldTrustPreviouslyCommittedAnswer()
-                && objectEquals(answerDataSerializer.serializeAnswerData(oldValue),
-                answerDataSerializer.serializeAnswerData(data))) {
+                && !valueChanged) {
             return;
         }
         setAnswer(data, node);
+        if (valueChanged) {
```
mdudzinski: I leverages the existing code for checking whether a value has changed.

src/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -28,34 +28,45 @@
...
         TreeReference qualifiedReference = contextRef == null ? target : target.contextualize(contextRef);
 
         //For now we only process setValue actions which are within the
         //context if a context is provided. This happens for repeats where
         //insert events should only trigger on the right nodes
-        if(contextRef != null){
+        //If the trigger is specified then it means it's Action.EVENT_XFORMS_VALUE_CHANGED and thus
+        //it's not important if the target is withing the context. Context points to the node which value has changed and the target
+        //to be updated may be at any position within the form (instance).
+        if(contextRef != null && trigger == null) {
```
mdudzinski: I had to add some additional logic to make this code work with the nested set value elements. I'm not sure if I understand the existing comments but they seem to refer to `jr-insert` event. The problem doesn't look like it's related to our use cases but perhaps it make more sense to split this and create a separate class for nested `setvalue` actions?

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -332,7 +333,7 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
             return DateUtils.roundDate(new Date());
         } else if (name.equals("now")) {
             assertArgsCount(name, args, 0);
-            return new Date();
+            return new DateTime().toDate();
```
mdudzinski: For unit testing purposes (`DateTimeUtils`).

lognaturel:
This feature is SO USEFUL. Thank you, @mdudzinski. I can't wait to get this out to users.

I'm happy to see a simpler approach than what I had initially imagined. Seems good to me. 

Perhaps @dcbriccetti will have thoughts. For example, handling nested `setvalue`s in `SetValueAction` seems fine to me but he may have other suggestions. I would appreciate a quick header comment on the class leaving a little more info for the next person who comes by. Maybe just that `trigger` represents the outer node in the case of a nested `setvalue`. Or maybe a reference to the relevant tests?

When it comes to dependency cycles, I'd go for detecting and forbidding forms with them. This is what happens when there are cycles in e.g. calculates.

mdudzinski:
@lognaturel Sure, I'll add some comments and a validation for dependency cycles.

src/org/javarosa/core/model/FormDef.java
```diff
@@ -376,12 +376,16 @@ public void setValue(IAnswerData data, TreeReference ref, TreeElement node,
                          boolean midSurvey) {
         IAnswerData oldValue = node.getValue();
         IAnswerDataSerializer answerDataSerializer = new XFormAnswerDataSerializer();
+        boolean valueChanged = !objectEquals(answerDataSerializer.serializeAnswerData(oldValue),
+                answerDataSerializer.serializeAnswerData(data));
         if (midSurvey && dagImpl.shouldTrustPreviouslyCommittedAnswer()
-                && objectEquals(answerDataSerializer.serializeAnswerData(oldValue),
-                answerDataSerializer.serializeAnswerData(data))) {
+                && !valueChanged) {
```
dcbriccetti: Is this line break (before the &&) deliberate?

src/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -28,34 +28,45 @@
...
+        }
+
         //Qualify the reference if necessary
         TreeReference qualifiedReference = contextRef == null ? target : target.contextualize(contextRef);
 
         //For now we only process setValue actions which are within the
         //context if a context is provided. This happens for repeats where
         //insert events should only trigger on the right nodes
-        if(contextRef != null){
+        //If the trigger is specified then it means it's Action.EVENT_XFORMS_VALUE_CHANGED and thus
```
dcbriccetti: means its Action

src/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -28,34 +28,45 @@
...
+        //it's not important if the target is withing the context. Context points to the node which value has changed and the target
```
dcbriccetti: “node whose value” is better?

src/org/javarosa/core/model/actions/SetValueAction.java
dcbriccetti: Space after `if`
Do we have a space after `//` usually?

src/org/javarosa/core/model/actions/SetValueAction.java
lognaturel: In this case we do want `it is`. Maybe `it means it's an Action.EVENT_XFORMS_VALUE_CHANGED event`?

dcbriccetti: How embarrassing. I misread it as “its action value changed”.

src/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -96,7 +107,7 @@ public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOExcep
 
     public void writeExternal(DataOutputStream out) throws IOException {
         ExtUtil.write(out, target);
-
+        ExtUtil.write(out, trigger);
```
dcbriccetti: Do you need a corresponding `ExtUtil.read` in `readExternal`?

A broader question: What do we do with already serialized FormDefs when we change the format of the serialized data?

lognaturel: Good catch. Perhaps this is a reason to split out `SetNestedValueAction`?

With this structure, there is some chance that someone would build a form with the feature, load it in a version of Collect that doesn't support it, upgrade Collect and then be confused that the feature "doesn't work" because the form was serialized without it. In fact, I think that happened to me the first time I tried this branch. Am I understanding the implications correctly?

src/org/javarosa/core/model/actions/SetValueAction.java
mdudzinski: Sure, will fix it.

src/org/javarosa/core/model/actions/SetValueAction.java
mdudzinski: I'll fix the space.

I don't think we usually have a space after `//` but I just followed the style of the file. 
But I prefer the space after to be clear :)

lognaturel:
As I was looking for inspiration for the upcoming `setlocation` action, I came across the following: https://github.com/dimagi/commcare-j2me/pull/243. This is an implementation from a sister project, CommCare, also built on Javarosa. I wish I had seen this earlier! There's even a discussion of serialization.

There's a lot of precedent for incorporating commits from their work (ideally pulling the commits directly so attribution is clearly maintained). I think some parts like the introduction of `ActionController` we'll definitely want to incorporate for `setlocation`. I'm not sure what the best course of action is now. @mdudzinski when you have a chance do you want to take a bit to look at what they have and see whether you think it makes more sense to go with this and then add on other pieces inspired by their work or whether to start from the commits in that PR.

mdudzinski:
@lognaturel Thanks for discovering this! Looks interesting but it seems like they have a slightly different approach for the timestamping as they use an attribute on an instance node but they solution seems to support more general scenarios as well:
`
 <setvalue event="xforms-value-changed" ref="/data/age" value="if(/data/age = '', int((today() - date(/data/birthday)) div 365), /data/age)"/>
`
Their approach for the nested set value actions is quite different and requires far more changes. I tried to keep as minimum changes as possible but I don't see any bigger impediments to incorporate their work besides making their refactoring work compliant with what we have now in our codebase. I'll review it and let you know!

mdudzinski:
@ln
I tried to pull the commits directly but I keep facing problems. `git am` breaks at the first patch. It yields 
```
error: src/org/javarosa/core/model/actions/SetValueAction.java: patch does not apply
```
but no conflicts marks (<<<< etc.) are added and `git status` shows no local changes. I googled this and it seems to be a problem with line endings. Commcare javarosa seems to use Windows format. I tried dozen potential solutions - like [this one](https://unix.stackexchange.com/questions/239364/how-to-fix-hunk-1-failed-at-1-different-line-endings-message) - but with no luck. I also tried converting the patch file to unix format etc.

Have we ever pulled any commcare-j2me code? Maybe there is something that I missed?

Attaching patches (squashed into a single file):
[question-level-actions-support.patch.txt](https://github.com/opendatakit/javarosa/files/1913121/question-level-actions-support.patch.txt)

You need to pass `-p5 --directory='src/'` to the `git am` command in order to match our directory structure. We have `src/` where `commcare-j2me` has `core/src/main/java`

lognaturel:
I apologize, @mdudzinski, I totally missed this. 

There has been code pulled in from Dimagi's JavaRosa a number of times. https://github.com/opendatakit/javarosa/commit/9589972e165c13aa979d862575ec66dbbce1ad8a is an example. In that case it looks like the changeset was merged into one commit and applied. It's even possible it was done manually by doing something like copying and pasting the changed files.

If at all possible, it would be much better to keep the commit authors and messages since that history is valuable. I'll do my best to carve out some time for it next week. I'll also post in #javarosa-code to see if maybe someone else will have an idea. #

lognaturel:
@mdudzinski I now realize that in cea35910096031be8be1660b0a202537ab50ce3e you had already identified and brought in the CommCare code. Was there a reason you didn't want to bring all of it in at that point?

lognaturel:
Closing in favor of #323 which turns out to be very similar to this. The advantages are that introducing `ActionController` will set us up for pulling other related changes and that the history indicates which code was pulled from CommCare. Thanks for getting this started, @mdudzinski!

Issue #271, opened 2018-03-06
Possible incoherence/errors on date/time parsing code


ggalmazor:
While working on other things, I've noticed this method on `DateUtils`:

```java
public static Date parseDateTime (String str) {
    DateFields fields = new DateFields();
    int i = str.indexOf("T");
    if (i != -1) {
        if (!parseDate(str.substring(0, i), fields) || !parseTime(str.substring(i + 1), fields)) {
            return null;
        }
    } else {
        if (!parseDate(str, fields)) {
            return null;
        }
    }
    return getDate(fields);
}
```

Manually parsing date/times tends to be a pain and a common bug source in my experience and I thought that maybe we could get rid of maintaining that by simply using Joda time, which we're currently depending on for other stuff as well.

I changed the method to this:

```java
public static Date parseDateTime (String str) {
    try {
        return DateTime.parse(str).toDate();
    } catch(IllegalArgumentException e) {
        throw new XPathTypeMismatchException(e.getMessage());
    }
}
```

And then tests start to fail. If our parsing code was OK, replacing it with some other supposedly OK code would make no difference. IMHO, this is very fishy and it could be a sign that our current date/time parsing code has bugs. I'm not saying that Joda time is 100% reliable but I think it's reasonable to assume that they have expended much more time checking this kind of operations than us with JR.

PR #272, opened 2018-03-11
WIP - Issue 271 datetime parsing

ggalmazor:
Related to #271 

This PR is just intended to document the date time parsing code and to serve as a tool bench to try different things and learn about any possible problems that our current code might have.

## In this PR

- Replication of tests in `XPathEvalTest` into a new JUnit 4 `XPathEvalJunitTest` test class to spot what's failing more easily
- Migration of parsing code at `DateUtils` to Joda Time parsers.

## Strange test behavior

I've noticed some curious things:
- Before changing the implementation of `DateUtils` - commit https://github.com/opendatakit/javarosa/commit/836243c121e2744415f079d3f3a61536eb1b259f
  - Running all the tests makes the new `XPathEvalJunitTest` tests fail
  - Running only the new `XPathEvalJunitTest` all the tests pass
- After changing the implementation of `DateUtils`
  - Running all the tests makes many tests fail, including the old `XPathEvalTest` and the new `XPathEvalJunitTest`
  - Running only the old `XPathEvalTest` all the tests pass
  - Same with the new `XPathEvalJunitTest`
  - Same with `DateUtilsTests` which starts failing when run along with others, but it works if run alone
  - The tests on `DateUtilsTests` will fail whether you run them alone or along with other tests

## Insights

I think it's very probable that our original implementation is correct, although replacing it with a standard library seems a good move.

It feels like the different test failure scenarios are because while testing, we produce hidden lateral effects that make tests affect each other.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ out
 # mvn
 target
 secrets.xml
+*.iml
--- a/src/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/org/javarosa/core/model/utils/DateUtils.java
@@ -25,7 +25,9 @@ import java.util.ArrayList;
 import org.javarosa.core.services.locale.Localization;
 import org.javarosa.core.util.MathUtils;
 
+import org.javarosa.xpath.XPathTypeMismatchException;
 import org.joda.time.DateTime;
+import org.joda.time.LocalDate;
 import org.joda.time.LocalDateTime;
 import org.joda.time.format.DateTimeFormat;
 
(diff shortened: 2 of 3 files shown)


ggalmazor:
@opendatakit-bot label "needs discussion"

ggalmazor:
@opendatakit-bot unlabel "needs review"

ggalmazor:
I think that we should throw this one out.

The only thing we could salvage would be the migration to junit4 of the `XPathEvalTest`, what do you think, @lognaturel?

PR #274, opened 2018-03-13
Issue 226 calculates datetime values

### LINKED ISSUE #226: Time component of now() calculation not available

ggalmazor:
Closes #226

Created a new unit test that proves the fix.

You can check that the test fails on https://github.com/opendatakit/javarosa/commit/d9e7b8efc7e243218f0112c8cbbf26b9bae75ec3 before applying the fix.

The solution is a simplified version of what @mdudzinski suggested on https://github.com/opendatakit/javarosa/issues/226#issuecomment-346837214

None.

### DIFF
--- a/src/org/javarosa/core/model/condition/Recalculate.java
+++ b/src/org/javarosa/core/model/condition/Recalculate.java
@@ -17,7 +17,6 @@
 package org.javarosa.core.model.condition;
 
 import java.util.Date;
-
 import org.javarosa.core.model.Constants;
 import org.javarosa.core.model.data.BooleanData;
 import org.javarosa.core.model.data.DateData;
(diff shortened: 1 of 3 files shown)


src/org/javarosa/core/model/condition/Recalculate.java
```diff
@@ -142,14 +140,9 @@ public static IAnswerData wrapData (Object val, int dataType) {
...
-                return new DateTimeData((Date)val);
-            }
-            else if ( dataType == Constants.DATATYPE_TIME ) {
-                return new TimeData((Date)val);
-            } else {
-                return new DateData((Date)val);
-            }
+            if (dataType == Constants.DATATYPE_TIME)
+                return new TimeData((Date) val);
+            return new DateTimeData((Date) val);
```
lognaturel: Would it be better to have an explicit case for `DATATYPE_DATE`? When we know the intended time is date, I don't think we really want to add on a time of 0.

ggalmazor: Done!

PR #276, opened 2018-03-15
Pass through known bind attributes

### LINKED ISSUE #275: Required message ignored when jr:requiredMsg is set
#### Software versions 
JavaRosa v2.8.1 and 2.8.0

#### Problem description
`jr:requiredMsg` on bind is ignored.

It was originally reported here: https://github.com/opendatakit/collect/issues/2013

1. Load a form with a required message such as 
[required.xml.txt](https://github.com/opendatakit/javarosa/files/1811859/required.xml.txt)
2. Swipe forward on the first question

#### Expected behavior
A custom message should be shown.

#### Other information 
The problem is caused by https://github.com/opendatakit/javarosa/pull/238
Changes implemented in that pr caused that `rows` and `requiredMsg` attributes are completely ignored and they don't exist in additionalAttributes collection like before.

So not only `requiredMsg` is broken but `rows` too (we cant specify the number of rows for StringWidgets).

lognaturel:
Closes #275

Automated test.

This makes it explicit why these attributes have special status and follows the original code structure. No changes are made to public interfaces. @grzesiek2010 has an alternative at grzesiek2010/javarosa@59ae8c5. I think that would also be ok but it buries the passed through attributes.

For `rows`, I don't think it's common/important enough to warrant invasive changes.

I removed a test that would never fail because it was checking for errors rather than warnings. Warnings go to the output stream so they can't easily be captured. I think that's ok.

This could further affect binding code or not actually fix the problem with `requiredMsg`.

### DIFF
--- a/src/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
+++ b/src/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
@@ -29,7 +29,8 @@ class StandardBindAttributesProcessor {
     private Map<String, Integer> typeMappings;
 
     DataBinding createBinding(IXFormParserFunctions parserFunctions, FormDef formDef,
-                              Collection<String> usedAttributes, Element element) {
+                              Collection<String> usedAttributes, Collection<String> passedThroughAttributes,
+                              Element element) {
         final DataBinding binding = new DataBinding();
 
         binding.setId(element.getAttributeValue("", ID_ATTR));
@@ -116,7 +117,7 @@ class StandardBindAttributesProcessor {
         binding.setPreload(element.getAttributeValue(NAMESPACE_JAVAROSA, "preload"));
         binding.setPreloadParams(element.getAttributeValue(NAMESPACE_JAVAROSA, "preloadParams"));
 
-        saveUnusedAttributes(binding, element, usedAttributes);
+        saveUnusedAttributes(binding, element, usedAttributes, passedThroughAttributes);
 
         return binding;
     }
@@ -184,10 +185,11 @@ class StandardBindAttributesProcessor {
         return dataType;
     }
 
-    private void saveUnusedAttributes(DataBinding binding, Element element, Collection<String> usedAttributes) {
+    private void saveUnusedAttributes(DataBinding binding, Element element, Collection<String> usedAttributes,
+                                      Collection<String> passedThroughAttributes) {
         for (int i = 0; i < element.getAttributeCount(); i++) {
             String name = element.getAttributeName(i);
-            if (! usedAttributes.contains(name)) {
+            if (!usedAttributes.contains(name) || passedThroughAttributes.contains(name)) {
                 binding.setAdditionalAttribute(element.getAttributeNamespace(i), name, element.getAttributeValue(i));
             }
         }
--- a/src/org/javarosa/xform/parse/XFormParser.java
+++ b/src/org/javarosa/xform/parse/XFormParser.java
@@ -180,9 +180,16 @@ public class XFormParser implements IXFormParserFunctions {
         groupLevelHandlers = new HashMap<String, IElementHandler>() {{
             put("input", new IElementHandler() {
                 @Override public void handle(XFormParser p, Element e, Object parent) {
-                    p.parseControl((IFormElement) parent, e, Constants.CONTROL_INPUT,
-                            Arrays.asList("rows") // Prevent warning about unexpected attributes
-                    );
+                    // Attributes that are passed through to additionalAttributes but shouldn't lead to warnings.
+                    // These are consistently used by clients but are expected in additionalAttributes for historical
+                    // reasons.
+                    final List<String> passedThroughInputAtts = Collections.unmodifiableList(Arrays.asList(
+                            "rows",
+                            "query"
+                    ));
+
+                    p.parseControl((IFormElement) parent, e, Constants.CONTROL_INPUT, passedThroughInputAtts,
+                            passedThroughInputAtts);
                 }
             });
             put("range", new IElementHandler() {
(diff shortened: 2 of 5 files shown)


dcbriccetti:
I’ll probably finish reviewing later today, or, if you prefer to go ahead without me, please do.

lognaturel:
Would be great to get your full review when you have a chance, @dcbriccetti!

src/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
```diff
@@ -116,11 +118,16 @@ DataBinding createBinding(IXFormParserFunctions parserFunctions, FormDef formDef
         binding.setPreload(element.getAttributeValue(NAMESPACE_JAVAROSA, "preload"));
         binding.setPreloadParams(element.getAttributeValue(NAMESPACE_JAVAROSA, "preloadParams"));
 
-        saveUnusedAttributes(binding, element, usedAttributes);
+        saveUnusedAttributes(binding, element, usedAttributes, passedThroughAttributes);
 
         return binding;
     }
 
+    DataBinding createBinding(IXFormParserFunctions parserFunctions, FormDef formDef,
```
dcbriccetti: Is this method called? Everything still works when I delete it.

dcbriccetti:
Since we’re letting rows result in a warning, perhaps we can fix the duplicate text in that message.
```
XForm Parse Warning: Warning: 1 Unrecognized attributes found in Element [input] and will be ignored: [rows] Location:
    Problem found at nodeset: /html/body/group[@appearance=field-list]/input
    With element <input ref="/data/meta/instanceID" rows="2"><label>
    Problem found at nodeset: /html/body/group[@appearance=field-list]/input
    With element <input ref="/data/meta/instanceID" rows="2"><label>
```
I recommend we delete this line 163 from unusedAttWarning in XFormUtils:
`warning+="Location:\n"+XFormParser.getVagueLocation(e);`

All callers of unusedAttWarning already display the “getVagueLocation”.

[unusedAttWarning.txt](https://github.com/opendatakit/javarosa/files/1823062/unusedAttWarning.txt)

With this change, the error message now looks like this:
```
XForm Parse Warning: Warning: 1 Unrecognized attributes found in Element [input] and will be ignored: [rows] 
     Problem found at nodeset: /html/body/group[@appearance=field-list]/input
     With element <input ref="/data/meta/instanceID" rows="2"><label>
```

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1713,8 +1712,15 @@ protected DataBinding processStandardBindAttributes(List<String> usedAtts, Eleme
             "saveIncomplete"
     ));
 
+    // Attributes that are passed through to additionalAttrs but shouldn't lead to warnings.
```
dcbriccetti: I’m a big fan of having comments like these be doc comments, so that I can press F1 on the identifier where it is referenced, and see a helpful display like this:

screen shot 2018-03-18 at 1 02 04 pm

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -869,11 +878,27 @@ else if (itemChildName.equals("value")) {
         return tags;
     }
 
-    protected QuestionDef parseControl(IFormElement parent, Element e, int controlType) {
-        return parseControl (parent, e, controlType, null);
+    private QuestionDef parseControl(IFormElement parent, Element e, int controlType) {
```
dcbriccetti: Just making sure you’re aware that the return value from this method is never used. I would make the return type void, but I understand if you have a different preference.

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -869,11 +878,27 @@ else if (itemChildName.equals("value")) {
...
+        return parseControl (parent, e, controlType, null, null);
```
dcbriccetti: Did you mean to have the space before the parenthesis? Same a few lines down.

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -869,11 +878,27 @@ else if (itemChildName.equals("value")) {
...
+     *
+     * @param parent                the form control element's parent
+     * @param e                     the form control element to parse
+     * @param controlType           one of the control types defined in {@link org.javarosa.core.model.Constants}
+     * @param additionalUsedAtts    attributes specific to the control type
+     * @param passedThroughAtts     attributes specific to the control type that should be passed through to
+     *                              additionalAttributes for historical reasons
+     * @return                      a {@link org.javarosa.core.model.QuestionDef} representing the form control element
+     */
+    private QuestionDef parseControl(IFormElement parent, Element e, int controlType, List<String> additionalUsedAtts,
```
dcbriccetti: This doc comment is quite a nice improvement.

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -180,9 +180,16 @@ private static void initProcessingRules() {
...
+                    // Attributes that are passed through to additionalAttributes but shouldn't lead to warnings.
+                    // These are consistently used by clients but are expected in additionalAttributes for historical
+                    // reasons.
+                    final List<String> passedThroughInputAtts = Collections.unmodifiableList(Arrays.asList(
+                            "rows",
+                            "query"
+                    ));
+
+                    p.parseControl((IFormElement) parent, e, Constants.CONTROL_INPUT, passedThroughInputAtts,
+                            passedThroughInputAtts);
```
dcbriccetti: Using `passedThroughInputAtts` for both `additionalUsedAtts` and `passedThroughAtts` bothers me a little. Am I too picky? For your consideration:

screen shot 2018-03-19 at 5 55 12 pm

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1713,8 +1739,17 @@ protected DataBinding processStandardBindAttributes(List<String> usedAtts, Eleme
             "saveIncomplete"
     ));
 
-    protected void parseBind(Element element) {
-        final DataBinding binding = processStandardBindAttributes(usedAtts, element);
+    /**
+     * Attributes that are passed through to additionalAttrs but shouldn't lead to warnings.
+     * These are consistently used by clients but are expected in additionalAttrs for historical reasons.
+     **/
+    private final List<String> passedThroughAtts = Collections.unmodifiableList(Arrays.asList(
```
dcbriccetti: This documenting is great!

dcbriccetti:
Did you want to do this in a separate PR? (see above)

>delete this line 163 from unusedAttWarning in XFormUtils:
>warning+="Location:\n"+XFormParser.getVagueLocation(e);

Or just git-er-done here?

src/org/javarosa/xform/parse/XFormParser.java
lognaturel: Oh? I'm seeing it used in `parseUpload`.

src/org/javarosa/xform/parse/XFormParser.java
lognaturel: 🎉

src/org/javarosa/xform/parse/XFormParser.java
dcbriccetti: That’s fine.

src/org/javarosa/xform/parse/XFormParser.java
dcbriccetti: The second parseControl is called from parseUpload, no?

screen shot 2018-03-19 at 6 37 35 pm

lognaturel:
@dcbriccetti if you are satisfied, you can merge. We'll do a QA pass when updating JR in Collect. 🙏

PR #283, opened 2018-03-23
Test for attributes

grzesiek2010:
@lognaturel we were talking about adding a form with attributes to our checklist. I added an automated test I think it's a better solution but if you want us to test it manually as well please let me know.


lognaturel:
Sounds great to me! There's some overlap with the ones added in a4357cff and 930c631. How about removing those and adding the comments from there in your test to make it clear why known attributes are ending up in the catch all maps?

I would also recommend using `assertEquals` instead of `assertTrue` with a call on `equals`.

grzesiek2010:
> How about removing those and adding the comments from there in your test

Done.

> I would also recommend using assertEquals instead of assertTrue with a call on equals

Done.

> We should also add all the additional ones you identified to the pass through lists to avoid warnings.

I'll do that in a separate pr

> Finally, max-pixels was added recently and it's very unlikely another client expects it in bindAttributes. It might be worth fixing that one and making it a field on DataBinding instead of passing it through (would require a Collect change)

It should also be as separate pr.

resources/org/javarosa/form/api/form_with_additional_attributes.xml
```diff
@@ -0,0 +1,57 @@
...
+            <itext>
+                <translation lang="English">
+                    <text id="/data/group:label">
+                        <value>group</value>
+                    </text>
+                    <text id="/data/group/text:label">
+                        <value>text</value>
+                    </text>
+                    <text id="/data/group/geopoint:label">
+                        <value>geopoit</value>
```
lognaturel: point

test/org/javarosa/form/api/AttributesTestCase.java
```diff
@@ -0,0 +1,68 @@
+package org.javarosa.form.api;
```
lognaturel: How did you choose the package to put this in? It feels more like it would belong in `xform.parse` to me.

test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -387,45 +386,6 @@ public void parseFormWithSetValueAction() throws IOException {
         assertThat(groupElement.getBind(), is(expectedXPathReference));
     }
 
-    /**
-     * Attributes that started being used by clients without being added as fields to DataBinding or QuestionDef should
-     * be passed through and made available in the bindAttributes or additionalAttributes list.
-     */
-    @Test public void passedThroughAttributesHaveExpectedValues() throws IOException {
-        ParseResult parseResult = parse(r("whitelisted-attributes.xml"));
```
lognaturel: The XML file should be unused and removable.

grzesiek2010: Yas you were right, I removed that file.

grzesiek2010:
@lognaturel it should be ready.

lognaturel:
@ggalmazor would be great to get your quick review partly as an FYI since you were just in this code but also in case you have suggestions!

PR #288, opened 2018-04-03
Improved the comment regarding #279

grzesiek2010:
@lognaturel 

I just would like to add a better comment here since after fixing #279 I discovered one thing. That problem takes places only when ref attributes are not specified. It's the same problem like here: https://github.com/opendatakit/javarosa/issues/218

and it's a problem wit ODKBuild because, if we build a form using xls everything is ok.

[nestedGroupsBuildUsingXLS.xml.txt](https://github.com/opendatakit/javarosa/files/1871749/nestedGroupsBuildUsingXLS.xml.txt)
[nestedGroupsBuildUsingODKBuild.xml.txt](https://github.com/opendatakit/javarosa/files/1871750/nestedGroupsBuildUsingODKBuild.xml.txt)

So just a short explanation of what the root cause is.


lognaturel:
Nice. 👍 There's an issue for Build at https://github.com/opendatakit/build/issues/195 so eventually that case should be less common.

PR #289, opened 2018-04-04
Issue 286 Java 8

### LINKED ISSUE #286: Migration to Java8
While working on #272, I discovered some weird inconsistencies on how we deal with dates, times and timezones in JR. My rationale for pursuing this was "since we're using Joda Time for date/time stuff, we shouldn't implement anything date/time related and delegate on Joda, which is far more battle-tested than JR".

But then, why don't we migrate JR to Java8 and use the official, native, immutable and generally-nicer new [Date and Time API](http://www.oracle.com/technetwork/articles/java/jf14-date-time-2125367.html)?

By migrating to Java8 we wouldn't only benefit from this new API. Java8 comes with other new APIs and improvements aimed to work better with efficiency focused code (Streams, new concurrency artifacts, etc.) (and Lambdas!). I'm a newcomer to JR but it feels like doing this would open a lot of room for improvements.

The other projects in ODK that depend on JR are already (or will be in the near future) using Java8, so, *a priori*,  it seems like a safe move. 

I don't know what's the specific situation with Collect, though.

Let's hear your opinions!

ggalmazor:
Closes #286 

Bump Gradle and Maven build Java source and target version to 1.8

Build and run tests with Gradle and Maven

N/A

None, apparently

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,8 +24,8 @@ sourceSets {
     }
 }
 
-targetCompatibility = '1.7'
-sourceCompatibility = '1.7'
+targetCompatibility = '1.8'
+sourceCompatibility = '1.8'
 
 dependencies {
     // Be sure to update dependencies in pom.xml to match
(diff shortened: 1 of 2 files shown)


lognaturel:
YES. Nice to see that was no sweat.

PR #295, opened 2018-04-10
Add rank form control element

### LINKED ISSUE #294: Add rank form control element in odk namespace
The `rank` form control element can be used the same way as `select` but the order of the resulting space-separated results list is meaningful.

Because `rank` is not defined in any other known XForms spec, it should be introduced in the `http://opendatakit.org/xforms` namespace: https://opendatakit.github.io/xforms-spec/#namespaces and https://github.com/opendatakit/xforms-spec/issues/73#issuecomment-373860590.

Like `select`, rank should be usable with either `item` children or an `itemset` child.

```xml
<!-- primary instance -->
<data>
     <ranking>b a</ranking>
</data>
<!-- body -->
<odk:rank ref="/data/ranking">
     <label>label</label>
     <hint/>
     <item>
           <label>A</label>
           <value>a</value>
     </item>
     <item>
           <label>B</label>
           <value>b</value>
     </item>
</odk:rank>
```

```xml
<!-- primary instance -->
<data>
     <ranking>b a</ranking>
</data>
<!-- body -->
<odk:rank ref="/data/ranking">
     <label>label</label>
     <hint/>
     <itemset nodeset="instance('something')/root/item">
           <label ref="./label" />
           <value ref="./value" />
     </itemset>
</odk:rank>
```

grzesiek2010:
Closes #294 

I tested this changes with Collect adding the project as a module.

I decide to use the same data class for both `SelectMutli` and `Rank` since the code is exactly the same. @lognaturel in the issue you also asked about reading values from external csv but since it's implemented only for `SelectOne` not even for `SelectMulti` I think it should be addressed in a separate pull request.

I renamed `SelectMultiData` class to share the code and also `DATATYPE_CHOICE_LIST`. I hope I handled everything well but there may be the risk.


dcbriccetti:
@lognaturel suggests “get a first draft of the actual widget up in Collect to take a look at before there’s a deep review on the JR side”. I’ll hold off on reviewing.

src/org/javarosa/core/model/CoreModelModule.java
```diff
@@ -41,7 +41,7 @@
             "org.javarosa.core.model.data.LongData",
             "org.javarosa.core.model.data.MultiPointerAnswerData",
             "org.javarosa.core.model.data.PointerAnswerData",
-            "org.javarosa.core.model.data.SelectMultiData",
+            "org.javarosa.core.model.data.MultipleItemsData",
```
dcbriccetti: Do you need to also keep SelectMultiData here? Do instances of that class get serialized?

lognaturel: Hmm, good question, @dcbriccetti. They do. `SelectMultiData` is now a subclass of `MultipleItemsData` but if `SelectMultiData` is serialized, it does need to be here despite that, right?

src/org/javarosa/core/model/Constants.java
```diff
@@ -57,7 +57,8 @@
     public static final int DATATYPE_TIME = 5;    /** Time question type. This has only time element without date*/
     public static final int DATATYPE_DATE_TIME = 6;    /** Date and Time question type. This has both the date and time components*/
     public static final int DATATYPE_CHOICE = 7;    /** This is a question with alist of options where not more than one option can be selected at a time. */
-    public static final int DATATYPE_CHOICE_LIST = 8;    /** This is a question with alist of options where more than one option can be selected at a time. */
+    public static final int DATATYPE_MULTIPLE_ITEMS = 8;    /** This is a question with a list of items used form selecting multiple answers or ordering them. */
```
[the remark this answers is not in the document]
grzesiek2010: Fixed.

src/org/javarosa/core/model/CoreModelModule.java
grzesiek2010: Fixed.

lognaturel:
Ok, there's been plenty of time for discussion on the forum and Collect PR so I think this is ready to go in. @dcbriccetti any other comments? Thanks for spotting the missing serialization.

src/org/javarosa/core/model/DataType.java
```diff
@@ -38,23 +38,23 @@
...
+    UNSUPPORTED     (DATATYPE_UNSUPPORTED),
+    NULL            (DATATYPE_NULL),
+    TEXT            (DATATYPE_TEXT),
+    INTEGER         (DATATYPE_INTEGER),
+    DECIMAL         (DATATYPE_DECIMAL),
+    DATE            (DATATYPE_DATE),
+    TIME            (DATATYPE_TIME),
+    DATE_TIME       (DATATYPE_DATE_TIME),
+    CHOICE          (DATATYPE_CHOICE),
+    MULTIPLE_ITEMS  (DATATYPE_MULTIPLE_ITEMS),
```
dcbriccetti: Good. Thanks for knowing that you needed to keep this enum in sync with Constants. And extra special thanks for preserving the beautifully aligned columns that I so lovingly created.

lognaturel: 😄

lognaturel:
Following the discussion at https://github.com/opendatakit/xforms-spec/issues/203, I see that namespace was not actually enforced here. Additionally, the test form puts `rank` in the wrong namespace.

PR #296, opened 2018-04-10
Issue 273 randomize nodeset

### LINKED ISSUE #273: Add randomize(nodeset) and randomize(nodeset, seed) functions
See https://github.com/opendatakit/xforms-spec/issues/51#issuecomment-371910870

* Use Fisher-Yates for shuffling
* If a seed is provided, that seed is passed when building a `Random` object. `nextInt` is used to get random values
* Tests and expected values should be shared on the XForms spec issue so other implementations can verify their outcomes are the same

ggalmazor:
Closes #273 

Wrote comprehensive unit tests, covering both positive cases and border cases:
- One for the low-level `RandomizeHelper`, which ultimately holds all the implementation to support everything around randomize()
- One for `XPathNodeset.shuffle()`, which ensures that everything works at the `XPathNodeset` level
- One outside-in test with a real form file that uses randomize for shuffling the choices on a select field.

This PR implements a generic randomize() implementation and an override to specifically using it with itemset nodeset definitions.

We need to consider a fully-generic support for randomize() which involves changing the way nodeset definitions work. We're already discussing this on this PR's comments and we're exploring options.

None.

-------
(Previous conversation about this PR)

I'd like to get this out to receive some feedback from more experienced contributors. Am I going in the right direction with this?

So far:
- I've written a new `XPathNodeset.shuffle(nodeset, seed?)` method and tested based on @lognaturel's https://github.com/lognaturel/javarosa/commit/daf9dfa6211eff6dba26e9c72d91dda4289b3e2f
- I've written a new xform `randomize(nodeset, seed)` function and a test with a form that uses it.
- I haven't changed yet anything related to the `ItemsetBinding` class

Especially, I'd like to validate that the form file I'm using on my `XPathFuncExprRandomizeTest` is correct and represents the new behavior we're trying to implement.

I'd also like to comment on the strategy I've been trying to follow (without success yet). Honestly, I am kind of reverse engineering the code and I still don't understand much of JR's code.

The error I get with my test is:
```
org.javarosa.xpath.XPathTypeMismatchException: XPath evaluation: type mismatch 
Expected XPath path, got XPath expression: [randomize(instance('counties')/root/item[state=/data/state])],null

	at org.javarosa.model.xform.XPathReference.getPathExpr(XPathReference.java:72)
	at org.javarosa.xform.parse.XFormParser.parseItemset(XFormParser.java:1276)
	at org.javarosa.xform.parse.XFormParser.parseControl(XFormParser.java:968)
	at org.javarosa.xform.parse.XFormParser.parseControl(XFormParser.java:882)
	at org.javarosa.xform.parse.XFormParser.access$200(XFormParser.java:118)
	at org.javarosa.xform.parse.XFormParser$1$5.handle(XFormParser.java:229)
	at org.javarosa.xform.parse.XFormParser.parseElement(XFormParser.java:513)
	at org.javarosa.xform.parse.XFormParser.parseElement(XFormParser.java:522)
	at org.javarosa.xform.parse.XFormParser.parseElement(XFormParser.java:522)
	at org.javarosa.xform.parse.XFormParser.parseDoc(XFormParser.java:429)
	at org.javarosa.xform.parse.XFormParser.parse(XFormParser.java:343)
	at org.javarosa.xform.util.XFormUtils.getFormFromInputStream(XFormUtils.java:81)
	at org.javarosa.core.test.FormParseInit.init(FormParseInit.java:74)
	at org.javarosa.core.test.FormParseInit.setFormToParse(FormParseInit.java:60)
	at org.javarosa.xpath.expr.XPathFuncExprRandomizeTest.name(XPathFuncExprRandomizeTest.java:19)
```
This is just like we would expect because in `XFormParser.parseItemset(XFormParser.java:1276)` we expect an `XPathPathExpr` and there's an `XPathFuncExpr` `randomize(instance('counties')/root/item[state=/data/state])` instead.

To work around this, we could change `getPathExpr(String nodeset)` to return an `XPathExpression` (`XPathPathExpr`'s and `XPathFuncExpr`'s super) like so:
```java
public static XPathExpression getPathExpr(String nodeset) {
    try {
        XPathExpression path = XPathParseTool.parseXPath(nodeset);
        if ((path instanceof XPathPathExpr) || (path instanceof XPathFuncExpr))
            return path;
        else
            throw new XPathTypeMismatchException("Expected an XPath path or an XPath function, got XPath expression: [" + nodeset + "]");
    } catch (XPathSyntaxException xse) {
        logger.error("Error", xse);
        throw new XPathException("Parse error in XPath path: [" + nodeset + "]." + (xse.getMessage() == null ? "" : "\n" + xse.getMessage()));
    }
}
```
(then, `XPathReference.ref` can be null)

At this point, I can make all the tests pass expect the new one focused on testing `randomize()`, which now errors:
```
java.lang.ClassCastException: org.javarosa.xpath.expr.XPathFuncExpr cannot be cast to org.javarosa.xpath.expr.XPathPathExpr

	at org.javarosa.core.model.ItemsetBinding.initReferences(ItemsetBinding.java:107)
	at org.javarosa.core.model.FormDef.updateItemsetReferences(FormDef.java:1469)
	at org.javarosa.xform.parse.FormInstanceParser.parseInstance(FormInstanceParser.java:81)
	at org.javarosa.xform.parse.XFormParser.parseDoc(XFormParser.java:463)
	at org.javarosa.xform.parse.XFormParser.parse(XFormParser.java:344)
	at org.javarosa.xform.util.XFormUtils.getFormFromInputStream(XFormUtils.java:81)
	at org.javarosa.core.test.FormParseInit.init(FormParseInit.java:74)
	at org.javarosa.core.test.FormParseInit.setFormToParse(FormParseInit.java:60)
	at org.javarosa.xpath.expr.XPathFuncExprRandomizeTest.name(XPathFuncExprRandomizeTest.java:19)
```
(finally got to `ItemsetBinding`)

If I understood correctly, we want an itemset to support values that can be actual nodesets or functions that return nodesets. This implies that the `ItemsetBinding.nodesetRef` won't always exist.

Adjusting this class to the fact that `((XPathConditional) this.nodesetExpr).getExpr()` will return an `XPathExpression` (which can be a path or a function) I can conditionally assign a null in `nodesetRef` to get going.

That gets my test to a new error:
```
java.lang.NullPointerException
	at org.javarosa.xform.parse.FormInstanceParser.getRepeatableRefs(FormInstanceParser.java:579)
	at org.javarosa.xform.parse.FormInstanceParser.flagRepeatables(FormInstanceParser.java:115)
	at org.javarosa.xform.parse.FormInstanceParser.processRepeats(FormInstanceParser.java:107)
	at org.javarosa.xform.parse.FormInstanceParser.parseInstance(FormInstanceParser.java:82)
	at org.javarosa.xform.parse.XFormParser.parseDoc(XFormParser.java:463)
	at org.javarosa.xform.parse.XFormParser.parse(XFormParser.java:344)
	at org.javarosa.xform.util.XFormUtils.getFormFromInputStream(XFormUtils.java:81)
	at org.javarosa.core.test.FormParseInit.init(FormParseInit.java:74)
	at org.javarosa.core.test.FormParseInit.setFormToParse(FormParseInit.java:60)
	at org.javarosa.xpath.expr.XPathFuncExprRandomizeTest.name(XPathFuncExprRandomizeTest.java:19)
```

That's normal because now the `nodesetRef` is null.

Here's the deal:
I wanted to explore the idea of lazy evaluating the refs of an itemset using its `nodesetExpr` but that requires to get the `EvaluationContext`. Does this make sense?

I've pushed a WIP branch with the changes I mention here: https://github.com/ggalmazor/javarosa/commit/c11b7ca8c42e016c5f997b051e189f6815f9b1d9

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -424,6 +424,17 @@ public class XPathFuncExpr extends XPathExpression {
                 (String) argVals[0],
                 args.length == 3 ? Encoding.from((String)argVals[2]) : Encoding.BASE64
             );
+        } else if (name.equals("randomize")) {
+            if (!(argVals[0] instanceof XPathNodeset))
+                throw new XPathTypeMismatchException("First argument to randomize must be a nodeset");
+
+            if (args.length == 1)
+                return XPathNodeset.shuffle((XPathNodeset) argVals[0]);
+
+            if (args.length == 2)
+                return XPathNodeset.shuffle((XPathNodeset) argVals[0], toNumeric(argVals[1]).longValue());
+
+            throw new XPathUnhandledException("function \'randomize\' requires 1 or 2 arguments. " + args.length + " provided.");
         } else {
             //check for custom handler
             IFunctionHandler handler = funcHandlers.get(name);
--- a/test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
+++ b/test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2018 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.javarosa.xpath.expr;
+
+import static org.javarosa.core.model.instance.TreeReference.CONTEXT_ABSOLUTE;
+import static org.javarosa.core.model.instance.TreeReference.INDEX_UNBOUND;
+import static org.javarosa.core.model.instance.TreeReference.REF_ABSOLUTE;
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Arrays;
+import java.util.List;
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.FormIndex;
+import org.javarosa.core.model.SelectChoice;
+import org.javarosa.core.model.instance.InstanceInitializationFactory;
+import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.test.FormParseInit;
+import org.javarosa.form.api.FormEntryPrompt;
+import org.junit.Before;
+import org.junit.Test;
+
+public class XPathFuncExprRandomizeTest {
+
+    private FormDef formDef;
+
+    @Before
+    public void setUp() {
+        // Take a look to the form file:
+        //  - The first pair of fields use randomize without a seed.
+        //  - The second pair of fields use randomize with a seed.
+        FormParseInit fpi = new FormParseInit();
+        fpi.setFormToParse(r("randomize.xml").toString());
+        formDef = fpi.getFormDef();
+    }
+
+    @Test
+    public void fields_without_seed_in_the_same_form_get_a_different_order_of_choices() {
+        initializeNewInstance(formDef);
+        List<SelectChoice> choices1 = getSelectChoices(formDef, "/randomize/fruit1");
+        List<SelectChoice> choices2 = getSelectChoices(formDef, "/randomize/fruit2");
+
+        assertFalse(nodesEqualInOrder(choices1, choices2));
+    }
+
+    @Test
+    public void the_same_field_without_seed_in_different_instances_gets_a_different_order_of_choices() {
+        initializeNewInstance(formDef);
+        List<SelectChoice> choices1 = getSelectChoices(formDef, "/randomize/fruit1");
+
+        initializeNewInstance(formDef);
+        List<SelectChoice> choices2 = getSelectChoices(formDef, "/randomize/fruit1");
+
+        assertFalse(nodesEqualInOrder(choices1, choices2));
+    }
+
+    @Test
+    public void seeded_fields_in_the_same_form_get_the_same_order_of_choices() {
+        initializeNewInstance(formDef);
+        List<SelectChoice> choices1 = getSelectChoices(formDef, "/randomize/seededFruit1");
+        List<SelectChoice> choices2 = getSelectChoices(formDef, "/randomize/seededFruit2");
+
+        assertTrue(nodesEqualInOrder(choices1, choices2));
+    }
+
+    @Test
+    public void the_same_seeded_field_in_different_instances_gets_the_same_order_of_choices() {
+        initializeNewInstance(formDef);
+        List<SelectChoice> choices1 = getSelectChoices(formDef, "/randomize/seededFruit2");
+
+        initializeNewInstance(formDef);
+        List<SelectChoice> choices2 = getSelectChoices(formDef, "/randomize/seededFruit2");
+
+        assertTrue(nodesEqualInOrder(choices1, choices2));
+    }
+
+    private static void initializeNewInstance(FormDef formDef) {
+        formDef.initialize(true, new InstanceInitializationFactory());
+    }
+
+    private List<SelectChoice> getSelectChoices(FormDef formDef, String ref) {
+        FormIndex formIndex = getFormIndex(formDef, absoluteRef(ref));
+        FormEntryPrompt formEntryPrompt = new FormEntryPrompt(formDef, formIndex);
+        return formEntryPrompt.getSelectChoices();
+    }
+
+    private FormIndex getFormIndex(FormDef formDef, TreeReference ref) {
+        for (int localIndex = 0, lastIndex = formDef.getChildren().size(); localIndex < lastIndex; localIndex++)
+            if (formDef.getChild(localIndex).getBind().getReference().equals(ref))
+                return new FormIndex(localIndex, 0, ref);
+        throw new IllegalArgumentException("Reference " + ref + " not found");
+    }
+
+    private static boolean nodesEqualInOrder(List<SelectChoice> left, List<SelectChoice> right) {
+        if (left.size() != right.size())
+            return false;
+
+        for (int i = 0; i < left.size(); i++)
+            if (!left.get(i).getValue().equals(right.get(i).getValue()))
+                return false;
+
+        return true;
+    }
+
+    private static TreeReference absoluteRef(String path) {
+        TreeReference tr = new TreeReference();
+        tr.setRefLevel(REF_ABSOLUTE);
+        tr.setContext(CONTEXT_ABSOLUTE);
+        tr.setInstanceName(null);
+        Arrays.stream(path.split("/"))
+            .filter(s -> !s.isEmpty())
+            .forEach(s -> tr.add(s, INDEX_UNBOUND));
+        return tr;
+    }
+}
--- a/test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
+++ b/test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
@@ -15,6 +15,10 @@
  */
 package org.javarosa.xpath.expr;
 
+import static java.nio.file.Files.createTempFile;
+import static java.nio.file.Files.delete;
+import static java.nio.file.Files.newInputStream;
+import static java.nio.file.Files.newOutputStream;
 import static org.javarosa.core.model.instance.TreeReference.CONTEXT_ABSOLUTE;
 import static org.javarosa.core.model.instance.TreeReference.INDEX_UNBOUND;
 import static org.javarosa.core.model.instance.TreeReference.REF_ABSOLUTE;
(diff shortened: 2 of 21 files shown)


lognaturel:
🤦‍♀️ Just realized there's a bug in the shuffle implementation: the last value can be removed and then the random index can go off the end of the list. The first two lines should be:
```java
SelectChoice current = choices.get(i);
choices.set(i, choices.get(randomIndex));
```

The test form you're using looks good but it has a lot going on (translations, cascades/choice filters/predicates). Here's a simpler one you may like - [randomize.xml.txt](https://github.com/opendatakit/javarosa/files/1905001/randomize.xml.txt). That said, I think lots of the tricky bits have to do with evaluating predicates so maybe it's good to directly go to that more complex one.

> Honestly, I am kind of reverse engineering the code

We all are for better or worse. @dcbriccetti had started writing down some of the results of his reverse engineering to be turned into documentation. @dcbriccetti any notes on `ItemsetBinding` and related parts of the code?

> If I understood correctly, we want an itemset to support values that can be actual nodesets or functions that return nodesets.

That's the absolute ideal. But we could also consider narrowly supporting `randomize` for now.

One strategy that has been percolating through my brain and which is definitely a hack is to add a special case for randomize in `XFormParser.parseItemset`. Honestly I'm a little embarrassed to even suggest it but "hey, it works." It's also similar to what's done to set `labelIsItext` just a little below. 

https://github.com/lognaturel/javarosa/commit/ee526ed7e3c06105598fc5c2a2d7c7fe5e69e965 has a quick demo of that approach. Something isn't right with seeds but I didn't want to dig too deep until we decide that's the way to go (for now?). One thing we could consider is to get that working and get it out since it's such useful functionality and consider "doing it right" as there's a deeper understanding of JR built up. And if it never gets reworked, at least it's a pretty contained hack. @ggalmazor, if you're on board, feel free to run with it and make it Actually Work. 

All that said, your approach is looking promising. How are you planning to eventually resolve the label and value references? Is that where you're thinking you'll need a context? It seems like what we want to be able to do is use each node as the base for the label and value expressions instead of using the nodeset path. That would then allow us not only to use functions that return nodesets but also to do things like use a compound nodeset built with `|` (join).

There's this comment with `contextRef`:
> //note: this is only here because its currently impossible to both (a) get a form control's parent, and (b) convert expressions into refs while preserving predicates. once these are fixed, this field can go away

It seems it's probably really insightful but it will take some unpacking! It might be worth spending some time figuring out where those limitations are coming from. I'm not totally clear on why the predicates need to be maintained.

I'm also intrigued by this `FormDef` comment about `jr:choice-name`:
> // NOTE: this is highly suspect. We have no context against
// which to evaluate
// a dynamic selection list. This will generally cause that
// evaluation to break
// if any filtering is done, or, worst case, give unexpected
// results.

ggalmazor:
Thanks, @lognaturel! All this is really good stuff and just having a conversation about it here helps me *a lot*.

I'll replace my form with yours and fix the shuffling bug.

I'll also review your "narrow support" option and get back to you with feedback.

Regarding a more generic solution, I've updated my [exploration branch](https://github.com/ggalmazor/javarosa/tree/issue_273_randomize_nodeset_exploration) with an implementation that will resolve an `ItemsetBinding.nodesetRef` from an `XPathPathExpr` *and* `XPathFuncExpr` depending on what the parsing process of the value of the `nodeset` attribute returns. I achieve that by passing the `FormInstance` instance and all the non-main `FormInstance` all along the chain up to `ItemsetBinding.initReferences()`, where I can build an `EvaluationContext` and eval the func.

The problem I'm getting now is what you mentioned about resolving the labels and values of each node in the nodeset. They're being interpreted as absolute paths on the instance, which will eventually throw:
```
org.javarosa.xform.parse.XFormParseException: <label> node for itemset doesn't exist! [/root/item[3]/itextId]

	at org.javarosa.xform.parse.FormInstanceParser.verifyItemsetBindings(FormInstanceParser.java:344)
	at org.javarosa.xform.parse.FormInstanceParser.verifyBindings(FormInstanceParser.java:177)
	at org.javarosa.xform.parse.FormInstanceParser.parseInstance(FormInstanceParser.java:84)
	at org.javarosa.xform.parse.XFormParser.parseDoc(XFormParser.java:465)
	at org.javarosa.xform.parse.XFormParser.parse(XFormParser.java:346)
	at org.javarosa.xform.util.XFormUtils.getFormFromInputStream(XFormUtils.java:81)
	at org.javarosa.core.test.FormParseInit.init(FormParseInit.java:62)
	at org.javarosa.core.test.FormParseInit.<init>(FormParseInit.java:48)
	at org.javarosa.xpath.expr.XPathFuncExprRandomizeTest.name(XPathFuncExprRandomizeTest.java:18)
```
The paths on those `TreeReference` objects is kind of right, no? The problem is that they're resolved against the main `FormInstance`, which can't resolve that path.

I don't know if this is a crazy idea, but maybe this could be solved by attaching a FormInstance reference to those `TreeReference` objects to always resolve them successfully.

ggalmazor:
OK, @lognaturel, I think I'm happy with the commits on this PR now. I've implemented your concept of narrow support for randomize() on itemset nodeset definitions while maintaining intact the randomize function support elsewhere.

It turned out that we already have a `Collections.shuffle(elements, random)` in Java8 :) Less code to maintain, yay! I won't use it directly, though, because it mutates the input. `RandomizeHelper. shuffle()` makes a copy of it to make it an immutable operation, which is easier to understand from the caller's perspective.

We discuss merging this one and continue exploring better options too

lognaturel:
Doesn't look so bad once you've cleaned it up! 😄

I added a question to the form to try the seeded version: [randomize.xml.txt](https://github.com/opendatakit/javarosa/files/1908622/randomize.xml.txt) and tried it in Collect.

On subsequent form launches, the orders are back to alphabetical. I realized that's expected because new values aren't serialized so I added:
```
this.randomize = ExtUtil.readBool(in);
this.randomSeed = (Long)ExtUtil.read(in, Long.class);
```
and 
```
ExtUtil.writeBool(out, this.randomize);
ExtUtil.write(out, new ExtWrapNullable(this.randomSeed == null ? null : new ExtWrapTagged(this.randomSeed)));
```

The second list order is always:
```
Banana
Apple
Elderberry
Fig
Durian
Cherimoya
```
which I have verified is expected with a seed of 42.

This is ready to go once you add the serialization. I think it's easy enough to merge now and replace when/if a more generic approach is figured out. We could also leave the PR open and merge right before the next release if we don't have that solution yet. What do you think? 

How about you open another PR with the exploration you're doing so we can continue that conversation.

ggalmazor:
@lognaturel, I've updated the PR adding the custom implementation of Fisher-Yates & Park-Miller algorithms. I've added a lot of JavaDoc and inlined comments to give context about the cross-platform support of the shuffling operation. I don't know if we need to expand on any topic.

ggalmazor:
@lognaturel, @yanokwa, I've added the test it was missing and now it's ready for review

src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -466,6 +466,17 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
                 (String) argVals[0],
                 args.length == 3 ? Encoding.from((String)argVals[2]) : Encoding.BASE64
             );
+        } else if (name.equals("randomize")) {
```
dcbriccetti: I ran all tests with coverage, and we don’t seem to get here. Should we?

screen shot 2018-04-25 at 6 57 49 pm

test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+        return true;
+    }
+
+    private static TreeReference absoluteRef(String path) {
+        TreeReference tr = new TreeReference();
+        tr.setRefLevel(REF_ABSOLUTE);
+        tr.setContext(CONTEXT_ABSOLUTE);
+        tr.setInstanceName(null);
+        Arrays.stream(path.split("/"))
```
dcbriccetti: Oh, I didn’t realize we could use modern Java features in JavaRosa because it is linked into Collect. But the tests *don’t* run on Collect. Right?

lognaturel: They don't so this should be 👌!

src/org/javarosa/xpath/expr/XPathFuncExpr.java
ggalmazor: I suppose we should. The outside-in tests I have written use `randomize()` in a nodeset definition, which isn't eval'ed on this file. The only problem is that I can't think of a use for 'randomize()` out of that context.

If we can think of an example where `randomize()` could have sense outside a nodeset definition, I can write a new test. Any ideas, @lognaturel, @dcbriccetti, @yanokwa?

test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
dcbriccetti: Android Studio, with Collect, alerts us when we try to add a feature that is not present in the API versions we support. IDEA, with JavaRosa, does not. Stream is not available.

src/org/javarosa/xpath/expr/XPathFuncExpr.java
dcbriccetti: Or maybe we delete this code block, 469–479?
I’m looking for how to hook in with the debugger and watch this new code run when processing, for example, this:
`<itemset nodeset="randomize(instance('fruit')/root/item)">`

ggalmazor: That code won't touch that block because we don't actually eval the nodeset. We would need to come up with some other use for `randomize()` that is not inside an itemset nodeset definition. 

Let's give it some days. Maybe @lognaturel or @yanokwa can come up with some reasonable use for `randomize()` and I can write a test that uses that block.

test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
ggalmazor: I understand. So, even if we can use Java8, not all the API is supported by some of the Android API versions we need to support. I didn't know about all that :)

lognaturel: I'm not totally sure whether this applies to dependencies! Would be good to check and document this. I /think/ what we're importing from Collect is bytecode so it may not matter. https://developer.android.com/studio/write/java8-support#supported_features doesn't make it clear to me either way but perhaps one of you will be able to know the naswer from there.

src/org/javarosa/xpath/expr/XPathFuncExpr.java
lognaturel: I can't think of a super meaningful example and that's why this narrow support isn't so bad.

But just for testing, you can do something silly in a calculate like `max(randomize(/path/to/int/in/repeat))`. You can use any of the [functions](https://opendatakit.github.io/xforms-spec/#xpath-functions) that take a nodeset parameter. Order doesn't matter for any of them so they won't be super meaningful but they'll exercise the code.

lognaturel:
Let's start doing a quick QA pass on JR pull requests to match our processes on other tools and avoid some mistakes like the ones we made recently! @kkrawczyk123, this would be a really good one to start with when you have a chance. Let us know in Slack if you need any support getting this set up in Collect. Thanks!

src/org/javarosa/xpath/expr/XPathFuncExpr.java
ggalmazor: Just to make it a little less silly, how could we take the first element of the output of `randomize(/path/to/int/in/repeat)`?

lognaturel: Good idea! I think you can add `[1]` or `[position() = 1]`. That is actually useful. 👍

kkrawczyk123:
### Tested with success!

Verified on Androids: 4.1, 4.2, 4.4, 5.1, 6.0, 7.0 and 8.1

lognaturel:
I'm really excited for this feature!!

Issue #293, opened 2018-04-04
current() in choice filter is inconsistent with current() elsewhere


lognaturel:
This has come up because there's some desire to have pyxform output relative references in repeats: https://github.com/XLSForm/pyxform/issues/4. That would be in better conformance with W3C XForms and would allow for choice filters in repeats. Currently ${fieldname} syntax is always expanded to `/absolute/path/to/node` even when in a repeat. In W3C XForms, that should always evaluate to the first repeated element, not to the one in the current context. In ODK XForms, the repeat context is considered even when an absolute path is used. https://opendatakit.github.io/xforms-spec/#a-big-deviation-with-xforms

Unfortunately, JavaRosa interprets `current()` differently in a `choice-filter`. This was discussed at lengths in https://forum.opendatakit.org/t/current-in-xpath-predicate/4982 with no real conclusion reached for the broader ODK community.

As Mitch explained [in that thread](https://forum.opendatakit.org/t/current-in-xpath-predicate/4982/7):

> if you use a filter condition in the nodeset specification of a field "myfield", within that nodeset specification `"current()/.."` refers to the parent of the group enclosing the "myfield" field. It does not refer to the group enclosing the "myfield" field.

To be super explicit, `current()` in a `choice-filter` should refer to the node it is used from. Instead, it refers to the group enclosing the node.

As the thread discusses, the big challenge here is how to start generating relative references in form builders without resulting forms breaking on a bunch of devices that may have older clients.

One option is to change the `current()` behavior as quickly as possible. I expect relative paths in choice filters continue to be relatively rare. Then after some period of time has passed (a year?), form builders can start using relative paths and we can communicate to users that upgrading their clients is really important.

There are lots of ideas with flags and other ways of marking compatibility in the thread.

MartijnR:
I wonder if doing a simple analysis to detect whether the XForm relies on the bug as follows would be sufficient to switch to `current-bug-compliant` mode in JavaRosa: 

If expression matches `/current\(\)\/[^.]+/` switch to `current-bug-compliant` mode.

This relies on the syntax where `current()/` is not followed by a `.` (or `..`). It may catch close to 100% of all the (very few) hand-coded forms that use current() currently. I would much prefer such a solution to tagging XForms for (likely) not relying on bugs.

lognaturel:
[...]

PR #298, opened 2018-04-16
WIP - Issue 273 broad support for randomize

ggalmazor:
This branch is based on #296 and the actual changes start with commit https://github.com/opendatakit/javarosa/commit/04745a54af978401a98a4453a009232de5ceb754

This is an exploration branch I'm working on to add broad support for the randomize function().

- Now an `ItemsetBinding` can be on one of these two states:
  - All its `*Ref` members have been resolved on initialization because we have a path in the nodeset definition
  - A nodeset has been eval-ed out from the nodeset defintion, which holds an expression.
- By making all the `*Ref` members private, we force caller sites to use getter methods that will check the presence of the eval-ed Nodeset or the `*Ref` members and act accordingly.

Currently, this implementation passes all the tests from #296 except for:
- `XPathFuncExprRandomizeTest.the_same_seeded_field_in_different_instances_from_deserialized_forms_gets_the_same_order_of_choices`
- `XPathFuncExprRandomizeTest.the_same_field_without_seed_in_different_instances_gets_a_different_order_of_choices`

I know it's heavy reading, but I would appreciate any thoughts on the strategy. Please, ask anything I should have to expand on this explanation.

Tagging @lognaturel and @dcbriccetti for advice.

### DIFF
--- a/src/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/org/javarosa/core/model/ItemsetBinding.java
@@ -1,11 +1,15 @@
 package org.javarosa.core.model;
 
+import static java.util.stream.Collectors.toMap;
 import static org.javarosa.core.model.instance.FormInstance.unpackReference;
 
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IConditionExpr;
 import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.model.instance.TreeReference;
@@ -20,6 +24,9 @@ import org.javarosa.core.util.externalizable.Externalizable;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.model.xform.XPathReference;
 import org.javarosa.xpath.XPathConditional;
+import org.javarosa.xpath.XPathNodeset;
+import org.javarosa.xpath.expr.XPathExpression;
+import org.javarosa.xpath.expr.XPathFuncExpr;
 import org.javarosa.xpath.expr.XPathPathExpr;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -105,11 +112,24 @@ public class ItemsetBinding implements Externalizable, Localizable {
         // To construct the xxxRef, we need the full model, which wasn't available before now.
         // Compute the xxxRefs now.
         XPathConditional nodesetExpr = (XPathConditional) this.nodesetExpr;
-        XPathPathExpr expr = (XPathPathExpr) nodesetExpr.getExpr();
-        TreeReference reference = expr.getReference(true);
-        XPathReference ref = new XPathReference(reference);
-        IDataReference absRef = FormDef.getAbsRef(ref, contextRef);
-        nodesetRef = unpackReference(absRef);
+        XPathExpression expr = nodesetExpr.getExpr();
+        if (expr instanceof XPathPathExpr) {
+            TreeReference reference = ((XPathPathExpr)expr).getReference(true);
+            XPathReference ref = new XPathReference(reference);
+            IDataReference absRef = FormDef.getAbsRef(ref, contextRef);
+            nodesetRef = unpackReference(absRef);
+        }
+        if (expr instanceof XPathFuncExpr) {
+            XPathFuncExpr f = (XPathFuncExpr) expr;
+            Stream<DataInstance> stream = nonMainInstances.stream();
+            Map<String, DataInstance> nmis = stream.collect(toMap(
+                DataInstance::getName,
+                di -> di
+            ));
+            EvaluationContext ec = new EvaluationContext(instanceModel, nmis);
+            XPathNodeset nodeset = (XPathNodeset) f.eval(ec);
+            nodesetRef = nodeset.getRefAt(0);
+        }
         if (labelExpr != null) {
             XPathConditional labelExpr = (XPathConditional) this.labelExpr;
             XPathPathExpr expr1 = (XPathPathExpr) labelExpr.getExpr();
--- a/src/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/org/javarosa/core/model/ItemsetBinding.java
@@ -1,5 +1,7 @@
 package org.javarosa.core.model;
 
+import static org.javarosa.xform.parse.RandomizeHelper.shuffle;
+
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
@@ -52,6 +54,9 @@ public class ItemsetBinding implements Externalizable, Localizable {
                                    //not serialized -- set by QuestionDef.setDynamicChoices()
     private List<SelectChoice> choices; //dynamic choices -- not serialized, obviously
 
+    public boolean randomize = false;
+    public Long randomSeed = null;
+
     public List<SelectChoice> getChoices () {
         return choices;
     }
@@ -61,7 +66,13 @@ public class ItemsetBinding implements Externalizable, Localizable {
             logger.warn("previous choices not cleared out");
             clearChoices();
         }
-        this.choices = choices;
+        this.choices = randomize ? shuffle(choices, randomSeed) : choices;
+
+        if (randomize) {
+            // Match indices to new positions
+            for (int i = 0; i < choices.size(); i++)
+                choices.get(i).setIndex(i);
+        }
 
         //init localization
         if (localizer != null) {
--- a/src/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/org/javarosa/core/model/ItemsetBinding.java
@@ -7,6 +7,7 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.List;
 import org.javarosa.core.model.condition.IConditionExpr;
+import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.core.model.util.restorable.RestoreUtils;
 import org.javarosa.core.services.locale.Localizable;
@@ -100,7 +101,7 @@ public class ItemsetBinding implements Externalizable, Localizable {
         return relRef != null ? RestoreUtils.xfFact.refToPathExpr(relRef) : null;
     }
 
-    public void initReferences(QuestionDef q) {
+    public void initReferences(DataInstance instanceModel, List<DataInstance> nonMainInstances, QuestionDef q) {
         // To construct the xxxRef, we need the full model, which wasn't available before now.
         // Compute the xxxRefs now.
         XPathConditional nodesetExpr = (XPathConditional) this.nodesetExpr;
--- a/src/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/org/javarosa/core/model/ItemsetBinding.java
@@ -1,7 +1,5 @@
 package org.javarosa.core.model;
 
-import static org.javarosa.xform.parse.RandomizeHelper.shuffle;
-
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
@@ -54,9 +52,6 @@ public class ItemsetBinding implements Externalizable, Localizable {
                                    //not serialized -- set by QuestionDef.setDynamicChoices()
     private List<SelectChoice> choices; //dynamic choices -- not serialized, obviously
 
-    public boolean randomize = false;
-    public Long randomSeed = null;
-
     public List<SelectChoice> getChoices () {
         return choices;
     }
@@ -66,13 +61,7 @@ public class ItemsetBinding implements Externalizable, Localizable {
             logger.warn("previous choices not cleared out");
             clearChoices();
         }
-        this.choices = randomize ? shuffle(choices, randomSeed) : choices;
-
-        if (randomize) {
-            // Match indices to new positions
-            for (int i = 0; i < choices.size(); i++)
-                choices.get(i).setIndex(i);
-        }
+        this.choices = choices;
 
         //init localization
         if (localizer != null) {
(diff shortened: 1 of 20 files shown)


src/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -183,4 +198,54 @@ public void writeExternal(DataOutputStream out) throws IOException {
...
+                TreeReference refAt = ns.getRefAt(0);
+                refAt.setInstanceName(ns.instance.getName());
+                return refAt;
+            })
+            .orElseGet(() -> nodesetRef.orElse(null));
+    }
+
+    public TreeReference getLabelRef() {
+        return nodeset
+            .map(ns -> computeRef(ns.getRefAt(0), labelExpr).get())
```
lognaturel: Am I understanding correctly that this will just get the reference for the first node in the nodeset which could be completely different than the ones for other nodes in the set but that it doesn't matter because it's just used for surface error checking?

lognaturel:
Wow, this is cool! 

8a775dd is kind of a grab bag of changes so I'm not sure I was able to fully appreciate all the different decisions that went into it. The main thing I'm wondering about is the implications of only looking at the first item in a nodeset when computing refs. Given how they're used it does seem ok but I'd like to know more about how you were thinking about that. Otherwise, the general concept of the two states seems good. I think making the fields private is fine even though it does change a public interface.

There are a couple things that need to be changed/added at some point:
- Cross-platform PRNG (likely Park Miller PRNG pending conclusion of XForms spec discussion)
- Support paths on the seed field as well as numbers as you mentioned in #296 

These can be kept for the end but just wanted to make sure they were written down somewhere.

src/org/javarosa/core/model/ItemsetBinding.java
ggalmazor: It looks like it, but I couldn't be sure. I'm thinking that moving all the validation and coherence checks inside ItemsetBinding to make sure about that.

ggalmazor:
@dcbriccetti maybe you could take a look to `XPathFuncExprRandomizeTest`. Right now, there are two tests that fail:
- `the_same_field_without_seed_in_different_instances_gets_a_different_order_of_choices` is focused on obtaining two instances from the same form and checking that, since we don't give a shuffling seed to the field `fruit1`, the choices you get are different.
- `the_same_seeded_field_in_different_instances_from_deserialized_forms_gets_the_same_order_of_choices` is focused on checking that, given a shuffling seed, deserialized forms will get the same ordering.

I guess that first, we should be sure that the testing strategy is correct taking account of the goals of those tests. I'm starting to suspect that I'm not running the correct code to set the scenarios I want.

src/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -32,31 +42,37 @@
...
     public IConditionExpr valueExpr;   //path expression for value; may be relative, no predicates (must be relative if copy mode)
 
     private TreeReference destRef; //ref that identifies the repeated nodes resulting from this itemset
-                                   //not serialized -- set by QuestionDef.setDynamicChoices()
+    //not serialized -- set by QuestionDef.setDynamicChoices()
     private List<SelectChoice> choices; //dynamic choices -- not serialized, obviously
 
-    public List<SelectChoice> getChoices () {
+    private DataInstance instanceModel;
+    private Optional<XPathNodeset> nodeset;
```
dcbriccetti: Will [Optional](https://developer.android.com/reference/java/util/Optional)—added in API level 24—work in Collect?
```
    defaultConfig {
        applicationId('org.odk.collect.android')
        minSdkVersion(16)
        targetSdkVersion(22)
```

lognaturel: Yes, `targetSdk` just determines forwards-compatibility UI features. We're still compiling with API 26.

lognaturel: I said that with high confidence and then I realized that's not true for Android features. Like I said in another comment, I'm not really sure how all that interacts with pure Java dependencies. I tried to add a jar with this code to Collect and it does compile.

lognaturel: Hmm, seems possible.

dcbriccetti: Did you get an answer to Why Guava Optional over Java 8?
https://github.com/opendatakit/collect/pull/1785#pullrequestreview-89992396

dcbriccetti:
I just noticed, after some refactoring to help my understanding of `XPathPathExpr`’s
`public XPathNodeset eval (DataInstance m, EvaluationContext ec)`,
that the formal parameter `m` is never read. It is either reassigned (which I don’t like as a practice), or there is an exception.

This observation is probably not relevant to the current work, but I thought others might find it interesting.

src/org/javarosa/core/model/ItemsetBinding.java
dcbriccetti: Well, here’s what it looks like:

```
05-05 23:52:44.902 21563-21696/org.odk.collect.android E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #2
                                                                         Process: org.odk.collect.android, PID: 21563
                                                                         java.lang.RuntimeException: An error occurred while executing doInBackground()
                                                                             at android.os.AsyncTask$3.done(AsyncTask.java:309)
                                                                             at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:354)
                                                                             at java.util.concurrent.FutureTask.setException(FutureTask.java:223)
                                                                             at java.util.concurrent.FutureTask.run(FutureTask.java:242)
                                                                             at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:234)
                                                                             at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
                                                                             at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
                                                                             at java.lang.Thread.run(Thread.java:818)
                                                                          Caused by: java.lang.NoClassDefFoundError: Failed resolution of: Ljava/util/Optional;
```

dcbriccetti: Maybe look at Guava’s Optional?

dcbriccetti:
I wanted to step through deserialization of this Randomize Test form in Collect, and compare that with what’s happening in the test, but I’ll have to wait for you to remove java.util.Optional. So that’s all for today on this.

src/org/javarosa/core/model/ItemsetBinding.java
ggalmazor: I don't know... We would have to add the dependency and, eventually we would have to remove it to migrate to native Optionals... it feels like too much just to use Optional in two classes. I think we can wait.

dcbriccetti: Makes sense.

ggalmazor: Alright, just pushed changes from your branch and removed Optional too

dcbriccetti: In that case I’ll have another go with Collect (unless you are already ahead of me with the problem solving).

ggalmazor: Please, go ahead :)

dcbriccetti: Darn. ;-)

dcbriccetti: You’re still using Stream. :-(

dcbriccetti: None of your modern sorcery in Collect! ;-)
```
    public void initReferences(DataInstance instanceModel, List<DataInstance> nonMainInstances, QuestionDef q) {

        Stream<DataInstance> stream = nonMainInstances.stream();
```

ggalmazor: Do we have to remove that too? :D

dcbriccetti: I’ll give you the answer in the form of a stack trace.
```
Caused by: java.lang.NoSuchMethodError: No interface method stream()Ljava/util/stream/Stream; in class Ljava/util/List; or its super classes (declaration of 'java.util.List' appears in /system/framework/core-libart.jar)
                                                                             at org.javarosa.core.model.ItemsetBinding.initReferences(ItemsetBinding.java:120)
                                                                         
```

ggalmazor: Is it OK to use Stream in tests?

ggalmazor: Nevermind, I'll get rid of those too

dcbriccetti: I assume so, since this is a runtime error.

dcbriccetti:
Collect catches the same failure when reading the cached form. But it discards the cache file, re-parses, successfully, and carries on. OK, that’s all from me for today. I’ll resume within 12 hours, most likely, unless you’ve solved it by then.
```
E/FormDefCache: org.javarosa.xpath.XPathMissingInstanceException: XPath evaluation: Instance referenced by instance(fruit)/root/item does not exist
                    at org.javarosa.xpath.expr.XPathPathExprEval.getDataInstance(XPathPathExprEval.java:63)
                    at org.javarosa.xpath.expr.XPathPathExprEval.eval(XPathPathExprEval.java:37)
                    at org.javarosa.xpath.expr.XPathPathExpr.eval(XPathPathExpr.java:196)
                    at org.javarosa.xpath.expr.XPathPathExpr.eval(XPathPathExpr.java:59)
                    at org.javarosa.xpath.expr.XPathFuncExpr.eval(XPathFuncExpr.java:176)
                    at org.javarosa.xpath.expr.XPathExpression.eval(XPathExpression.java:33)
                    at org.javarosa.core.model.ItemsetBinding.initReferences(ItemsetBinding.java:145)
                    at org.javarosa.core.model.FormDef.updateItemsetReferences(FormDef.java:1500)
                    at org.javarosa.core.model.FormDef.setInstance(FormDef.java:261)
                    at org.javarosa.core.model.FormDef.readExternal(FormDef.java:1189)
                    at org.odk.collect.android.utilities.FormDefCache.deserializeFormDef(FormDefCache.java:120)
                    at org.odk.collect.android.utilities.FormDefCache.readCache(FormDefCache.java:85)
                    at org.odk.collect.android.tasks.FormLoaderTask.createFormDefFromCacheOrXml(FormLoaderTask.java:225)
                    at org.odk.collect.android.tasks.FormLoaderTask.doInBackground(FormLoaderTask.java:126)
                    at org.odk.collect.android.tasks.FormLoaderTask.doInBackground(FormLoaderTask.java:70)
```

dcbriccetti:
```
<randomValue/>
<seededRandomValue/>
```
These appear empty in the form in Collect.

test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
```diff
@@ -0,0 +1,179 @@
...
+        formDef = fpi.getFormDef();
+    }
+
+    @Test
+    public void fields_without_seed_in_the_same_form_get_a_different_order_of_choices() {
+        initializeNewInstance(formDef);
+        List<SelectChoice> choices1 = getSelectChoices(formDef, "/randomize/fruit1");
+        List<SelectChoice> choices2 = getSelectChoices(formDef, "/randomize/fruit2");
+
+        assertFalse(nodesEqualInOrder(choices1, choices2));
```
dcbriccetti: On average this test will only work 99.99998842592592% of the time (if my math is right)
`100*1-6*pow((1/factorial(6)), 2)`

ggalmazor: I'll put that as a comment for future reference :)

ggalmazor:
>    &lt;randomValue/>
>    &lt;seededRandomValue/>
>These appear empty in the form in Collect.

Can we rule out any problem in the form definition?

ggalmazor:
> `E/FormDefCache: org.javarosa.xpath.XPathMissingInstanceException: XPath evaluation: Instance referenced by instance(fruit)/root/item does not exist`

I'm getting the same trace... Before digging too deep, my first concern is to verify that both the tests I mentioned are OK. Let's be sure I'm not doing anything funny there...

lognaturel:
@ggalmazor Following my exploration in #293 I feel like I can actually engage with this. Right now it's a bit hard to see what changes are meaningful. I'm wondering whether you could perhaps pull some of the formatting/cleanup changes into their own PR that could be merged immediately? I'm also seeing some good testing helpers that could be extracted so they can be used right away.

test/org/javarosa/xpath/expr/XPathFuncExprRandomizeTest.java
```diff
@@ -0,0 +1,179 @@
...
+            return false;
+
+        for (int i = 0; i < left.size(); i++)
+            if (!left.get(i).getValue().equals(right.get(i).getValue()))
+                return false;
+
+        return true;
+    }
+
+    private static TreeReference absoluteRef(String path) {
```
lognaturel: I tried to solve the same problem at https://github.com/opendatakit/javarosa/commit/ee5527fd3c2d49549e288de6beaab5df48068be8. I think your approach always works, right? If so, it's simpler than mine and is a good example of something we can set up in some kind of test helper.

ggalmazor: Yes, I'm sure it works at least in the context of this test. If you think this is useful, I can extract that into a helper class.

ggalmazor:
OK, @lognaturel! :)

I'll do as you ask. I'm very excited to see how we can turn this around :)

ggalmazor:
@lognaturel, I've reviewed the commits to see what can be salvaged and there's only one commit that reformats `ItemsetBinding`. To make things worse, it can't be merged cleanly because the branch has diverged too much from master. 

I could recreate this PR's branch to branch off of a recent master and have just one commit with the reformatting of that file, but it feels like too much. What do you think?

PR #300, opened 2018-04-20
SMSSerializer updated to latest spec. 

jd-alexander:
This pull request contains modifications that were made to the SMSSerializer class that handles the serialization of a form to an SMS string based on the [discussion of the SMS Spec](https://forum.opendatakit.org/t/send-submissions-via-sms/12166/11). 

I have written several tests that utilize a sample form that has an instance block with elements that should be included in the SMS output. This output is then compared to several scenarios that ensure the correct SMS is being generated based on the tags and if the respective elements have answers.

N/A

No.


jd-alexander:
Just reviewing the diff and seeing some extra lines. Gonna get to them shortly :sweat_smile:

lognaturel:
😊Thanks for getting those extra lines! It's looking really good.

Any chance you could cherry pick the commits from Medic? That would preserve attribution and the very good commit messages. You can do that by adding https://github.com/medic/javarosa as a remote and checking out the master branch (you'll probably want to name it something like `medic/master` locally). Then you can do `git cherry-pick 90c46a6a3fc19b1b75783cb27cfa907d37099cca..867ee6d9c3327f8d58924260d741471b330e1f2a`.

You may also want to look to those commit messages for good examples. It's helpful to have a really short summary message and a longer explanation in the extended comment.

jd-alexander:
Okay, I am going to revisit that approach. I tried cherrypicking at first by doing it and then creating a patch to apply to ODK's version of Javarosa but it failed for some strange reason so that's the reason why I ended up just pulling the changes in. Will redo it tomorrow though and share any issues I have here. Thanks :)

jd-alexander:
@lognaturel I added the commit history! Should be ready for a review and merge now 😄

dcbriccetti:
I’ll just note here that `SMSSerializingVisitor` should be `SmsSerializingVisitor`. Just to slow “normalization” (a term from the news) of the all-caps initialisms in identifiers.

jd-alexander:
@dcbriccetti noted :+1:

lognaturel:
History looks great, thanks! The resulting changes will need to be moved from `core/src/org/javarosa/model/xform/` to `src/org/javarosa/model/xform/`. Right now the one with the changes is not actually being used.

@dcbriccetti You are correct that it should be `SmsSerializingVisitor` but since it's a legacy class it's not a change that should be made now. Seems you may have expected that but just to make it explicit.

What we've done for other feature additions and that I think we should do here is to wait until the Collect code is approved and merge to merge here. That makes it easier to keep all changes together and make sure they are fully verified before merge.

dcbriccetti:
Is this ready for review, or are we waiting for something?

jd-alexander:
Hey Dave I think it’s ready for a review. For the most part the functionality has been verified in Collect.

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+                    //of while condition  checks 2 indexes ahead.
+                    answer += sms.substring(sms.length() - 1, sms.length());
+
+                    break;
+                }
+            }
+        }
+
+        //Prints each tagged answer to the console.
+        //System.out.println(answer);
```
dcbriccetti: I don’t think we should have commented-out code.

test/org/javarosa/test/utils/Utils.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public static String convertToString(InputStream inputStream) throws IOException {
+        BufferedInputStream bis = new BufferedInputStream(inputStream);
+        ByteArrayOutputStream buf = new ByteArrayOutputStream();
+        int result = bis.read();
+        while (result != -1) {
+            buf.write((byte) result);
+            result = bis.read();
+        }
+        // StandardCharsets.UTF_8.name() > JDK 7
```
dcbriccetti: I don’t understand this. Is it meant to be a comment, or commented.out (pseudo)code?

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Performs a comprehensive suite of tests on the results of a form
+ * being serialized to the SMS format.
+ */
+public class SMSSerializingVisitorTest {
+    private FormParseInit formParser;
+    private String SMS;
+    private static final String delimiter = " ";
+    private FormEntryController formEntryController;
+    private FormInstance formInstance;
```
dcbriccetti: Following these IDEA suggestions can simplify the code.

screen shot 2018-06-18 at 11 45 39 am

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Performs a comprehensive suite of tests on the results of a form
+ * being serialized to the SMS format.
+ */
+public class SMSSerializingVisitorTest {
+    private FormParseInit formParser;
+    private String SMS;
```
dcbriccetti: `sms` would be a better name, no?

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.javarosa.test.utils.Utils.convertToString;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Performs a comprehensive suite of tests on the results of a form
+ * being serialized to the SMS format.
+ */
+public class SMSSerializingVisitorTest {
```
dcbriccetti: I realize that SMSSerializingVisitor is an existing class, so I wouldn’t ask you to correct its name to SmsSerializingVisitor for this PR, but SMSSerializingVisitorTest is a new class, so how about calling it SmsSerializingVisitorTest? I think following the naming convention violation is more important than temporary inconsistency between the test class name and the class being tested.

test/org/javarosa/test/utils/Utils.java
```diff
@@ -0,0 +1,24 @@
+package org.javarosa.test.utils;
+
+import java.io.BufferedInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class Utils {
```
dcbriccetti: You can remove this file.

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+        formParser.setFormToParse(new File(PathConst.getTestResourcePath(), "sms_form.xml").toString());
+        formEntryController = formParser.getFormEntryController();
+        formInstance = formEntryController.getModel().getForm().getInstance();
+
+        SMSSerializingVisitor serializer = new SMSSerializingVisitor();
+
+        ByteArrayPayload payload = (ByteArrayPayload) serializer.createSerializedPayload(formInstance);
+
+        SMS = convertToString(payload.getPayloadStream());
+        SMS = SMS.replace("\\", "").replace("\\\\", "\\");
```
dcbriccetti: You can replace the two lines with
```
sms = new String(payload.getPayloadBytes(), "UTF-8").replace("\\", "").replace("\\\\", "\\");
```
See my commits here for details: https://github.com/dcbriccetti/javarosa/commits/my-jd-sms
Feel free to cherry pick.

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+    @Test
+    public void testTagsAreBeingSerialized() {
+        for (String tag : getAnswerTags()) {
+            assertTrue(SMS.contains(tag));
+        }
+    }
+
+    /**
+     * Checks to see if all answers for the selected tags exist. This
+     * targets answers of various data types  and structures such as
```
dcbriccetti: two spaces in a row

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+     * the SMS.
+     * the CTY tag is attached to  the country  question.
+     */
+    @Test
+    public void testTagWithNoAnswer() {
+        assertFalse(SMS.contains("CTY"));
+    }
+
+    /**
+     * returns a list of all the tags that are bounded to
```
dcbriccetti: You probably want “bound”, past tense of “bind”. “Bounded” has to do with boundaries.

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+        List<String> tags = new ArrayList<>();
+
+        tags.add("FN");
+        tags.add("LN");
+        tags.add("DOB");
+        tags.add("CN");
+        tags.add("PIC");
+
+        return tags;
+    }
```
dcbriccetti: Replace body with `return Arrays.asList("FN", "LN", "DOB", "CN", "PIC");`

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * This function  checks to see if  a  tagged node within the
```
dcbriccetti: Please check all comments for extra spaces.

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        String answer = "";
+        int currentIndex;
+        int delimiterSize = 1;
+
+        Pattern pattern = Pattern.compile(tag);
+        Matcher matcher = pattern.matcher(sms);
+
+        /**
```
dcbriccetti: This is not a place for a doc comment, so /* instead of /**.

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
```diff
@@ -0,0 +1,161 @@
...
+     * @param tag
+     * @param sms
+     * @return true if  an answer exists after a tag.
+     */
+    private static boolean taggedAnswerExists(String tag, String sms) {
+        if (!sms.contains(tag)) {
+            return false;
+        }
+
+        String answer = "";
```
dcbriccetti: At the risk of optimizing prematurely, consider using StringBuilder here instead of String. Strings are immutable, so a new object must be allocated (and later garbage collected) for every change.
screen shot 2018-06-18 at 12 35 42 pm

core/src/org/javarosa/model/xform/SMSSerializingVisitor.java
```diff
@@ -0,0 +1,238 @@
...
+		delimiter = root.getAttributeValue("", "delimiter");
+		if ( delimiter == null ) {
+			// for the spelling-impaired...
+			delimiter = root.getAttributeValue("", "delimeter");
+		}
+		prefix = root.getAttributeValue("", "prefix");
+
+		xmlns = (xmlns != null)? xmlns : " ";
+		delimiter = (delimiter != null ) ? delimiter : " ";
+		prefix = (prefix != null) ? prefix : " ";
```
lognaturel: Why default the prefix to space rather than to blank?

core/src/org/javarosa/model/xform/SMSSerializingVisitor.java
```diff
@@ -0,0 +1,238 @@
...
+import org.kxml2.kdom.Node;
+
+/**
+ * A modified version of Clayton's XFormSerializingVisitor that constructs
+ * SMS's.
+ *
+ * @author Munaf Sheikh, Cell-Life
+ *
+ */
+public class SMSSerializingVisitor implements IInstanceSerializingVisitor {
```
lognaturel: Hmm, this file shouldn't exist, right? It's in a /core root folder that doesn't exist in this project.

jd-alexander: Okay will move it to the right location :)

test/org/javarosa/model/xform/SMSSerializingVisitorTest.java
jd-alexander: Noted.

jd-alexander:
thanks for the comprehensive review @dcbriccetti! Making these changes now.

test/org/javarosa/model/xform/SmsSerializingVisitorTest.java
```diff
@@ -10,123 +10,110 @@
...
         for (String tag : getAnswerTags()) {
-            assertTrue(taggedAnswerExists(tag, SMS));
+            assertTrue(taggedAnswerExists(tag, sms));
         }
     }
 
     /***
-     * Checks to see an answer that doesn't have a tag is present .
-     * Placeholder  is an answer bounded to the <maiden_name> tag
+     * Checks to see than an answer doesn't have a tag is present.
```
dcbriccetti: Not “than”
How about this:
Ensures that the SMS doesn’t contain a specific tag

test/org/javarosa/model/xform/SmsSerializingVisitorTest.java
```diff
@@ -10,123 +10,110 @@
...
+     * Placeholder is an answer bounded to the <maiden_name> tag
```
dcbriccetti: “bounded” has a different meaning than “bound” (past tense of “bind”), which is what you probably mean.

src/org/javarosa/core/services/transport/payload/ByteArrayPayload.java
```diff
@@ -104,34 +97,31 @@ public void writeExternal(DataOutputStream out) throws IOException {
...
     public <T> T accept(IDataPayloadVisitor<T> visitor) {
         return visitor.visit(this);
     }
 
-    /*
-     * (non-Javadoc)
-     * @see org.javarosa.core.services.transport.IDataPayload#getPayloadId()
-     */
+    public byte[] getPayloadBytes() {
+        return payload;
```
dcbriccetti: I realized that this passes back a byte array, which could be modified. Sorry.
What do you think of replacing this method with a `toString` method that returns the String that you need, instead of making the string in the caller?

test/org/javarosa/model/xform/SmsSerializingVisitorTest.java
```diff
@@ -10,123 +10,110 @@
...
-    private FormInstance formInstance;
 
     @Before
     public void setup() throws IOException {
-        formParser = new FormParseInit();
+        FormParseInit formParser = new FormParseInit();
         formParser.setFormToParse(new File(PathConst.getTestResourcePath(), "sms_form.xml").toString());
-        formEntryController = formParser.getFormEntryController();
-        formInstance = formEntryController.getModel().getForm().getInstance();
+        FormEntryController formEntryController = formParser.getFormEntryController();
```
dcbriccetti: We have a new helper method to find form files, and a new convention where we place the form ... well, see this commit comment:
https://github.com/dcbriccetti/javarosa/commit/272ca8cd5b09659ea8c8fc7dbf36dd81c3ebaefa

src/org/javarosa/model/xform/SMSSerializingVisitor.java
```diff
@@ -133,83 +127,83 @@ public IDataPayload createSerializedPayload(FormInstance model, IDataReference r
     public void visit(FormInstance tree) {
         nodeSet = new String();
 
-        //TreeElement root = tree.getRoot();
-        TreeElement root = tree.resolveReference(rootRef);
+        TreeElement root = tree.getRoot();
```
dcbriccetti: Just checking that you really meant to uncomment this line and remove the line below it. If so, what does that change do?

src/org/javarosa/model/xform/SMSSerializingVisitor.java
```diff
@@ -133,83 +127,83 @@ public IDataPayload createSerializedPayload(FormInstance model, IDataReference r
...
+        delimiter = (delimiter != null) ? delimiter : " ";
         prefix = (prefix != null) ? prefix : " ";
 
-        //Don't bother adding any delimiters, yet. Delimiters are
-        //added before tags/data
-        theSmsStr = prefix;
+        theSmsStr = prefix.concat(delimiter);
 
-        // serialize each node to get it's answers
+        // serialize each node (and it's children) to get it's answers
```
dcbriccetti: To make “it” possessive, use “its”. “It’s” is only a contraction for “it is”.

src/org/javarosa/model/xform/SMSSerializingVisitor.java
```diff
@@ -133,83 +127,83 @@ public IDataPayload createSerializedPayload(FormInstance model, IDataReference r
...
         for (int j = 0; j < root.getNumChildren(); j++) {
-            TreeElement tee = root.getChildAt(j);
-            String e = serializeNode(tee);
-            if(e != null) {
-                theSmsStr += e;
+            TreeElement treeElement = root.getChildAt(j);
+            if (treeElement.isLeaf() && treeElement.getAttribute("", "tag") != null) {
+                String result = serializeNode(treeElement);
+                if (result != null) {
+                    theSmsStr += result;
```
dcbriccetti: I think we should improve on repeatedly mutating this field here. Let the method be pure (not modify any fields). Here’s what I’m thinking: 
https://github.com/dcbriccetti/javarosa/commit/5f016622a1ed0523b9a5f54e301beb9bbd2f8101

test/org/javarosa/model/xform/SmsSerializingVisitorTest.java
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Performs a comprehensive suite of tests on the results of a form
+ * being serialized to the SMS format.
+ */
+public class SmsSerializingVisitorTest {
+    private String sms;
+    private static final String delimiter = " ";
+
+    @Before
+    public void setup() throws IOException {
```
dcbriccetti: I know it’s rare to see a method spelled `setUp` (instead of setup), but `setUp` is correct because this method name is a verb, and it’s a phrasal verb made of two words. So we use camelcase. If it were a noun or adjective, it would be `setup`.

test/org/javarosa/model/xform/SmsSerializingVisitorTest.java
```diff
@@ -0,0 +1,145 @@
...
+    public void testTagWithNoAnswer() {
+        assertFalse(sms.contains("CTY"));
+    }
+
+    /**
+     * returns a list of all the tags that are bounded to
+     * nodes that have answers.
+     */
+    private static List<String> getAnswerTags() {
+        return Arrays.asList("FN", "LN", "DOB", "CN", "PIC");
```
dcbriccetti: Come to think of it, this should be a constant field, ANSWER_TAGS, don’t you think?

test/org/javarosa/model/xform/SmsSerializingVisitorTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        sms = new String(payload.getPayloadBytes(), "UTF-8").replace("\\", "").replace("\\\\", "\\");
+    }
+
+    @Test
+    public void SmsNotNull() {
+        assertNotNull(sms);
+    }
+
+    @Test
```
dcbriccetti: Would it please you to know you could shrink this file from 145 to 82 lines, and remove most of the complexity? If you’re like me, you’ll say “Yes, yes, yes!”
Help yourself to this if you like it:
https://github.com/dcbriccetti/javarosa/commit/d52fe396f4b0e622ceb8b29d9fac5618eb4735c0

test/org/javarosa/model/xform/SmsSerializingVisitorTest.java
```diff
@@ -0,0 +1,145 @@
...
+     * groups.
+     */
+    @Test
+    public void testTaggedAnswers() {
+        for (String tag : getAnswerTags()) {
+            assertTrue(taggedAnswerExists(tag, sms));
+        }
+    }
+
+    /***
```
dcbriccetti: May I suggest some language improvements?
```
    /** Ensures that the answer for maiden_name, which doesn’t have the “tag” attribute, is not present */
    @Test
    public void ensureAnswerInNonTaggedElementNotPresent() {
        assertFalse(sms.contains("Placeholder"));
    }

    /** Ensures that “CTY”, which does have the “tag” attribute, but has no answer, is not present */
    @Test
    public void ensureTagWithNoAnswerNotPresent() {
        assertFalse(sms.contains("CTY"));
    }
```

jd-alexander: Sure!

test/org/javarosa/model/xform/SmsSerializingVisitorTest.java
jd-alexander: Nice!

src/org/javarosa/model/xform/SMSSerializingVisitor.java
jd-alexander: I pulled in those changes from the Medic branch https://github.com/medic/javarosa/commit/90c46a6a3fc19b1b75783cb27cfa907d37099cca#diff-700d0f975ca498a121d65ff76e95a8c7
So it's probably can be reversed once it's not something that has to be done.

dcbriccetti: I should probably read more of the background to see how we got to where we are.

lognaturel:
This is looking nice!

Update: we have the ok to officially introduce `prefix`, `delimiter` and `tag` in the `http://www.opendatakit.org/xforms` namespace: https://github.com/opendatakit/xforms-spec/pull/195#issuecomment-398919147. I think all this will take is to specify a namespace as the first parameter to `getAttributeValue`. For example, `delimiter = root.getAttributeValue("", "delimiter");` will become `delimiter = root.getAttributeValue("NAMESPACE_ODK", "delimiter");`. That constant can be introduced in `XFormParser` along with the existing `NAMESPACE_JAVAROSA` one. The test forms and forms used for verification on the Collect side will need to be updated.

I'm getting some second thoughts about how these changes should come in. Basically, this new version of `SMSSerializingVisitor` is not compatible with the prior one. Forms that used to be serializable are no longer. We don't know for sure that anyone is a client of the class but I think this is big enough that we should consider marking `SMSSerializingVisitor` as deprecated and introducing a new class. If nothing else, this gets us in the habit of doing things right. A possible name for the new class could be `CompactSerializingVisitor` -- I realized when writing up https://github.com/opendatakit/xforms-spec/pull/195#issuecomment-398919147 that really, there's nothing about this spec that makes it SMS-specific. I'm really not sure about the best way to proceed here. Would be good to get your thoughts @dcbriccetti @jd-alexander @yanokwa.

yanokwa:
Good catch on breaking downstream users of JavaRosa! Creating a `CompactSerializingVisitor` and deprecating `SMSSerializingVisitor` seems like a great solution.

lognaturel:
Ok, that does seem like the more correct way. I would address this by adding a commit that moves `SMSSerializingVisitor` as it exists in this branch to `CompactSerializingVisitor`. The commit description would state why the file was moved and that history can be found by looking at `SMSSerializingVisitor`. Then I would add another commit that brings back `SMSSerializingVisitor` as it currently exists in `master` but with a `@Deprecated` annotation. That way it is possible to retrace the steps from this PR if need be. If you see a better approach, @jd-alexander, go for it. I just want to make sure we don't lose your useful commit messages.

jd-alexander:
Noted @lognaturel will do!

src/org/javarosa/core/services/transport/payload/ByteArrayPayload.java
jd-alexander: Yes, I could do that. Thanks for the idea.

lognaturel: This method can now be removed, right? It wasn't part of the original API and isn't used anymore.

src/org/javarosa/model/xform/SMSSerializingVisitor.java
```diff
@@ -131,85 +109,87 @@ public IDataPayload createSerializedPayload(FormInstance model, IDataReference r
...
-            delimiter = root.getAttributeValue("", "delimeter");
+            delimiter = root.getAttributeValue(NAMESPACE_ODK, "delimeter");
         }
-        prefix = root.getAttributeValue("", "prefix");
+        String prefix = root.getAttributeValue(NAMESPACE_ODK, "prefix");
 
-        xmlns = (xmlns != null)? xmlns : " ";
-        delimiter = (delimiter != null ) ? delimiter : " ";
+        delimiter = (delimiter != null) ? delimiter : " ";
         prefix = (prefix != null) ? prefix : " ";
```
lognaturel: Feels to me like the default prefix should be an empty string, not a space. Since we're making broader changes and introducing this in a new class I think we should make improvements of that kind.

src/org/javarosa/model/xform/SMSSerializingVisitor.java
```diff
@@ -131,85 +109,87 @@ public IDataPayload createSerializedPayload(FormInstance model, IDataReference r
...
+    }
+
+    private void serializeTree(TreeElement root, StringBuilder sb) {
         for (int j = 0; j < root.getNumChildren(); j++) {
-            TreeElement tee = root.getChildAt(j);
-            String e = serializeNode(tee);
-            if(e != null) {
-                theSmsStr += e;
+            TreeElement treeElement = root.getChildAt(j);
+            if (treeElement.isLeaf() && treeElement.getAttribute("", "tag") != null) {
```
lognaturel: This should also be in the ODK namespace.

src/org/javarosa/model/xform/SMSSerializingVisitor.java
```diff
@@ -131,85 +109,87 @@ public IDataPayload createSerializedPayload(FormInstance model, IDataReference r
...
-                        + serializedAnswer);
+                    + instanceNode.getValue().toString() + ", "
+                    + serializedAnswer);
             } else if (serializedAnswer instanceof String) {
-                Element e = new Element();
-                e.addChild(Node.TEXT, (String) serializedAnswer);
+                Element element = new Element();
+                element.addChild(Node.TEXT, serializedAnswer);
 
                 String tag = instanceNode.getAttributeValue("", "tag");
```
lognaturel: ODK namespace

src/org/javarosa/model/xform/CompactSerializingVisitor.java
```diff
@@ -0,0 +1,210 @@
...
+
+                stringBuilder.append(tag);
+
+                stringBuilder.append(delimiter);
+
+                for (int k = 0; k < element.getChildCount(); k++) {
+                    stringBuilder.append(element.getChild(k).toString().
+                        replace("\\", "\\\\")
+                        .replace(delimiter, "\\" + delimiter)
+                        .replace(tag, "\\" + tag));
```
lognaturel: I'm not sure about escaping tags. I expect most recipients will tokenize incoming messages using the delimiter and consider values in even positions to be tags and values in odd positions to be corresponding answer values. This kind of escaping makes processing much more complex.

Take this simple example where a form designer is using alphabetic designations as tags (a, b, c):
```
<fname odk:tag="a">Alabama</fname>
```
The serialized version given a delimiter of " " would be `a Al\ab\am\a`. 

I believe this is something @ekigamba asked for. Could you please explain in more depth why you would like it? Can we launch an initial version without it and reconsider if needed?

src/org/javarosa/core/services/transport/payload/ByteArrayPayload.java
dcbriccetti: It should be removed

src/org/javarosa/model/xform/CompactSerializingVisitor.java
jd-alexander: Oh yes, that's a really good point and example. I don't really have any feelings towards it. So I think we can remove it for now.

lognaturel: Yes, let's. Once it's in it's really hard to take away. If there's user demand for it we can draft up a rationale and consider implications.

src/org/javarosa/core/services/transport/payload/ByteArrayPayload.java
```diff
@@ -126,4 +124,15 @@ public int getTransportId() {
         //TODO: Most messages can include this data
         return -1;
     }
+
+    @Override
+    public String toString() {
+        try {
+            return new String(payload, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
```
dcbriccetti: Perhaps we should write a message via the logger.

jd-alexander: Yes, was thinking of that. Thanks @dcbriccetti Was looking at the `Logger` class but realized it's depreciated. Which logger should I use? Looking in other parts of the code for dets

src/org/javarosa/core/services/transport/payload/ByteArrayPayload.java
```diff
@@ -72,68 +74,69 @@ public ByteArrayPayload(byte[] payload) {
...
+
+    @Override
+    public String toString() {
+        try {
+            return new String(payload, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            logger.error("The encoding being utilized isn't supported.", e);
+        }
+
+        return null;
```
dcbriccetti: Thanks for the log message.

I’m thinking that a toString method should never return null. Maybe an empty String in this case?

lognaturel: Yes, very good catch.

src/org/javarosa/model/xform/CompactSerializingVisitor.java
```diff
@@ -0,0 +1,210 @@
...
+        String prefix = root.getAttributeValue(NAMESPACE_ODK, "prefix");
+
+        delimiter = (delimiter != null) ? delimiter : " ";
+
+        if (prefix != null) {
+            resultText = prefix.concat(delimiter);
+        }
+
+        // serialize each node (and its children) to get its answers
+        resultText = serializeTreeToString(root);
```
lognaturel: This loses the prefix. Should be `resultText +=`.

lognaturel: Would be good to have a test case for this.

src/org/javarosa/model/xform/CompactSerializingVisitor.java
```diff
@@ -0,0 +1,209 @@
...
+     * org.javarosa.core.model.utils.ITreeVisitor#visit(org.javarosa.core.model
+     * .DataModelTree)
+     */
+    public void visit(FormInstance tree) {
+
+        TreeElement root = tree.getRoot();
+
+        delimiter = root.getAttributeValue(NAMESPACE_ODK, "delimiter");
+        if (delimiter == null) {
+            // for the spelling-impaired...
```
yanokwa: If this was for a US/UK spelling difference or is for legacy reasons, then fine, but I think we shouldn't do this sort of thing. I mean where do you draw the line?

jd-alexander: Yeah we can take it out. It has been in there since the original SMSSerializingVisitor`.

Issue #277, opened 2018-03-15
Add distance(nodeset<geopoint> | geoshape | geotrace) function


lognaturel:
https://opendatakit.github.io/xforms-spec/#fn:distance

https://github.com/opendatakit/xforms-spec/issues/129

dcbriccetti:
Just to clarify,

> Returns the distance in meters of either a nodeset of geopoints or a single geoshape value or a single geotrace value (not a combination of these) on Earth, in the sequence provided by the points in the parameter. It takes into account the circumference of the Earth around the Equator and does not take altitude into account.

This distance is the sum of the great circle distances (arcs over the Earth’s surface) between each point and the previous, starting with the 2nd point. Yes?

dcbriccetti:
I’m coordinating with Martijn on a common implementation and shared tests, and after reading some background from him, think we should leave the area calculation alone, despite its imprecision.

dcbriccetti:
I’ve discovered that the area calculator web tool used to generate test data for our area calculation doesn’t close the polygon when producing its perimeter. The 2,076 meter value is the sum of the lengths of the four sides between the five points moving anti-clockwise from the top-right point.

So, in our tests using this data, we also won’t close the polygons for perimeter calculations.

![area calculator](https://user-images.githubusercontent.com/55099/39219776-78f242c6-47e1-11e8-8927-7200adf546c0.png)

lognaturel:
That sounds right. The calculation should use the points defined by the expression passed in and not add any. In the case of a geoshape argument, the first and last point will be the same.

dcbriccetti:
Good to know. Thanks!

dcbriccetti:
Getting close to done here.

![screen shot 2018-04-24 at 9 41 46 pm](https://user-images.githubusercontent.com/55099/39226616-5d9743c4-4808-11e8-9347-6c504bd5911e.png)

dcbriccetti:
OK, reviewers, at your convenience. 3 PRs. 302–304. I’m thinking 15–25 minutes would be needed. And if you like, I’ll even walk you through it if we can coordinate our schedules.

dcbriccetti:
I need to complete so that it satisfies:

>Returns the distance in meters of either a nodeset of geopoints or a single geoshape value or a single geotrace value

PR #303, opened 2018-04-25
Add distance calculation to GeoUtils (Issue #277 part 2)

dcbriccetti:
Partially implements #277 

Automated tests

The Spherical Law of Cosines approach treats the Earth as a sphere with a given radius. It provides results consistent with our area calculations.

No. It’s a new feature not yet connected to the API.


dcbriccetti:
I just rebased on master. @lognaturel, thanks for merging 302.

lognaturel:
Thank you, @dcbriccetti!

PR #305, opened 2018-04-25
Fix comments in core.model.Constants

dcbriccetti:
Fix comments in core.model.Constants, especially doc comments in the wrong place (they’re really useful in an IDE when they are correct).

screen shot 2018-04-25 at 2 56 46 pm

I ran all tests. There are only changes to comments.

No.

### DIFF
--- a/src/org/javarosa/core/model/Constants.java
+++ b/src/org/javarosa/core/model/Constants.java
@@ -16,18 +16,11 @@
 
 package org.javarosa.core.model;
 
-
-/**
- * Constants shared throught classes in the containing package.
- *
- * @version ,
- */
+/** Constants shared throughout classes in the containing package. */
 public class Constants {
 
-    /** Empty strig representation */
     public static final String EMPTY_STRING = "";
 
-    /** Index for no selection */
     public static final int NO_SELECTION = -1;
 
     /** ID not set to a value */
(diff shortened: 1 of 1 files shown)


src/org/javarosa/core/model/Constants.java
```diff
@@ -42,29 +35,45 @@
...
-    public static final int DATATYPE_CHOICE = 7;    /** This is a question with alist of options where not more than one option can be selected at a time. */
-    public static final int DATATYPE_CHOICE_LIST = 8;    /** This is a question with alist of options where more than one option can be selected at a time. */
-    public static final int DATATYPE_BOOLEAN = 9;    /** Question with true and false answers. */
-    public static final int DATATYPE_GEOPOINT = 10; /** Question with location answer. */
-    public static final int DATATYPE_BARCODE = 11; /** Question with barcode string answer. */
-    public static final int DATATYPE_BINARY = 12; /** Question with external binary answer. */
-    public static final int DATATYPE_LONG = 13; /** Question with external binary answer. */
-    public static final int DATATYPE_GEOSHAPE = 14; /** Question with GeoShape answer. */
-    public static final int DATATYPE_GEOTRACE = 15; /** Question with GeoTrace answer. */
+    /* For nodes that have no data, or data type otherwise unknown */
```
lognaturel: Does this need to have two * at the beginning?

src/org/javarosa/core/model/Constants.java
```diff
@@ -81,11 +90,9 @@
     public static final int CONTROL_AUDIO_CAPTURE = 12;
     public static final int CONTROL_VIDEO_CAPTURE = 13;
     public static final int CONTROL_OSM_CAPTURE = 14;
-    public static final int CONTROL_FILE_CAPTURE = 15; // generic upload
+    /** generic upload */
+    public static final int CONTROL_FILE_CAPTURE = 15;
 
-    /** constants for xform tags */
+    /* constants for xform tags */
```
lognaturel: Same question here.

dcbriccetti: No, this is sort of a description of a group that follows, and has no way to express it in a doc comment.

src/org/javarosa/core/model/Constants.java
dcbriccetti: Yes. Thanks!

src/org/javarosa/core/model/Constants.java
lognaturel: Ahh, makes sense.

ggalmazor: Sorry for stepping in, but we're trying to improve the support for automatic reformatting on Briefcase and looking into extending that to other projects as well.

How do these block comments react to automatic reformatting? 

If they're just one-line comments, wouldn't it be better to use one-line comment syntax `// blablabla`, which won't be relocated/reformatted by the automatic reformatting tool?

dcbriccetti: Hi @ggalmazor. They are doc comments (JavaDoc). They allow IDEA users to press F1 (depending on key bindings) anywhere the fields are used to get the description (see the picture). This change was simply to enable this. 

I just changed the verb in the title in the description from “fix” to “improve”, to be consistent with this being an incremental change.

I agree with you about better names, and valueless redundant comments. Going further, most of these should be changed to Enums.

Feel free to continue the work. I was just solving something that was bothering me at the moment (while looking at something in Collect).

ggalmazor: Thanks, @dcbriccetti. I just wanted to discuss it a little bit to see if this could get traction.

PR #306, opened 2018-04-30
Add logging to XPathEqExpr and XPathPathExpr

dcbriccetti:
Add logging to `XPathEqExpr` and `XPathPathExpr` to help show their behavior. Do some safe cleanup of bloated code in `XPathEqExpr`.

See https://github.com/opendatakit/collect/issues/2041

Tests pass
N/A
No

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathEqExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathEqExpr.java
@@ -25,71 +25,77 @@ import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.util.externalizable.DeserializationException;
 import org.javarosa.core.util.externalizable.ExtUtil;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.lang.Math.abs;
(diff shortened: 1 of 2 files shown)


src/org/javarosa/xpath/expr/XPathEqExpr.java
```diff
@@ -25,71 +25,77 @@
...
+import static org.javarosa.xpath.expr.XPathFuncExpr.unpack;
 
 public class XPathEqExpr extends XPathBinaryOpExpr {
+    private static final Logger logger = LoggerFactory.getLogger(XPathEqExpr.class.getSimpleName());
     public boolean equal;
 
-    public XPathEqExpr () { } //for deserialization
+    /** Deserialization constructor */
+    public XPathEqExpr() {
+        logger.debug("XPathEqExpr{}()", id(), equal, a, b);
```
lognaturel: Does this only log the first of the substitution params logged?

dcbriccetti: Thanks for catching that. I’ll don’t think that’s what I intended.

dcbriccetti: That’s fixed, I ran tests and saw that the output from that was unchanged, rebased on master, and force pushed.

PR #307, opened 2018-05-08
Improve XPathPathExpr code quality

dcbriccetti:
Move eval method from XPathPathExpr to new class, XPathPathExprEval, and modernize its code and make it much more understandable.

(cherry picked from commit 2d5998d)

I saw that all tests still pass, and I stepped through the code in the debugger.

The goal was to separate the `eval` concern from the rest of XPathPathExpr, and to modernize code using automatic IDEA refactorings.

The risk seems low. I used refactorings like extract method, which are safe.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathPathExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathPathExpr.java
@@ -21,6 +21,7 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import org.javarosa.core.model.condition.EvaluationContext;
@@ -49,8 +50,6 @@ import org.javarosa.core.util.externalizable.ExtUtil;
 import org.javarosa.core.util.externalizable.ExtWrapList;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.xform.util.XFormAnswerDataSerializer;
-import org.javarosa.xpath.XPathException;
-import org.javarosa.xpath.XPathMissingInstanceException;
 import org.javarosa.xpath.XPathNodeset;
 import org.javarosa.xpath.XPathTypeMismatchException;
 import org.javarosa.xpath.XPathUnsupportedException;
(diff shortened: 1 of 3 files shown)


src/org/javarosa/xpath/expr/XPathPathExpr.java
```diff
@@ -70,209 +69,134 @@
...
-        //to fix conditions based on non-relevant data, filter the nodeset by relevancy
-        for (int i = 0; i < nodesetRefs.size(); i++) {
-            if (!m.resolveReference(nodesetRefs.get(i)).isRelevant()) {
-                nodesetRefs.remove(i);
-                i--;
-            }
-        }
-
-        return new XPathNodeset(nodesetRefs, m, ec);
+    public XPathNodeset eval(final DataInstance unusedDataInstance, final EvaluationContext ec) {
```
ggalmazor: I would argue to remove these `final`s here. Maybe the one for `TreeReference ref` in `getReference()` is useful while it gets smaller but here it's just one line of code and those variables are just passed downstream.

dcbriccetti: Thanks. I intended to remove them, and will now. They were part of the evolution of the code (originally the first parameter was never read, but reassigned!).

src/org/javarosa/xpath/expr/XPathPathExpr.java
```diff
@@ -355,16 +285,16 @@ public String toString () {
...
+            XPathPathExpr x = (XPathPathExpr) o;
 
             //Shortcuts for easily comparable values
-            if(init_context != x.init_context || steps.length != x.steps.length) {
+            if (init_context != x.init_context || steps.length != x.steps.length) {
                 return false;
             }
 
-            return ExtUtil.arrayEquals(steps, x.steps) && (init_context == INIT_CONTEXT_EXPR ? filtExpr.equals(x.filtExpr) : true);
+            return ExtUtil.arrayEquals(steps, x.steps) && (init_context != INIT_CONTEXT_EXPR || filtExpr.equals(x.filtExpr));
```
ggalmazor: I might have to wait for tomorrow to be fresh again to really be sure that this does the same, but I'm guessing you've just let IntelliJ simplify the boolean expression :+1:

dcbriccetti: Yes.

src/org/javarosa/xpath/expr/XPathPathExpr.java
```diff
@@ -70,209 +69,134 @@
...
+            case XPathPathExpr.INIT_CONTEXT_EXPR:
+                if (filtExpr.x instanceof XPathFuncExpr) {
+                    XPathFuncExpr func = (XPathFuncExpr) (this.filtExpr.x);
+                    switch (func.id.toString()) {
+                        case "instance":
+                            ref.setRefLevel(TreeReference.REF_ABSOLUTE); //i assume when refering the non main instance you have to be absolute
+
+                            parentsAllowed = false;
+                            if (func.args.length != 1) {
+                                throw new XPathUnsupportedException("instance() function used with " + func.args.length + " arguments. Expecting 1 argument");
```
ggalmazor: "arguement" sounds much more interesting. Spelling is overrated :D

src/org/javarosa/xpath/expr/XPathPathExpr.java
ggalmazor: Wow, it's great that you fixed that one :)

dcbriccetti: At least I made it obvious that all those callers are passing in something that is ignored.

PR #310, opened 2018-05-11
Fix randomize test form and add extra tests

ggalmazor:
This fixes some defects that got merged in #296 by accident:
- Fixed the randomize function params in the test form
- Added specific tests that make sure calculated fields that use randomize get some expected value

Run the tests & checked coverage report to ensure tests exercise the code.

Nope.

Nope.


lognaturel:
Thank you! Evidently it slipped through my net as well. I just happened to come back to look at it while working on https://github.com/XLSForm/pyxform/pull/199 and it caught my eye.

PR #311, opened 2018-05-13
Add DataType enum

dcbriccetti:
I assume JavaRosa was created before enums were added to the language in Java 5. This change adds an enum for DataType. Gradually we can replace uses of the “int enum anti-pattern” with the enum. Immediately we can use it in [Briefcase](https://github.com/opendatakit/briefcase/pull/457).

I threw in a tiny [unrelated change](https://github.com/opendatakit/javarosa/commit/47c950d322f41f390a2b825108f33851c583b229) that will speed up the automated tests.

I saw through tests and code coverage that the code is exercised and works.

No

### DIFF
--- a/src/org/javarosa/core/model/DataType.java
+++ b/src/org/javarosa/core/model/DataType.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2018 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


src/org/javarosa/core/model/DataType.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.javarosa.core.model.Constants.DATATYPE_DECIMAL;
+import static org.javarosa.core.model.Constants.DATATYPE_GEOPOINT;
+import static org.javarosa.core.model.Constants.DATATYPE_GEOSHAPE;
+import static org.javarosa.core.model.Constants.DATATYPE_GEOTRACE;
+import static org.javarosa.core.model.Constants.DATATYPE_INTEGER;
+import static org.javarosa.core.model.Constants.DATATYPE_LONG;
+import static org.javarosa.core.model.Constants.DATATYPE_NULL;
+import static org.javarosa.core.model.Constants.DATATYPE_TEXT;
+import static org.javarosa.core.model.Constants.DATATYPE_TIME;
+import static org.javarosa.core.model.Constants.DATATYPE_UNSUPPORTED;
```
dcbriccetti: I’d like to use import * here, but don’t know how to suppress the warning.

PR #313, opened 2018-05-15
Improve the DataType enum and the use of data type classes

dcbriccetti:
Ran all automated tests.

This simplifies code and moves us closer to eliminating the “int enum antipattern” of DATATYPE.* in Constants.

Not that I can see.


lognaturel:
This looks good to me and with a nice commit comment to remind us what's going on. Would you like a sanity check from @grzesiek2010 or @ggalmazor or do you think we can merge?

PR #316, opened 2018-05-17
Issue 314 revert use of java8 datetime api

### LINKED ISSUE #314: Collect runtime error when trying to access ZoneId, LocalDateTime introduced in Android 26
From user report at https://forum.opendatakit.org/t/odk-collect-v1-15-beta/13086

@ggalmazor can you please take a look and see if you have some good alternative ideas? We can always go back to joda.

It's good that we're getting QA involved in JR pull requests and we should make sure all of them go through the process! IntelliJ also does mark recent Java additions by default so when it does it would be good to check to see whether they're available in Android. I'm also going to add a note in the README about needing to check for Android support. It's so easy to forget!

ggalmazor:
Closes #314

- The changes in the implementation have been made **before** the changes in the tests, thus being safe.
- All unit tests are passing
- Searching for imports of `java.time` namespace doesn't show any results

I've just translated to Joda Time every date/time related operation made with Java8's Date/Time API.

Nope.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,12 +43,18 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
   <dependencies>
+    <!-- Be sure to update dependencies in build.gradle to match -->
     <dependency>
-      <!-- Be sure to update dependencies in build.gradle to match -->
       <groupId>net.sf.kxml</groupId>
       <artifactId>kxml2</artifactId>
       <version>2.3.0</version>
       <scope>compile</scope>
+    </dependency>
+    <dependency>
+       <groupId>joda-time</groupId>
+       <artifactId>joda-time</artifactId>
+       <version>2.9.9</version>
+       <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/test/org/javarosa/core/model/Safe2014DagImplTest.java
+++ b/test/org/javarosa/core/model/Safe2014DagImplTest.java
@@ -1,12 +1,12 @@
 package org.javarosa.core.model;
 
-import java.time.LocalTime;
 import org.javarosa.core.model.instance.FormInstance;
 import org.javarosa.core.model.instance.InstanceInitializationFactory;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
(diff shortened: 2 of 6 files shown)


test/org/javarosa/core/model/Safe2014DagImplTest.java
```diff
@@ -381,7 +381,7 @@ public void deleteRepeatGroupWithCalculationsTimingTest() throws Exception {
         }
 
         // Then
-        LocalTime duration = LocalTime.ofNanoOfDay(System.nanoTime() - start);
+        LocalTime duration = LocalTime.fromMillisOfDay(System.nanoTime() - start);
```
dcbriccetti: Mixed millis and nanos?

yanokwa:
If a form is needed to test this issue, the first question in this form crashes Collect...
[biotikaForm_2018_BfNPI_HUN.zip](https://github.com/opendatakit/javarosa/files/2014878/biotikaForm_2018_BfNPI_HUN.zip)

yanokwa:
The "sample" form on opendatakit.appspot.com also has the same problem. 
You can see a video here: https://drive.google.com/file/d/1vLyskjHI9RsseI0WzLSKXqtY1b1WbeGW/view

test/org/javarosa/core/model/Safe2014DagImplTest.java
ggalmazor: Fixed!

yanokwa:
@ggalmazor One reminder on your JavaRosa PRs is to check both builds pass! We run both gradle and mvn. The latter is what we use to generate the builds for release and you forgot to revert the dependencies there.

pom.xml
```diff
@@ -43,13 +43,19 @@
...
   <dependencies>
+    <!-- Be sure to update dependencies in build.gradle to match -->
     <dependency>
-      <!-- Be sure to update dependencies in build.gradle to match -->
       <groupId>net.sf.kxml</groupId>
       <artifactId>kxml2</artifactId>
       <version>2.3.0</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
```
ggalmazor: Thx, @yanokwa. I totally forgot about this :/

kkrawczyk123:
### Tested with success!

Verified on Androids: 4.1, 4.2, 4.4, 6.0, 7.0 and 8.1 

PR #319, opened 2018-06-04
Distance function doesn't work if shape/trace has only two points

### LINKED ISSUE #318: Distance function doesn't work if shape/trace has only two points
#### Software versions 
JavaRosa v2.10.0

#### Problem description
Distance function doesn't work if shape/trace has only two points

@kkrawczyk123 was right https://github.com/opendatakit/javarosa/pull/308#issuecomment-389180325 and I can confirm there is a bug

grzesiek2010:
Closes #318 

I confirmed that it works well using the existed `GeoShapeDistanceTest` test.

The bug was here https://github.com/opendatakit/javarosa/pull/319/files#diff-aac5e2989d5cd1037caa28a21dadfbeaR29

it required at least three points even for `distance` function.

No.


lognaturel:
Good catch, @grzesiek2010!

FYI, @dcbriccetti -- mystery solved.

dcbriccetti:
Thanks for solving that, @grzesiek2010. I’d like to tweak your change.

lognaturel:
Now that I think more about it, is that conditional even necessary? Seems it can be up to the functions and/or the clients to verify the number of points.

dcbriccetti:
Agreed. I’m working on error messages for <3 for area and <2 for distance, in the right place.

dcbriccetti:
The other change is undoing the altering of the test data to exercise the 2-points distance case.

lognaturel:
> I’m working on error messages for <3 for area and <2 for distance, in the right place.

`XPathFuncExpr`, right? We can check the length of `latLongs` for each of the functions.

> undoing the altering of the test data to exercise the 2-points distance case.

Do you mean adding back the 3-points case?

There's no rush on this so as long as we're clear on what the changes should be @grzesiek2010 can make them when he has a chance.

dcbriccetti:
Here’s where I’m thinking:
```
        } else if (name.equals("enclosed-area") || name.equals("area")) {
            assertArgsCount(name, args, 1);
            List<GeoUtils.LatLong> latLongs = new XPathFuncExprGeo().getGpsCoordinatesFromNodeset(name, argVals[0]);
            if (latLongs.size() < 3) {
                throw new XPathUnhandledException(String.format("function '%s' requires at least three points.", name));
            }
            return GeoUtils.calculateAreaOfGPSPolygonOnEarthInSquareMeters(latLongs);
        } else if (name.equals("distance")) {
            assertArgsCount(name, args, 1);
            List<GeoUtils.LatLong> latLongs = new XPathFuncExprGeo().getGpsCoordinatesFromNodeset(name, argVals[0]);
            if (latLongs.size() < 2) {
                throw new XPathUnhandledException(String.format("function '%s' requires at least two points.", name));
            }
            return GeoUtils.calculateDistance(latLongs);
```

dcbriccetti:
Yes, the three points test provides the same three points through the three types of input. We can change them all to only two points, or add another test for two points.

dcbriccetti:
@grzesiek2010, I turn it back to you. Here’s my thinking:

https://github.com/dcbriccetti/javarosa/commit/2eda7086f32718a01704090ab8bc4b84f85a7c2c

It would be good to add tests to see that the error messages get issued when not enough points are supplied for area and distance.

Thanks, again.

dcbriccetti:
Shall I finish this, @grzesiek2010, or would you like to?

grzesiek2010:
@dcbriccetti sorry for the delay I've just updated the pr so please review.

grzesiek2010:
I also noticed something strange using that form from tests on a real device.
Saved instance file contains:
```
   <shape-result>1.013001489695198E7</shape-result>
   <trace-result>1.013001489695198E7</trace-result>
   <point-result>2.026002979390396E7</point-result>
```

the last one looks wrong am I missing something? I made a quick investigation and here https://github.com/opendatakit/javarosa/pull/319/files#diff-42f4fbe32c7dcb45477406e32de2803fR427

it's always 1.013001489695198E7.

lognaturel:
Does seem weird. I think we can treat it as a separate bug and get this merged in.

test/org/javarosa/core/util/GeoDistanceTest.java
```diff
@@ -0,0 +1,81 @@
...
+                {"point"},
+            });
+        }
+
+        /**
+         * Shows that the distance function can be used with inputs
+         * <a href="https://opendatakit.github.io/xforms-spec/#fn:distance">as specified</a>.
+         */
+        @Test
+        public void parsingAndDistanceIsCorrectForThreeGeoTypes() throws Exception {
```
dcbriccetti: This method has a history. In general, I think we shouldn’t copy and paste code from existing modules into new ones if instead we could rename the file and preserve the history.

test/org/javarosa/core/util/GeoShapeAreaTest.java
```diff
@@ -36,20 +36,31 @@
...
+        assertEquals(1_801, (Double) distance.getValue(), 0.5);
+    }
+
+    @Test
+    public void testAreaWithLessThanThreePoints() throws Exception {
+        FormDef formDef = parse(r("area_with_less_than_three_points.xml")).formDef;
+        try {
+            formDef.initialize(true, new InstanceInitializationFactory());
+        } catch (Exception e) {
+            assertEquals("Error evaluating field 'arearesult': The problem was located in calculate expression for /area/arearesult\n" +
```
dcbriccetti: This is fragile. If we change the exception text even slightly this will break.

test/org/javarosa/core/util/GeoShapeDistanceTest.java
```diff
@@ -1,62 +0,0 @@
...
-
-import static org.javarosa.core.model.instance.TreeReference.DEFAULT_MULTIPLICITY;
-import static org.javarosa.core.util.GeoUtils.EARTH_EQUATORIAL_CIRCUMFERENCE_METERS;
-import static org.javarosa.test.utils.ResourcePathHelper.r;
-import static org.javarosa.xform.parse.FormParserHelper.parse;
-import static org.junit.Assert.assertEquals;
-
-@RunWith(Parameterized.class)
-public class GeoShapeDistanceTest {
-    @Parameterized.Parameter(value = 0)
```
dcbriccetti: We should have preserved the history here.

test/org/javarosa/core/util/GeoDistanceTest.java
grzesiek2010: I renamed the file using `Refactor -> Rename` option but after adding changes in the test it's treated as a new file.

test/org/javarosa/core/util/GeoShapeDistanceTest.java
grzesiek2010: The same as above I used `rename` option.

test/org/javarosa/core/util/GeoShapeAreaTest.java
grzesiek2010: Yes, it is a little bit... any idea how to check that in a better way?

grzesiek2010: Fixed. Thanks.

test/org/javarosa/core/util/GeoShapeAreaTest.java
```diff
@@ -36,20 +41,34 @@
...
+        try {
+            formDef.initialize(true, new InstanceInitializationFactory());
+        } catch (RuntimeException e) {
+            if (e.getCause() instanceof XPathUnhandledException) {
+                assertThat(e.getCause().getMessage(), containsString(REQUIRES_THREE_POINTS_MESSAGE));
+                return;
+            }
+        }
+        fail("The expected exception with error about three points being required was not thrown");
+    }
```
dcbriccetti: screen shot 2018-06-27 at 6 42 42 pm

Two things:
1) I know your IDE alerts you to things that should be fixed. Does it show you that you created a duplicated block of non-trivial code?
2) I wrote and tested this nice code for you, yet git says you wrote it. Did you cherry pick it?

grzesiek2010: 1. I can't see such a prompt.
2. Fixed.

test/org/javarosa/core/util/GeoDistanceTest.java
grzesiek2010: No I just renamed the class and after that, it was ok (treated as the same class) but when I implemented changes in the code it's treated as a new class.

test/org/javarosa/core/util/GeoShapeAreaTest.java
dcbriccetti: For 1, I’ll be glad to help you remove the duplication if you like. Thanks for 2.

dcbriccetti: Here’s what I’m thinking, to remove the duplication:

https://github.com/dcbriccetti/javarosa/commit/0d7f526b57ea5276ffd3ef7c7c06f05ce9353ad6

grzesiek2010: Ok, I think it's ready.

PR #321, opened 2018-06-08
Fixed constraint message

### LINKED ISSUE #320: Constraint message doesn't work if it's only one word and without itext function
#### Software versions 
JavaRosa v2.10.0

#### Problem description
Constraint message doesn't work if it's only one word and without `itext` function

[constraint_without_itext.xml.txt](https://github.com/opendatakit/javarosa/files/2084061/constraint_without_itext.xml.txt)
[constraint_with_itext.xml.txt](https://github.com/opendatakit/javarosa/files/2084062/constraint_with_itext.xml.txt)

1. Open the first attached form constraint_without_itext.xm
2. Navigate to the second question.
3. Check displayed constraint message - a default message is displayed

#### Expected behavior
Custom constraint_message should be displayed.

grzesiek2010:
Closes #320 

I've tested attached forms.

It doesn't make sense to use `XPathParseTool.parseXPath` if a constraint doesn't use `itext`.

No.


src/org/javarosa/core/model/condition/Constraint.java
```diff
@@ -70,7 +73,7 @@ public String getConstraintMessage(EvaluationContext ec, FormInstance instance,
     private void attemptConstraintCompile() {
         xPathConstraintMsg = null;
         try {
-            if(constraintMsg != null) {
+            if (constraintMsg != null && constraintMsg.startsWith(ITEXT_OPEN) && constraintMsg.endsWith(ITEXT_CLOSE)) {
```
dcbriccetti: As I see from running in the debugger (before this change), 
- parsing a two-word string, “Your message”, results in an exception, `org.javarosa.xpath.parser.XPathSyntaxException: 
Couldn't understand the expression starting at this point: string(Your͎message)`. The exception is ignored, and `xPathConstraintMsg` remains `null`.

- parsing a one-word string, “Message”, is successful, and results in xPathConstraintMsg being set to `{func-expr:string,{{path-expr:rel,{{step:child,Message}}}}}`

For both of these strings, xPathConstraintMsg should remain `null`. The defect in the code is that plain strings (not inside itext) are being parsed unnecessarily, with the result that in at least this case of a single word, xPathConstraintMsg is set to something incorrect, when it should be null. Your correction is to avoid the parsing when itext is not used by searching the message for itext tokens.

If all that is correct, then let me ask:
- Is it necessary to look for ITEXT_CLOSE at the end?
- Might constraintMsg have leading or trailing whitespace or any other characters that might prevent the check for ITEXT_OPEN from working?
- For more robustness, did you consider parsing in all cases, and only using the result if an examination of the parse result shows that itext was used, such as is the case here?
```
{func-expr:string,{{func-expr:jr:itext,{{str:'/data/q1:constraintMsg'}}}}}
```

grzesiek2010: > For both of these strings, xPathConstraintMsg should remain null. The defect in the code is that plain strings (not inside itext) are being parsed unnecessarily, with the result that in at least this case of a single word, xPathConstraintMsg is set to something incorrect, when it should be null. Your correction is to avoid the parsing when itext is not used by searching the message for itext tokens.

correct.

> Is it necessary to look for ITEXT_CLOSE at the end?

probably it's not necessary but it's not a problem as well.

>Might constraintMsg have leading or trailing whitespace or any other characters that might prevent the check for ITEXT_OPEN from working?

maybe but why would it be better? I think we should just avoid parsing.

> For more robustness, did you consider parsing in all cases, and only using the result if an examination of the parse result shows that itext was used, such as is the case here?

I think I can't check that easily.

dcbriccetti: >>Might constraintMsg have leading or trailing whitespace or any other characters that might prevent the check for ITEXT_OPEN from working?

>maybe but why would it be better? I think we should just avoid parsing.

No, I’m saying what if the form has whitespace before the ITEXT_OPEN?

grzesiek2010: Ahh I see you are right I just added `trim()` function to avoid such problems.

dcbriccetti: OK, I’ll look for the trim when you push. Also, should we have a test for this?

grzesiek2010: I think it's ready. Please review.

src/org/javarosa/core/model/condition/Constraint.java
```diff
@@ -81,6 +84,9 @@ private void attemptConstraintCompile() {
     public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
         constraint = (IConditionExpr)ExtUtil.read(in, new ExtWrapTagged(), pf);
         constraintMsg = ExtUtil.nullIfEmpty(ExtUtil.readString(in));
+        if (constraintMsg != null) {
+            constraintMsg = constraintMsg.trim();
+        }
```
dcbriccetti: There’s no way to write an untrimmed constraintMsg so this will never do anything. Right?

test/org/javarosa/form/api/ConstraintTextTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+        formEntryPrompt = new FormEntryPrompt(formDef, formEntryModel.getFormIndex());
+        assertEquals("Your message", formEntryPrompt.getConstraintText());
+
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+        formEntryPrompt = new FormEntryPrompt(formDef, formEntryModel.getFormIndex());
+        assertEquals("Message", formEntryPrompt.getConstraintText());
+    }
+}
```
dcbriccetti: The test is a big help. Thanks. Now can we remove the duplication?

screen shot 2018-07-04 at 9 35 51 pm

grzesiek2010: Please add your commit if you want me to cherry pick it or let me know if you want me to do that.

src/org/javarosa/core/model/condition/Constraint.java
grzesiek2010: It's needed if you have for example a `constraintMsg` like in the test form
`constraintMsg=" jr:itext('/data/q1:constraintMsg')"`
with a space before `jr:itext`.

test/org/javarosa/form/api/ConstraintTextTest.java
dcbriccetti: Thanks. No need to cherry pick this one.

src/org/javarosa/core/model/condition/Constraint.java
dcbriccetti: Sorry, what I meant was the trimming happens in the Constraint constructor, so at this point in the code there’s no way the constraintMsg would need trimming.

grzesiek2010: Trimming is needed in both (constructor and here) otherwise it doesn't work.

test/org/javarosa/form/api/ConstraintTextTest.java
grzesiek2010: Fixed.

src/org/javarosa/core/model/condition/Constraint.java
dcbriccetti: If you’re reading constraintMsg with readExternal, it’s been written with writeExternal, right? In order to be written with writeExternal, constraintMsg must have arrived via the constructor, which trims it. Do you follow my reasoning?

grzesiek2010: Ahhh yes you are right. Probably I opened the form for the first time without trimming in the constructor and that's why I thought it's needed after reading too. Fixed, thanks!

kkrawczyk123:
**Tested with success**

PR #323, opened 2018-06-19
Add support for question-level actions from CommCare

### LINKED ISSUE #227: Add support for xforms-value-changed event

lognaturel:
Closes #227

Following the discussion at https://github.com/opendatakit/javarosa/pull/265#issuecomment-379292630, I pulled in changes from https://github.com/dimagi/commcare-j2me/pull/243. I initially tried to cherry-pick the commits but I found the in-progress commits confusing. Because paths are different and changes like all `Vector`s becoming `List`s have been made, it's hard to get a clean patch and I ended up applying the changes manually.

The good news is that this works great with the latest Collect version on a device using the test forms both from Dimagi and from @mdudzinski.

The bad news is that I pulled in @mdudzinski's tests from #265 and those are failing. The best fix I can think of at the moment is to go up to a higher level in the test and change the question value with `FormEntryController.answerQuestion` since that's where `triggerActionsFromEvent` is called from. That seems to more closely resemble the code that a client would make.

The other thing I'm puzzled by is whether `Action` should be in `CoreModelModule`. Dimagi's code does have it. We have several tests that explicitly `registerPrototypes` for all the classes in `CoreModelModule`s and all of those would fail. If the `Action` base class isn't directly serialized then it seems it doesn't belong there?

Perhaps the biggest thing we need to decide is whether to go with this more complex approach or with what @mdudzinski had at #227. The big advantage I see to this is that it keeps us closer to Dimagi's fork and lets us pull in follow-on changes such as support for a `setlocation` action.

CC @dcbriccetti @ggalmazor @yanokwa 

Notes for testing:
- Changes have been made to the code that governs adding repeats. The CommCare branch also implements the `jr:insert` action and I worked around changes related to that. Regression testing around repeats will be important.
- The only change in behavior should be that it's now possible to nest `setvalue` in a body element and set it to happen when the enclosing element's value changes.


src/org/javarosa/core/model/FormDef.java
```diff
@@ -1553,6 +1560,11 @@ public String getAppearanceAttr() {
                 "This method call is not relevant for FormDefs getAppearanceAttr ()");
     }
 
+    @Override
+    public ActionController getActionController() {
+        return this.actionController;
```
dcbriccetti: Better without the `this.`?

src/org/javarosa/core/model/FormDef.java
```diff
@@ -162,7 +166,7 @@ public static IDataReference getAbsRef(IDataReference ref, TreeReference parentR
     private HashMap<String, DataInstance> formInstances;
     private FormInstance mainInstance = null;
 
-    private HashMap<String, List<Action>> eventListeners;
+    private ActionController actionController;
```
dcbriccetti: How do you feel about moving the initializer here?
`private ActionController actionController = new ActionController();`
Too bad it can’t be final (because of `readExternal`).

src/org/javarosa/core/model/actions/ActionController.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @author Aliza Stone
+ */
+public class ActionController implements Externalizable {
+
+    // map from an event to the actions it should trigger
+    private HashMap<String, List<Action>> eventListeners;
+
+    public ActionController() {
+        this.eventListeners = new HashMap<String, List<Action>>();
```
dcbriccetti: I’m thinking maybe you want to see if the general approach is the right one, before improving the code.
`eventListeners = new HashMap<>();`
So I’ll stop with these.

lognaturel: Yes... so any high-level thoughts? 😄

dcbriccetti:
I support this approach, especially since it smoothes the path of pulling future features from CommCare.

lognaturel:
I'm going to take a look at the tests shortly.

@dcbriccetti Any thoughts on 

> whether Action should be in CoreModelModule. Dimagi's code does have it. We have several tests that explicitly registerPrototypes for all the classes in CoreModelModules and all of those would fail. If the Action base class isn't directly serialized then it seems it doesn't belong there?

lognaturel:
Going over this in more detail, I how see that @mdudzinski had pulled in CommCare code in cea35910096031be8be1660b0a202537ab50ce3e so really the two PRs have a lot in common. 

There's something I don't quite understand yet that I need to dive more deeply into. Since `triggerActionsFromEvent` is called from `answerQuestion` which is called each time the form is validated, I would expect the value changed event to be triggered on form load, whenever a value for which the trigger is a dependent value, when the form is finalized, etc. That's not good.

That's also the major difference now between this branch and #227. That branch uses data serialization to compare old and new values (kind of hacky but there's no IAnswerData comparison; see https://github.com/opendatakit/javarosa/pull/265/files#diff-96c7e556ff24192dbc4851b9b4410bbfR380) and only triggers the event if the value has actually changed.

lognaturel:
Ok. I'm happy with where this is so it's ready for your review @dcbriccetti. Odd that the CommCare implementation didn't actually work as-is!

The one question I still have is 
> whether Action should be in CoreModelModule. Dimagi's code does have it. We have several tests that explicitly registerPrototypes for all the classes in CoreModelModules and all of those would fail. If the Action base class isn't directly serialized then it seems it doesn't belong there?

I think the `testSerializationAndDeserialization` test introduced in bc02563 is sufficient evidence that it's not needed but I'm not a (de)serialization expert.

dcbriccetti:
Looking

test/org/javarosa/core/model/actions/ValueChangedTimestampTest.java
```diff
@@ -0,0 +1,186 @@
...
+import java.nio.file.Path;
+import java.util.Date;
+
+import static org.javarosa.core.util.externalizable.ExtUtil.defaultPrototypes;
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.javarosa.xform.parse.FormParserHelper.parse;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class ValueChangedTimestampTest {
```
dcbriccetti: Could we please make these readability improvements?

https://github.com/dcbriccetti/javarosa/commit/284c5b5a370164398a184f2b38ebd2df29b1650f

dcbriccetti: (Updated)

test/org/javarosa/core/model/actions/ValueChangedTimestampTest.java
```diff
@@ -0,0 +1,186 @@
...
+    @Test
+    public void when_triggerNodeIsUpdatedWithinRepeat_targetNodeCalculation_isEvaluated() throws IOException {
+        // Given
+        final FormDef formDef =
+                parse(r("nested-setvalue-action-with-repeats.xml")).formDef;
+        FormEntryController formEntryController = new FormEntryController(new FormEntryModel(formDef));
+
+        TreeReference[] targetRefs = new TreeReference[3];
+
+        for (int i = 0; i < 3; i++) {
```
dcbriccetti: Change 3 to targetRefs.length

dcbriccetti:
I agree that registering the (abstract) Action class is not needed. Its only subclass, SetValueAction, is registered, and as the image here shows, it does get serialized and deserialized.

screen shot 2018-06-26 at 10 41 47 pm

dcbriccetti:
@lognaturel Do we mark this for testing now, or merge?

lognaturel:
Let's do a testing round. There's no rush to get this in -- the real money feature is adding an action for background geopoint. I can build on this branch for now and then rebase once it gets merged. Thanks!

kkrawczyk123:
For testing, I used two forms (which are based on form from pr)
[repeatEvent.xml.txt](https://github.com/opendatakit/javarosa/files/2176382/repeatEvent.xml.txt)
[x1.xml.txt](https://github.com/opendatakit/javarosa/files/2176383/x1.xml.txt)
which I believe are showing an intended behavior of this pr (please, correct me if I am wrong). 
I also tested other forms with repeats to look for any regression and I didn't find anything suspicious - so I think we have success here. 
@lognaturel, there is one thing that I would need clarification with, what behavior should be expected when decimal and time widgets would be on the same page? Because I am not sure if this is a bug or this scenario was not considered at all? I have an example form for that 
[x3.xml.txt](https://github.com/opendatakit/javarosa/files/2176454/x3.xml.txt)

lognaturel:
> what behavior should be expected when decimal and time widgets would be on the same page?

That's a good question. `cost_timestamp` should be set when the value of `cost` changes but should only be able to see it after you move away from the screen (swipe forward, jump, exit the form, etc) and then move back to it.

kkrawczyk123:
> That's a good question. cost_timestamp should be set when the value of cost changes but should only be able to see the timestamp value after you move away from the screen (swipe forward, jump, exit the form, etc) and then move back to it.

So I was expecting that behavior but I didn't obtained that, nothing happens with timestamp value when I swipe forward and go back or exit the form. form x3 form my previous comment. 

> A really important thing to verify in general is that the target value (e.g. timestamp) only changes if the value of the source (e.g. cost) changes. You've probably verified that but just wanted to make sure!

Yes, this is exactly what I tested :)

lognaturel:
@kkrawczyk123 I dug deeper into the field-list issue and I can't reproduce it in JavaRosa directly (I wrote a quick test). I'm pretty sure it's a symptom of https://github.com/opendatakit/collect/issues/378.

I think for now we can document it as a limitation. Another limitation is that it's not possible to have a change in value in a group set a value outside that group. That's because of the context check in `SetValueAction.processAction`.

@kkrawczyk123 if there are no other issues, I think this is ready to merge. Let's wait until after the next Collect release, though.

PR #329, opened 2018-06-26
Issue 328 fix randomize with choice filters

### LINKED ISSUE #328: randomize can't be used with choice filter that has comma
#### Software versions 
JavaRosa v2.10.x

#### Problem description
Trying to randomize an itemset with a choice filter such as `randomize(instance('fruit')/root/item[selected(/filter-rank/fruit, name)])` leads to an error because whatever is after the comma tries to be used as a seed (in this case `name)]`).

1. Transfer a form with a randomized list with choice filter such as [filter-rank.xml.txt](https://github.com/opendatakit/javarosa/files/2134965/filter-rank.xml.txt)
1. Fill the form and see the error

#### Expected behavior
No error.

ggalmazor:
Closes #328

This PR fixes getting the args of a `randomize()` function in a nodeset definition to support commas in the first argument, i.e. using a choice filter with more than one arg.

This PR will still fail if, somehow, the first argument has a comma and doesn't include the `[` character, which is the way it knows it has to run the regexp Pattern instead of the simple split-at-comma procedure.

Wrote new unit test cases to verify we support commas in the first argument.

This PR has the narrowest fix to support the use of choice filters, as described in #328.

I considered writing a unique `Pattern` that would cover any `randomize()` args but it turned out to be quite difficult and I opted for introducing an `if` that would run the old code (which is less expensive) when there is no choice filter.

None.

### DIFF
--- a/src/org/javarosa/xform/parse/RandomizeHelper.java
+++ b/src/org/javarosa/xform/parse/RandomizeHelper.java
@@ -16,6 +16,8 @@
 package org.javarosa.xform.parse;
 
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * This class contains all the code needed to implement the xform randomize()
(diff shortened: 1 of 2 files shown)


lognaturel:
🎉

PR #334, opened 2018-07-09
Test helper methods

ggalmazor:
Per @lognaturel's request, this PR has some minor reformatting changes and a class with useful test helpers.

The code in this PR is a very, very, **very** incipient work in progress. Test helpers should use high-level semantics that explain and give examples of specific form interactions, like creating a new form instance, answering a question, looking what question's label holds, etc.

The test helper methods in this PR reflect my lack of knowledge of how JR works and will probably have sense only in the context of the tests I've written but I'd love to help anyone trying to adapt them for broader use cases.

## Some ideas about testing JR

The rationale behind these test helpers is that, ideally, we could have a single entry point to handle form interaction: a form test driver that could imitate user actions in Collect and provide some introspection API to build meaningful assertions.

JavaRosa can be overwhelmingly complex and the current test code implies a deep knowledge of how JR works, involving classes such as, but not limited to, `FormParseInit`, `FormController`, `FormEntryPoint`, which have to be used in a certain way and provide APIs that cover different testing concerns that can even overlap.

For example, a JR test could have the following structure:

```java
Context(form("the_form.xml"))
  .given(answer("/the/ref/to/question1", "some value"))
  .given(answer("/the/ref/to/question2", "some other value"))

  .when(answer("/the/ref/to/question3", "yet another value"))

  .then(question("/the/ref/to/question4"), contains(choices("a", "b", "c")));
```

- The code for `Context()` and the `form()` factory should explain how the context of a test gets constructed and how a form is loaded and initialized.
- The code for `answer()` should explain how to navigate an instance to search for some question and how those values get bound.
- The code for `choices()` is a standard, composable Hamcrest matcher for `QuestionDef` objects.
- This test structure can seem too simple but thanks to type safe operation and composable matchers, it can be support much more complex interactions. It also enforces a test style that talks about behavior, which helps writers focus on what the users will perceive instead of implementation details.
- This doesn't mean that we have to avoid other tests that focus on implementation details, like when we need to implement a shuffling algorithm, for example.

As a corollary, I'd strongly suggest adding Hamcrest to the list of test dependencies in the project to start building a library of composable `Matcher`s that would greatly improve reuse of test code and expressiveness.

Run the tests

This PR is a straightforward extraction of methods to their own class using IntelliJ's safe refactors.

Nope.

### DIFF
--- a/test/org/javarosa/TestHelper.java
+++ b/test/org/javarosa/TestHelper.java
@@ -1,0 +1,93 @@
+package org.javarosa;
+
+import static org.javarosa.core.model.instance.TreeReference.CONTEXT_ABSOLUTE;
+import static org.javarosa.core.model.instance.TreeReference.INDEX_UNBOUND;
+import static org.javarosa.core.model.instance.TreeReference.REF_ABSOLUTE;
+
+import java.util.Arrays;
+import java.util.List;
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.FormIndex;
+import org.javarosa.core.model.SelectChoice;
+import org.javarosa.core.model.instance.InstanceInitializationFactory;
+import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.form.api.FormEntryPrompt;
+
+public class TestHelper {
+    /**
+     * Builds an absolute {@link TreeReference} from a string path description
+     * like <code>/some/path</code>.
+     * <p>
+     * Can be used in combination of other methods that take {@link TreeReference},
+     * like {@link #getFormIndex(FormDef, TreeReference)} or {@link #getSelectChoices(FormDef, String)}
+     * to make them easier to use. Example:
+     *
+     * <code>FormIndex formIndex = getFormIndex(formDef, absoluteRef("/data/field"));</code>
+     */
+    public static TreeReference absoluteRef(String path) {
+        TreeReference tr = new TreeReference();
+        tr.setRefLevel(REF_ABSOLUTE);
+        tr.setContext(CONTEXT_ABSOLUTE);
+        tr.setInstanceName(null);
+        Arrays.stream(path.split("/"))
+            .filter(s -> !s.isEmpty())
+            .forEach(s -> tr.add(s, INDEX_UNBOUND));
+        return tr;
+    }
+
+    /**
+     * Returns a {@link FormIndex} corresponding to the given string reference.
+     * <p>
+     * The absolute {@link TreeReference} of the question is built with {@link #absoluteRef(String)}.
+     */
+    public static FormIndex getFormIndex(FormDef formDef, String ref) {
+        return getFormIndex(formDef, absoluteRef(ref));
+    }
+
+    /**
+     * Returns a {@link FormIndex} corresponding to the given {@link TreeReference}.
+     */
+    public static FormIndex getFormIndex(FormDef formDef, TreeReference ref) {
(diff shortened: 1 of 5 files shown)


test/org/javarosa/TestHelper.java
```diff
@@ -0,0 +1,93 @@
...
+     * <code>FormIndex formIndex = getFormIndex(formDef, absoluteRef("/data/field"));</code>
+     */
+    public static TreeReference absoluteRef(String path) {
+        TreeReference tr = new TreeReference();
+        tr.setRefLevel(REF_ABSOLUTE);
+        tr.setContext(CONTEXT_ABSOLUTE);
+        tr.setInstanceName(null);
+        Arrays.stream(path.split("/"))
+            .filter(s -> !s.isEmpty())
+            .forEach(s -> tr.add(s, INDEX_UNBOUND));
```
lognaturel: Does `INDEX_UNBOUND` work for nodes in groups?

test/org/javarosa/TestHelper.java
```diff
@@ -0,0 +1,93 @@
...
+     * The absolute {@link TreeReference} of the question is built with {@link #absoluteRef(String)}.
+     */
+    public static FormIndex getFormIndex(FormDef formDef, String ref) {
+        return getFormIndex(formDef, absoluteRef(ref));
+    }
+
+    /**
+     * Returns a {@link FormIndex} corresponding to the given {@link TreeReference}.
+     */
+    public static FormIndex getFormIndex(FormDef formDef, TreeReference ref) {
```
lognaturel: This would need to be recursive to work for a depth more than 1 (nodes in groups or repeats). I'm starting to prefer what I did in https://github.com/opendatakit/javarosa/commit/ee5527fd3c2d49549e288de6beaab5df48068be8 because it guarantees it gets something that's actually in the form with all the correct multiplicities, context, etc set. I also like that it's not in a grab-bag of static methods. Do you prefer keeping all methods most helpful for testing together?

ggalmazor: Yeah, this only makes sense for the tests I've used it with, which doesn't include groups. 

I have many doubts about all the methods in this class because I haven't worked with JR enough to understand what's useful and what's not. For example, in your ee5527f you return a `TreeElement` instead of a `FormIndex`. Is that more generally useful or just for the tests you're writing? `¯\_(ツ)_/¯`

If it turns out that the implementation in ee5527f is more useful, I'd suggest we improve this method with it and you use it in your PR.

In my approach, this would be the first step to generalize a BDD harness for tests: get every test utility method into the same bag and start figuring out what structures emerge. This is not something that we can do in one PR and we will have lots of duplicated ugly code, but that'd be by design.

Another approach would be to forget about that and start top-down implementing everything required to write this:
```
Context(form("the_form.xml"))
  .given(answer("/the/ref/to/question1", "some value"))
  .given(answer("/the/ref/to/question2", "some other value"))

  .when(answer("/the/ref/to/question3", "yet another value"))

  .then(question("/the/ref/to/question4"), contains(choices("a", "b", "c")));
```
We can do that too and take what we do from all the different tests we already have. I can prepare a scaffolding for that but I'd need help to write the implementations

dcbriccetti:
Issue 273 is closed. Is there a better name for this PR?

ggalmazor:
I've renamed the class. I'd wait for #335 to do anything on this one, though

ggalmazor:
Closing this one in favor of an upcoming PR to solve #349

PR #335, opened 2018-07-11
[WIP] Fix current() expansion in selects

### LINKED ISSUE #293: current() in choice filter is inconsistent with current() elsewhere

lognaturel:
Closes #293

TODO:
- [ ] clean up comments around `contextRef`
- [ ] possibly add more tests for `current()` usage in other contexts?

Ran automated tests and tried form in Collect. To understand what's going on, I recommend pulling each commit in order, reading the commit comment and running the tests.

The changes themselves are very small and surgical. The biggest question is whether the tests are sufficient to feel confident that they don't result in regressions. I'd love your feedback on this, @dcbriccetti and @ggalmazor. In particular, I don't think there's yet enough testing to feel confident in 507f56d.

I considered removing `contextRef` altogether but I'd rather do that once we have confidence in the approach either in more commits here or in a future PR.

I'm fairly confident that the tests for the change to validate 20e357f are exhaustive. I think we've identified all the places `contextRef` is used in the discussion at #293 and tested all such cases. 

`getAbsRef` is used more broadly so I'm less confident in the change in 507f56d. Risks include other relative expressions using `..` or `current()` or `.` not being expanded properly.

### DIFF
--- a/src/org/javarosa/core/model/FormDef.java
+++ b/src/org/javarosa/core/model/FormDef.java
@@ -122,7 +122,7 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
             tref = TreeReference.selfRef(); //only happens for <group>s with no binding
         }
 
-        tref = tref.parent(parentRef);
+        tref = tref.anchor(parentRef);
         if (tref == null) {
             throw new XFormParseException("Binding path [" + tref + "] not allowed with parent binding of [" + parentRef + "]");
         }
--- a/src/org/javarosa/core/model/FormDef.java
+++ b/src/org/javarosa/core/model/FormDef.java
@@ -353,6 +353,10 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
         return ref;
     }
 
+    public TreeElement getFirstDescendantWithName(String name) {
+        return mainInstance.getRoot().getFirstDescendantWithName(name);
+    }
+
     public void setLocalizer(Localizer l) {
         if (this.localizer != null) {
             this.localizer.unregisterLocalizable(this);
--- a/src/org/javarosa/core/model/instance/TreeElement.java
+++ b/src/org/javarosa/core/model/instance/TreeElement.java
@@ -36,6 +36,7 @@ import org.javarosa.core.model.instance.utils.DefaultAnswerResolver;
 import org.javarosa.core.model.instance.utils.IAnswerResolver;
 import org.javarosa.core.model.instance.utils.ITreeVisitor;
 import org.javarosa.core.model.instance.utils.TreeElementChildrenList;
+import org.javarosa.core.model.instance.utils.TreeElementNameComparator;
 import org.javarosa.core.model.util.restorable.RestoreUtils;
 import org.javarosa.core.util.DataUtil;
 import org.javarosa.core.util.externalizable.DeserializationException;
@@ -225,6 +226,28 @@ import org.javarosa.xpath.expr.XPathStringLiteral;
     @Override
     public List<TreeElement> getChildrenWithName(String name) {
         return children.get(name);
+    }
+
+    /** Returns the first element across descendants that has the specified name. Useful for testing. **/
+    public TreeElement getFirstDescendantWithName(String name) {
+        return getFirstDescendantWithName(this, name);
+    }
+
+    private TreeElement getFirstDescendantWithName(TreeElement node, String name) {
+        if (node.getMultiplicity() != TreeReference.INDEX_TEMPLATE &&
+                TreeElementNameComparator.elementMatchesName(node, name)) {
+            return node;
+        }
+
+        for (TreeElement child : node.children) {
+            TreeElement result = getFirstDescendantWithName(child, name);
+
+            if (result != null) {
+                return result;
+            }
+        }
+
+        return null;
     }
 
     private int getNumChildrenWithName(String name) {
--- a/src/org/javarosa/form/api/FormEntryController.java
+++ b/src/org/javarosa/form/api/FormEntryController.java
@@ -21,6 +21,7 @@ import org.javarosa.core.model.QuestionDef;
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.instance.InvalidReferenceException;
 import org.javarosa.core.model.instance.TreeElement;
+import org.javarosa.core.model.instance.TreeReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -213,6 +214,28 @@ public class FormEntryController {
     public int jumpToIndex(FormIndex index) {
         model.setQuestionIndex(index);
         return model.getEvent(index);
+    }
+
+    /**
+     * Jump to the first node with the given name. Useful for testing.
+     */
+    public int jumpToFirstQuestionWithName(String name) {
+        return jumpToReference(getModel().getForm().getFirstDescendantWithName(name).getRef());
+    }
+
+    private int jumpToReference(TreeReference reference) {
+        jumpToIndex(FormIndex.createBeginningOfFormIndex());
+        FormIndex index = model.getFormIndex();
+
+        do {
+            if (index.getReference() != null && index.getReference().equals(reference)) {
+                return jumpToIndex(index);
+            }
+
+            index = model.incrementIndex(index);
+        } while (index.isInForm());
+
+        return jumpToIndex(index);
     }
 
     public FormIndex descendIntoRepeat (int n) {
--- a/test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
+++ b/test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2018 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.javarosa.xpath.expr;
+
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.ItemsetBinding;
+import org.javarosa.core.model.SelectChoice;
+import org.javarosa.core.model.data.SelectOneData;
+import org.javarosa.core.model.data.StringData;
+import org.javarosa.core.model.data.helper.Selection;
+import org.javarosa.core.model.instance.InstanceInitializationFactory;
+import org.javarosa.core.test.FormParseInit;
+import org.javarosa.form.api.FormEntryController;
+import org.javarosa.form.api.FormEntryModel;
+import org.javarosa.form.api.FormEntryPrompt;
+import org.junit.Before;
+import org.junit.Test;
+
+import static junit.framework.TestCase.assertEquals;
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+
+public class XPathPathExprCurrentItemsetNodesetTest {
+    private FormDef formDef;
+    private FormEntryController formEntryController;
+
+    @Before
+    public void setUp() {
+        FormParseInit fpi = new FormParseInit(r("relative-current-ref-itemset-nodeset.xml"));
+        formDef = fpi.getFormDef();
+        formDef.initialize(true, new InstanceInitializationFactory());
+        FormEntryModel formEntryModel = new FormEntryModel(formDef);
+        formEntryController = new FormEntryController(formEntryModel);
+    }
+
+    /**
+     * current() in an itemset nodeset expression should refer to the select node. This is verified by building an
+     * itemset from a repeat which is a sibling of the select.
+     */
+    @Test
+    public void current_in_itemset_nodeset_should_refer_to_node() {
+        // don't know how to jump to repeat directly so jump to following question and step backwards
+        formEntryController.jumpToFirstQuestionWithName("selected_person");
+        formEntryController.stepToPreviousEvent(); // repeat
+        formEntryController.descendIntoRepeat(0);
+        formEntryController.stepToNextEvent(); // person
+
+        StringData nameValue = new StringData("Bob");
+        formEntryController.answerQuestion(nameValue, true);
+
+        formEntryController.stepToNextEvent();
+        formEntryController.descendIntoNewRepeat();
+        formEntryController.stepToNextEvent(); // person
+
+        StringData nameValue2 = new StringData("Janet");
+        formEntryController.answerQuestion(nameValue2, true);
+
+        formEntryController.jumpToFirstQuestionWithName("selected_person");
+        FormEntryPrompt personPrompt = formEntryController.getModel().getQuestionPrompt();
+        personPrompt.getAnswerValue();
+        ItemsetBinding dynamicChoices = personPrompt.getQuestion().getDynamicChoices();
+
+        SelectChoice personChoice = dynamicChoices.getChoices().get(1);
+        assertEquals("Janet", personPrompt.getSelectChoiceText(personChoice));
+        SelectOneData personSelection = new SelectOneData(new Selection(personChoice));
+        formEntryController.answerQuestion(personSelection, true);
+
+        SelectOneData personSelectionValue = (SelectOneData) formDef.getFirstDescendantWithName("selected_person")
+                .getValue();
+        assertEquals("2", personSelectionValue.getDisplayText());
+    }
+}
--- a/test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
+++ b/test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2018 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.javarosa.xpath.expr;
+
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.ItemsetBinding;
+import org.javarosa.core.model.SelectChoice;
+import org.javarosa.core.model.data.SelectOneData;
+import org.javarosa.core.model.data.StringData;
+import org.javarosa.core.model.data.helper.Selection;
+import org.javarosa.core.model.instance.InstanceInitializationFactory;
+import org.javarosa.core.test.FormParseInit;
+import org.javarosa.form.api.FormEntryController;
+import org.javarosa.form.api.FormEntryModel;
+import org.javarosa.form.api.FormEntryPrompt;
+import org.junit.Before;
+import org.junit.Test;
+
+import static junit.framework.TestCase.assertEquals;
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+
+public class XPathPathExprCurrentTest {
+    private FormDef formDef;
+    private FormEntryController formEntryController;
+
+    @Before
+    public void setUp() {
+        FormParseInit fpi = new FormParseInit(r("relative-current-ref.xml"));
+        formDef = fpi.getFormDef();
+        formDef.initialize(true, new InstanceInitializationFactory());
+        FormEntryModel formEntryModel = new FormEntryModel(formDef);
+        formEntryController = new FormEntryController(formEntryModel);
+    }
+
+    /**
+     * current() in a calculate should refer to the node it is in (in this case, /data/my_group/name_relative).
+     * This means that to refer to a sibling node, the path should be current()/../<name of sibling node>. This is
+     * verified by changing the value of the node that the calculate is supposed to refer to
+     * (/data/my_group/name) and seeing that the dependent calculate is updated accordingly.
+     */
+    @Test
+    public void current_in_calculate_should_refer_to_node() {
+        formEntryController.jumpToFirstQuestionWithName("name");
+        StringData nameValue = new StringData("Bob");
+        formEntryController.answerQuestion(nameValue, true);
+
+        StringData relativeName = (StringData) formDef.getFirstDescendantWithName("name").getValue();
+        assertEquals(nameValue.getValue(), relativeName.getValue());
+    }
+
+    /**
+     * current() in a choice filter should refer to the select node the choice filter is called from, NOT the expression
+     * it is in. See https://developer.mozilla.org/en-US/docs/Web/XPath/Functions/current -- this is the difference
+     * between current() and .
+     *
+     * The behavior of current() in a choice filter is verified by selecting a value for a first, static select and then
+     * using that value to filter a second, dynamic select.
+     */
+    @Test
+    public void current_in_choice_filter_should_refer_to_node() {
+        formEntryController.jumpToFirstQuestionWithName("fruit");
+        Selection fruitSelection = new Selection(1);
+        fruitSelection.attachChoice(formEntryController.getModel().getQuestionPrompt().getQuestion());
+        formEntryController.answerQuestion(new SelectOneData(fruitSelection), true);
+
+        SelectOneData fruitSelectionValue = (SelectOneData) formDef.getFirstDescendantWithName("fruit").getValue();
+        assertEquals("blueberry", fruitSelectionValue.getDisplayText());
+
+        // === Variety question ===
+        formEntryController.stepToNextEvent();
+        // Collect calls getAnswerValue to populate dynamic choices
+        FormEntryPrompt varietyPrompt = formEntryController.getModel().getQuestionPrompt();
+        varietyPrompt.getAnswerValue();
+        ItemsetBinding dynamicChoices = varietyPrompt.getQuestion().getDynamicChoices();
+
+        // There are three blueberry varieties defined by the form
+        assertEquals(dynamicChoices.getChoices().size(), 3);
+
+        SelectChoice varietyChoice = dynamicChoices.getChoices().get(1);
+        SelectOneData varietySelection = new SelectOneData(new Selection(varietyChoice));
+        assertEquals("Collins", varietyPrompt.getSelectChoiceText(varietyChoice));
+        formEntryController.answerQuestion(varietySelection, true);
+
+        SelectOneData variety = (SelectOneData) formDef.getFirstDescendantWithName("variety").getValue();
+        assertEquals("collins", variety.getDisplayText());
+    }
+}
--- a/resources/org/javarosa/xpath/expr/relative-current-ref-itemset-nodeset.xml
+++ b/resources/org/javarosa/xpath/expr/relative-current-ref-itemset-nodeset.xml
@@ -1,0 +1,51 @@
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa">
+    <h:head>
+        <h:title>Untitled Form</h:title>
+        <model>
+            <instance>
+                <data id="relative-current-ref-repeat">
+                    <meta>
+                        <instanceID/>
+                    </meta>
+                    <people>
+                        <count/>
+                        <person/>
+                    </people>
+
+                    <selected_person/>
+                </data>
+            </instance>
+            <itext>
+                <translation lang="English">
+                    <text id="/data/people:label">
+                        <value>People</value>
+                    </text>
+                    <text id="/data/people/person:label">
+                        <value>Person</value>
+                    </text>
+                </translation>
+            </itext>
+            <bind nodeset="/data/meta/instanceID" type="string" readonly="true()" calculate="concat('uuid:', uuid())"/>
+            <bind nodeset="/data/people/count" type="int" calculate="position(..)"/>
+            <bind nodeset="/data/people/person" type="string"/>
+            <bind nodeset="/data/selected_person" type="select1"/>
+        </model>
+    </h:head>
+    <h:body>
+        <group>
+            <label ref="jr:itext('/data/people:label')"/>
+            <repeat nodeset="/data/people">
+                <input ref="/data/people/person">
+                    <label ref="jr:itext('/data/people/person:label')"/>
+                </input>
+            </repeat>
+
+            <select1 ref="/data/selected_person">
+                <itemset nodeset="current()/../people">
+                    <value ref="count"/>
+                    <label ref="person"/>
+                </itemset>
+            </select1>
+        </group>
+    </h:body>
+</h:html>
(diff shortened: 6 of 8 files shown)


src/org/javarosa/core/model/FormDef.java
```diff
@@ -353,6 +353,10 @@ public TreeReference getChildInstanceRef(List<IFormElement> elements, List<Integ
         return ref;
     }
 
+    public TreeElement getFirstDescendantWithName(String name) {
```
ggalmazor: If this is only used for testing, it would be better to have it in another separate class under the test sources root. It will make our production classes thinner and more manageable and will clearly set the right context (testing vs. production) for all future readers.

Moving this to another separate class under the test sources root would also contribute towards our goal for a test harness by adding more testing use cases in a centralized place, which will make easier to make sense of the testing needs in JR.

test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        formEntryController.jumpToFirstQuestionWithName("selected_person");
+        FormEntryPrompt personPrompt = formEntryController.getModel().getQuestionPrompt();
+        personPrompt.getAnswerValue();
+        ItemsetBinding dynamicChoices = personPrompt.getQuestion().getDynamicChoices();
+
+        SelectChoice personChoice = dynamicChoices.getChoices().get(1);
+        assertEquals("Janet", personPrompt.getSelectChoiceText(personChoice));
+        SelectOneData personSelection = new SelectOneData(new Selection(personChoice));
+        formEntryController.answerQuestion(personSelection, true);
```
ggalmazor: ~It's hard to figure out what we're asserting without studying the blank form. Also, why are we jumping to the same question we answered with "Bob" and "Janet" in the first part? Some comments would help.~

It took me several times going back and forth and read the form definition to understand what was going on. The most confusing part for me was to see the same "selected_person" on lines 55 and 70. I though we were answering "selected_person" from lines 55 to 70, but now I understand that we were using that question as our first pivot point in the form's DOM to navigate to the "/data/people" repeat. 

Also, is really necessary to actually make a selection here? Would it be enough to just check that there are two choices and that they match "Bob" and "Janet"?

test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        SelectOneData personSelectionValue = (SelectOneData) formDef.getFirstDescendantWithName("selected_person")
+                .getValue();
+        assertEquals("2", personSelectionValue.getDisplayText());
```
ggalmazor: Why is 2 the display text value for "Janet"? Where does that 2 come from?

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
```diff
@@ -0,0 +1,100 @@
...
+     * current() in a calculate should refer to the node it is in (in this case, /data/my_group/name_relative).
+     * This means that to refer to a sibling node, the path should be current()/../<name of sibling node>. This is
+     * verified by changing the value of the node that the calculate is supposed to refer to
+     * (/data/my_group/name) and seeing that the dependent calculate is updated accordingly.
+     */
+    @Test
+    public void current_in_calculate_should_refer_to_node() {
+        formEntryController.jumpToFirstQuestionWithName("name");
+        StringData nameValue = new StringData("Bob");
+        formEntryController.answerQuestion(nameValue, true);
```
ggalmazor: So, `answerQuestion("/data/my_group/name", "Bob");`

It's confusing not seeing the complete path of the model we're filling since there can be multiple "name" fields.

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        StringData relativeName = (StringData) formDef.getFirstDescendantWithName("name").getValue();
+        assertEquals(nameValue.getValue(), relativeName.getValue());
```
ggalmazor: Maybe `assertThat(answerOf("/data/my_group/name_relative"), is("Bob"))`
(we would need Hamcrest for that)

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * The behavior of current() in a choice filter is verified by selecting a value for a first, static select and then
+     * using that value to filter a second, dynamic select.
+     */
+    @Test
+    public void current_in_choice_filter_should_refer_to_node() {
+        formEntryController.jumpToFirstQuestionWithName("fruit");
+        Selection fruitSelection = new Selection(1);
+        fruitSelection.attachChoice(formEntryController.getModel().getQuestionPrompt().getQuestion());
+        formEntryController.answerQuestion(new SelectOneData(fruitSelection), true);
```
ggalmazor: `answerQuestion("/data/fruit", "blueberry");`

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        SelectOneData fruitSelectionValue = (SelectOneData) formDef.getFirstDescendantWithName("fruit").getValue();
+        assertEquals("blueberry", fruitSelectionValue.getDisplayText());
```
ggalmazor: Is this just for confirmation?

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        // === Variety question ===
+        formEntryController.stepToNextEvent();
+        // Collect calls getAnswerValue to populate dynamic choices
+        FormEntryPrompt varietyPrompt = formEntryController.getModel().getQuestionPrompt();
+        varietyPrompt.getAnswerValue();
+        ItemsetBinding dynamicChoices = varietyPrompt.getQuestion().getDynamicChoices();
```
ggalmazor: `List<SelectChoice> choices = getChoicesOf("/data/variety");`

And then `getChoicesOf()` takes care of calling whatever method needs to be called.

In Briefcase, I managed to [populate dynamic choices](https://github.com/ggalmazor/briefcase/commit/00df91b05a8de61355daeca1d375d23e78437c07#diff-5a1a0d134a81a1a17735fb53c0b5c563R120) by doing this: 
```java
formDef.initialize(false, new InstanceInitializationFactory());
formDef.populateDynamicChoices(control.getDynamicChoices(), (TreeReference) control.getBind().getReference());
```

I don't know if it could be applied here.

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        // There are three blueberry varieties defined by the form
+        assertEquals(dynamicChoices.getChoices().size(), 3);
+
+        SelectChoice varietyChoice = dynamicChoices.getChoices().get(1);
```
ggalmazor: Are the next assertions really necessary to verify the behavior of `current()` in this context?

test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
ggalmazor: Never mind! I figured out this is the count of people :)

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
lognaturel: The form def initialization is in the `before` block.

`populateDynamicChoices` is called by `getAnswerValue()`. I did it this way to most closely match what Collect does (it doesn't call `populateDynamicChoices` directly).

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
lognaturel: I suppose not. This block is mimicking what a client does when actually making a selection. It feels more complete to be able to say with certainty that a choice can not only be saved but also successfully retrieved from the model after that but I suppose we could do that in a separate test.

ggalmazor: Sorry to be picky but it's a huge load of complex stuff that's going on. Anything we can remove will make the whole easier to understand for everyone.

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
ggalmazor: So, the main advantage of using the `populateDynamicChoices` versus the `getAnswerValue()` is that it doesn't depend on where the current form index is located, making it less prone to error and easier to understand.

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
lognaturel: I built the answer by selecting an index of 1. This gives me some certainty that I have actually picked the value I had intended to so that if the assertions below fail I know for sure it's not because of an issue with the static select or the form or something like that. Tests that are based on reading form XML can break catastrophically because of a small form change so I like to have breadcrumbs there ensuring the different steps actually did what they were supposed to.

ggalmazor: Then I'd like to suggest putting a comment to explain that the assertion is not part of the test.

src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -227,6 +228,28 @@ public TreeElement getChild(String name, int multiplicity) {
         return children.get(name);
     }
 
+    /** Returns the first element across descendants that has the specified name. Useful for testing. **/
+    public TreeElement getFirstDescendantWithName(String name) {
+        return getFirstDescendantWithName(this, name);
+    }
+
+    private TreeElement getFirstDescendantWithName(TreeElement node, String name) {
+        if (node.getMultiplicity() != TreeReference.INDEX_TEMPLATE &&
```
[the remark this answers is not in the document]
lognaturel: This is from other search methods such as `findChildrenWithName`. It's important because we want to only identify actual nodes, not templates. Repeats can be templates. This means they're not part of the instance but that their values are used for all repeats that are created. You can read more about `jr:template` at https://opendatakit.github.io/xforms-spec/#primary-instance. So no, it has nothing to do with whether it has children. We want to know whether the node is a template and if so not consider it as a result. I guess the semantic method would be `!node.isTemplate()`.

src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -227,6 +228,28 @@ public TreeElement getChild(String name, int multiplicity) {
...
+                TreeElementNameComparator.elementMatchesName(node, name)) {
```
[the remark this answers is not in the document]
lognaturel: I'm trying to be minimally invasive, here! 😄You'll see other search methods do the same.

src/org/javarosa/core/model/FormDef.java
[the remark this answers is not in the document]
lognaturel: Ok. I think this kind of code can be hard to understand without context and that it feels odd to call it not on a `formDef` but I guess I don't mind too much.

src/org/javarosa/form/api/FormEntryController.java
```diff
@@ -215,6 +216,28 @@ public int jumpToIndex(FormIndex index) {
...
+    /**
+     * Jump to the first node with the given name. Useful for testing.
+     */
+    public int jumpToFirstQuestionWithName(String name) {
+        return jumpToReference(getModel().getForm().getFirstDescendantWithName(name).getRef());
+    }
+
+    private int jumpToReference(TreeReference reference) {
+        jumpToIndex(FormIndex.createBeginningOfFormIndex());
+        FormIndex index = model.getFormIndex();
```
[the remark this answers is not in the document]
lognaturel: Yes, they're the same.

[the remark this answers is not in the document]
ggalmazor: OK, I think today I've improved a lot my understanding of how JR works

test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        StringData nameValue = new StringData("Bob");
+        formEntryController.answerQuestion(nameValue, true);
+
+        formEntryController.stepToNextEvent();
+        formEntryController.descendIntoNewRepeat();
+        formEntryController.stepToNextEvent(); // person
+
+        StringData nameValue2 = new StringData("Janet");
+        formEntryController.answerQuestion(nameValue2, true);
```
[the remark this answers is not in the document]
lognaturel: Agreed the jump to `selected_person` is odd. I wanted to do exactly what a client might do but am not too picky about how it eventually gets structured now that I know for sure that it works. 😄

test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        formEntryController.jumpToFirstQuestionWithName("selected_person");
+        FormEntryPrompt personPrompt = formEntryController.getModel().getQuestionPrompt();
+        personPrompt.getAnswerValue();
```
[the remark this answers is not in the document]
lognaturel: Discussed at https://github.com/opendatakit/javarosa/pull/335/files/1f5ad90dfa23f7e271bd865454f9cb8161c3c8de..b6aa58d32ae24e07a8d17d497ba0f4ceed42313a#diff-09a9326da058a89cc2db380641519257

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
lognaturel: Sure. Something weird happened but I think you saw my comment that used to be here?

> I built the answer by selecting an index of 1. This gives me some certainty that I have actually picked the value I had intended to so that if the assertions below fail I know for sure it's not because of an issue with the static select or the form or something like that. Tests that are based on reading form XML can break catastrophically because of a small form change so I like to have breadcrumbs there ensuring the different steps actually did what they were supposed to.

ggalmazor: Yup

test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
lognaturel: This is the same pattern as https://github.com/opendatakit/javarosa/pull/335#discussion_r201705617. I didn't feel totally confident in the test until I knew the round trip worked.

test/org/javarosa/xpath/expr/XPathPathExprCurrentItemsetNodesetTest.java
lognaturel: For this specific test, I feel it's really important to make sure both value and label are correctly populated because those come from relative expressions built off of the nodeset path. This first assert confirms that the value was correctly resolved. The assert before it verifies that the label was correctly resolved.  

Like with the other related tests, I'd be ok with doing both asserts where the first one is and dropping the "round trip" aspect of saving a value and reading it back.

As an additional note, I also think it's important to not check the first choice here. For example, using `position()` rather than `position(..)` in the form would mean all choices have value 1 and that could hide another problem. So in general, checking something where an index actually has to change seems like a good idea.

lognaturel:
@dcbriccetti I'd also appreciate your thoughts on whether:
1) verifying a relative itemset nodeset declaration using `current()` and a relative choice filter expression using `current()` are enough to give us confidence in the `ItemsetBinding.contextRef` change
2) we have sufficient coverage for the change from `parent` to `anchor` in `getAbsRef` and if not, what high-level scenarios we should verify.

My preference would be to stay with relatively low-level tests as we are still all improving our understanding of what abstractions would be helpful to include in test infrastructure. That said, I do see @ggalmazor has already done a lot to convert these tests and so as long as we're all in agreement about what coverage is needed I can close this and turn it over to him.

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        // === Variety question ===
+        formEntryController.stepToNextEvent();
+        // Collect calls getAnswerValue to populate dynamic choices
+        FormEntryPrompt varietyPrompt = formEntryController.getModel().getQuestionPrompt();
+        varietyPrompt.getAnswerValue();
+        ItemsetBinding dynamicChoices = varietyPrompt.getQuestion().getDynamicChoices();
+
+        // There are three blueberry varieties defined by the form
+        assertEquals(dynamicChoices.getChoices().size(), 3);
```
dcbriccetti: In case we keep this, the 3 should be the first argument (the expected value).

dcbriccetti:
Trying to understand this so I can help, I’m making a rough diagram.
screen shot 2018-07-15 at 5 02 15 pm

src/org/javarosa/core/model/instance/TreeElement.java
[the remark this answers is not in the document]
dcbriccetti: >some static member of TreeReference ...
Do I detect a hint of derision? Heh heh heh. 
I, also, was bothered by that when I first saw it many moons ago.

src/org/javarosa/form/api/FormEntryController.java
```diff
@@ -215,6 +216,28 @@ public int jumpToIndex(FormIndex index) {
         return model.getEvent(index);
     }
 
+    /**
+     * Jump to the first node with the given name. Useful for testing.
+     */
+    public int jumpToFirstQuestionWithName(String name) {
```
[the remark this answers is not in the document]
dcbriccetti: I find myself persuaded by the argument that these would be better in test classes.

In Scala we might use an `implicit` to “monkey-patch” those two test-related methods onto FormDef for testing.

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
dcbriccetti: I would prefer the simpler assertEqual (but you already know that). (What does Hamcrest have on you‽ 😁)

dcbriccetti:
For too much to keep in mind.
screen shot 2018-07-15 at 9 49 53 pm

dcbriccetti:
It’s taking me hours to catch up to where you two are, and to where I can help. It’s a good time investment, this study, even if you finish before I reach you.

ggalmazor:
Hi, @lognaturel and @dcbriccetti!

I'm learning a huge lot too :) Here are my insights:

## Regarding the changes in the `ItemsetBinding.contextRef` and `FormDef.getAbsRef()`

Following the code, it's kind of clear that the `contextRef` field is used only when populating the dynamic choices of an itemset binding. We could maybe add an extra test scenario, to verify that we haven't broken the "normal operation" of an itemset binding.

- Test a dynamic itemset that doesn't use `current()` as the root of the nodeset, and without a predicate.

The change in `FormDef.getAbsRef()` is harder to assess because it's broadly used while parsing.

I don't know how confident we can be about this change with the current test suite's coverage, as in "since no old test is failing after the change, we're sure it's safe". If we're not sure, we could add some test cases (all using relative paths):

- `submission` inside the model of a form  [XFormParser.java#753](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L753)
- `setvalue` action [XFormParser.java#689](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L689)
- A form field with a binding with a relative path [XFormParser.java#967](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L967)
  - Specifically, a group with a binding with a relative path [XFormParser.java#1464](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L1464) and [XFormParser.java#1473](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L1473)

## Regarding the changes to support `currrent()`, generally speaking

If my understanding of the situation is correct, we know that currently there's almost no use of `current()`, so we can't ask users for other use cases. We need to think of them.

We currently have two use cases covered:

- A `<select>` field inside potentially nested repeats that takes its choices a list of choices in the same repeat group.

```xml
<select1 ref="/data/repeat_group/selected_person">
    <itemset nodeset="current()/../people">
        <value ref="count"/>
        <label ref="person"/>
    </itemset>
</select1>
```
 - A `<select>` field inside a potentially nested repeat that takes its choices from a non-main instance, filtered with some value from another field in the same repeat group:

```xml
<select1 ref="/data/my_repeat/repeat_variety">
    <label>Variety</label>
    <itemset nodeset="instance('variety')/root/item[fruit = current()/../repeat_fruit]">
        <value ref="name"/>
        <label ref="jr:itext(itextId)"/>
    </itemset>
</select1>
```

So, _what other scenarios make sense to cover?_ I don't really know, but this is a list of what comes to mind.

- An obvious one would be to combine `current()` in the the root **and** in the predicate of the same `nodeset`.
- Any other relevant use of `current()` outside a `<select>` field
  - For example, in a `<label>` that shows an `<itext>` related to the selected value from a previous field.

lognaturel:
Thank you both for the deep dive!

## Responses to comments about the `ItemsetBinding.contextRef` change

> Test a dynamic itemset that doesn't use current() as the root of the nodeset, and without a predicate.

Sounds good to me. As far as I can tell all the select tests (`SelectOneDataTests`, etc) use static choices (so no `ItemsetBinding`). 

> we know that currently there's almost no use of current(), so we can't ask users for other use cases. We need to think of them.

I was initially thinking we should verify broader usage of `current()` but after further analysis I'm pretty sure it's not necessary. We know the change in 20e357f is tightly focused on relative expressions  in `itemsets` so we just need to ensure coverage related to that. Furthermore, the only case where `current()` is required (because it's distinct from `.`) is in a predicate. It's possible users are using it elsewhere but it would be odd usage. For example, when building a nodeset from a relative path as in `current_in_itemset_nodeset_should_refer_to_node`, it would really be nicer to use `./../people` or `../people` (both also work). I used `current()` in the test form for consistency but they're identical.

> An obvious one would be to combine current() in the the root and in the predicate of the same nodeset.

The two are evaluated independently so I don't think it broadens coverage but I'm not against adding the case.

> Any other relevant use of current() outside a <select> field
For example, in a <label> that shows an <itext> related to the selected value from a previous field.

Like I said above, I don't think this is really affected by the change here but I think it would be a good idea to have a series of tests around evaluating relative expressions in general.

## Further analysis on parent() vs. anchor()

This comment strongly suggests that the two methods are nearly identical:
https://github.com/opendatakit/javarosa/blob/60da3858b1f95b581ba55349a76a8097c4c4eaa8/src/org/javarosa/core/model/instance/TreeReference.java#L287

As far as I understand, they just have different requirements and limitations.

`parent()` clones the parent and then copies the levels from the node to be parented to the clone. To make matters a bit more complicated, I think there may be a bug in the `refLevel > 0` conditional. https://github.com/opendatakit/javarosa/blob/60da3858b1f95b581ba55349a76a8097c4c4eaa8/src/org/javarosa/core/model/instance/TreeReference.java#L239
 This is for the "parent ref is a relative ref consisting only of other 'parent' steps" case described in the comment. I can't actually think of when this would get used but it doesn't make sense to me that in that case the `parentRef`'s level would be mutated. I think the intent was to mutate `newRef`. That would make sense -- I could do `../../foo.parent(../)` and get `../../../bar/foo`. That is, `newRef`'s `refLevel` would be 3. I'm 98% sure that was the intent but again, I can't think of a form that would hit that case. There's no coverage in the current tests suite. Any thoughts on that?

`anchor()` also clones the parent and copies the levels from the node to be parented to the clone. There are 2 differences I can identify between the two methods:
- `anchor` removes levels from the cloned parent for each `..` in the node to anchor. As described above, `parent` just increments `refLevel` for each `..` (or intends to) so it can't deal with any additional levels on a parent with parent steps. That is, `../../../` is a valid parent for `parent()` but `../../../foo` is not. Both are accepted by `anchor()`.
- `anchor` requires the parent/context node to be absolute. At first glance, I thought this distinction was not really very important given the likely bug I described above. But actually it's really important because of the `parent()` usage in `TreeReference.BuildRef`. It looks like when forms are read from XML, references end up with `contextType` `CONTEXT_INHERITED` and they still need to be parented. Try changing the call on `parent()` to `anchor()` in `TreeReference.BuildRef` to see what I mean. I don't think this applies to any use of `getAbsRef`, though. 

The last parts of `anchor()` and `parent()` are identical:
```
for(TreeReferenceLevel l : data) {
  newRef.add(l.shallowCopy());
}
```

```java
for (int i = 0; i < size(); i++) {
  newRef.add(data.get(i).shallowCopy());
}
```

This analysis gives me pretty high confidence that the change from `parent` to `anchor` in `getAbsRef` is safe **as long as we can't come up with a case where the parent/context node needs to be relative**.

> If we're not sure, we could add some test cases (all using relative paths):
submission inside the model of a form XFormParser.java#753
setvalue action XFormParser.java#689
A form field with a binding with a relative path XFormParser.java#967
Specifically, a group with a binding with a relative path XFormParser.java#1464 and XFormParser.java#1473

These look like good ideas as part of the series of relative path tests previously discussed. 

An alternative specifically in support of this change is that we could come up with cases of pairs of references and the context to parent or anchor against and demonstrate that `anchor` does the same thing as `parent` for all cases that `parent` supports. An example pair would be a relative reference that just consists of a node name (e.g. `age`) and an absolute reference to the document root (e.g. `/data`). We'd expect both `anchor` and `parent` to return an absolute ref with path `/data/age`. This might also provide us with good, targeted coverage for eventually merging the two.

Where do we want to go from here? I see a few courses of action:
- Hop on a short call to try to untangle that last big unknown -- can we come up with a form that results in `getAbsRef` being called with a relative parent -- and to figure out next steps.
- Merge this so we have a record of the first pass low-level tests and then @ggalmazor follows up with refactoring them.
- Rework the tests in this PR

lognaturel:
Hah. I just was reading through `getAbsRef` again and saw line 115:
```java
        if (!parentRef.isAbsolute()) {
            throw new RuntimeException("XFormParser.getAbsRef: parentRef must be absolute");
        }
```
This means there is no case where `parentRef` could be relative so there's no advantage to using `parent` over `anchor` as far as I can tell.

Unless either of you sees an issue with my analysis (and PLEASE do check me!), I'd like to propose we merge this with the understanding that it's not totally complete in order to have the breadcrumbs and then Guillermo can use the tests here plus the additional relative reference tests he's described above as an opportunity to develop a nice set of test abstractions. That way we decouple the testing conversation from these targeted fixes.

ggalmazor:
> I don't think this is affected by the change here but I think it would be a good idea to have a series of tests around evaluating relative expressions in general. Or do you think there's a chance it would be affected?

You're right, it's not affected by the change in ItemsetBinding. I think we can consider adding it to a test suite around `current()` use cases if we want. 

> This analysis gives me pretty high confidence that the change from parent to anchor in getAbsRef is safe as long as we can't come up with a case where the parent/context node needs to be relative.

I can follow your comments about what the code does, but I think I need more context to really understand the difference between anchor and parent. For example: why would a user want to write `../../foo.parent(../)` instead of `../../../bar/foo`? Is there an alternative version of that with `anchor`?

Also, I'm confused about this XPath expression you've used: `../../foo.parent(../)`. Is that syntax supported? I've tried to search for the parent function but I can't find it. I see that we support the `parent` XPath Axis, as in `../../parent::foo`, which would be the same as `../../../foo` (I think).

> A narrower alternative is that we could come up with cases of pairs of references and the context to parent or anchor against and demonstrate that anchor does the same thing as parent for all cases that parent supports.

That's a really good idea, and I think it would help me to understand it better.

>  I'd like to propose we merge this with the understanding that the tests are not totally complete in order to have the breadcrumbs and then Guillermo can use the tests here plus the additional relative reference tests he's described above as an opportunity to develop a nice set of test abstractions

I still have a lot of gaps in my knowledge to cover but everything you say seems right. I'm up for working on those tests with your help to compile use cases after we merge this one.

lognaturel:
> why would a user want to write ../../foo.parent(../) instead of ../../../bar/foo? Is there an alternative version of that with anchor?

I really don't know! As I noted previously, there's no test that exercises this behavior. And presumably no one has ever hit the case because they would have run into the bug fixed at https://github.com/opendatakit/javarosa/pull/341/commits/c54e789f065f013629e3fbda6e756252cff7bf89.

There's no alternative with `anchor` because `anchor` only allows for an absolute reference to be used as the base.

> Also, I'm confused about this XPath expression you've used: ../../foo.parent(../)

Sorry, I'm just using shorthand, there. I'm still talking about the same [`parent`](https://github.com/opendatakit/javarosa/blob/60da3858b1f95b581ba55349a76a8097c4c4eaa8/src/org/javarosa/core/model/instance/TreeReference.java#L233) method. It's a member of `TreeReference` and called with another `TreeReference`. I'm basically showing the `toString` of the two references if that makes sense. So imagine a `TreeReference` that represents the `../../foo` expression. I want to call `parent` with a base of a `TreeReference` that represents the `../` expression.

> I see that we support the parent XPath Axis, as in ../../parent::foo, which would be the same as ../../../foo (I think).

Yes, agreed. But hopefully the paragraph above makes it clear what I actually meant.

> I'm up for working on those tests with your help to compile use cases after we merge this one.

👍 👍 Over to you @dcbriccetti to see if you agree and are comfortable merging this first batch of work. It may also be helpful to look at some of the cleanup and comment improvement at #341.

dcbriccetti:
>Over to you @dcbriccetti

Looking now.

lognaturel:
And just to confirm, @dcbriccetti, you were convinced by the analysis of `parent` vs. `anchor` for that substitution?

Are you satisfied with the list of additional tests we've described above or are there others you've thought of as part of your working through this?

dcbriccetti:
I followed the discussion and didn’t see any problems with it, but why don’t I look again, more closely.

dcbriccetti:
>And just to confirm, @dcbriccetti, you were convinced by the analysis of parent vs. anchor for that substitution?

I’m still struggling to understand it thoroughly, but I’ll keep monitoring your progress.

>Are you satisfied with the list of additional tests we've described above or are there others you've thought of as part of your working through this?

Some low level TreeReference tests of anchor and parent might be useful. Such as
```
@Test(expected=XPathException.class) public void anchor_base_must_be_absolute() {
    TreeReference trBase = new TreeReference();
    TreeReference tr = new TreeReference();
    tr.anchor(trBase);
}
```

PR #338, opened 2018-07-15
Remove PrototypeFactoryDeprecated

dcbriccetti:
This class is used to get values from a map, but nothing is ever put into the map. It is called in only one place, in XFormParser.buildInstanceStructure, to get a value (always null, because of the always-empty map). It is called only if `node.getAttributeValue(NAMESPACE_JAVAROSA, "modeltype")` is not null. 

Do we support this “modeltype” attribute? I don’t find it used anywhere in ODK. If we don’t, we can pull out a bit more code, and if we don’t know we can go ahead with the change in this PR.

I have satisfied myself that the code I removed is dead code, and that the same path will continue to be executed.

No.


ggalmazor:
My only concern is that any other third party could be using the `PrototypeFactoryDeprecated.addNewPrototype(String name, Class prototype)`, and this change would affect them. 

I don't know if this should be something to worry about, taking into account that the class is called PrototypeFactory**Deprecated** since April 2016 (feels like we gave plenty of time to react for anyone using)

lognaturel:
Merging now because we're about to do a major version bump. Thanks!

PR #342, opened 2018-07-18
Revert namespace to the one set in Aggregate years ago

yanokwa:
Addresses https://github.com/opendatakit/xforms-spec/issues/203

Reverts df7c725e3cc6ec44febeb425a5be166a5564186a
Reverts e441a66e35dbab6939b4a32523fc208b8da24bf1

Visual inspection.

I searched for `opendatakit.org/xforms` in the repo and replaced them. 

Yes, my visual inspection could have missed something.


lognaturel:
Looks good but holding on final ok from opendatakit/xforms-spec#203.

PR #343, opened 2018-07-18
Correct the namespace for rank in test form

lognaturel:
Namespace is not actually enforced for it so no code change is needed.

Javarosa is a little lax about enforcing namespace in general. Should we enforce it for `rank`? Currently `parseElement` just looks at name so we'd need to add a mechanism to do so I believe.


yanokwa:
Blocking on #342

PR #347, opened 2018-07-21
Improve readability and remove duplication in ItemsetBinding.initReferences; fix @link’s

dcbriccetti:
Stepping through the code in the debugger, I needed to clean up ItemsetBinding.initReferences in order to understand it. If the improvement is useful, let’s merge it.

I also found some doc comment `link`s missing the `@`.

Changes are made with safe IDEA tools like extract variable and method.

### DIFF
--- a/src/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/org/javarosa/core/model/ItemsetBinding.java
@@ -1,5 +1,6 @@
 package org.javarosa.core.model;
 
+import static org.javarosa.core.model.FormDef.getAbsRef;
 import static org.javarosa.xform.parse.RandomizeHelper.shuffle;
 
 import java.io.DataInputStream;
@@ -8,7 +9,6 @@ import java.io.IOException;
 import java.util.List;
 
 import org.javarosa.core.model.condition.IConditionExpr;
-import org.javarosa.core.model.instance.FormInstance;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.core.model.util.restorable.RestoreUtils;
 import org.javarosa.core.services.locale.Localizable;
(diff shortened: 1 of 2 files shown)


src/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -114,32 +114,39 @@ public IConditionExpr getRelativeValue () {
...
+    /** Returns a TreeReference from an absolute reference from an IConditionExpr and a parent TreeReference */
+    private static TreeReference getCondAbsTreeRef(IConditionExpr condExpr, TreeReference parentTreeRef) {
+        XPathPathExpr xPathPathExpr = getCastExpr(condExpr);
+        return getPathAbsTreeRef(new XPathReference(xPathPathExpr), parentTreeRef);
+    }
+
+    /** Returns a TreeReference from an absolute reference from an XPathReference and a parent TreeReference */
+    private static TreeReference getPathAbsTreeRef(XPathReference xPathReference, TreeReference parentTreeRef) {
+        IDataReference dataReference = getAbsRef(xPathReference, parentTreeRef);
+        return (TreeReference) dataReference.getReference();
```
dcbriccetti: I inlined FormInstance.unpackReference (which doesn’t provide any value here):
```
    public static TreeReference unpackReference(IDataReference ref) {
        return (TreeReference) ref.getReference();
    }
```

dcbriccetti:
Thanks, @lognaturel. I think I made the changes you recommended, and I added your explanatory text.

src/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -114,32 +114,34 @@ public IConditionExpr getRelativeValue () {
...
-                    nodesetRef));
-        }
-        if ( valueExpr != null ) {
-            valueRef = FormInstance.unpackReference(FormDef.getAbsRef(new XPathReference(((XPathPathExpr) ((XPathConditional) valueExpr).getExpr())),
-                    nodesetRef));
-        }
 
-        if ( q != null ) {
+        // Convert the expression to a relative reference and then anchor it to the context.
+        nodesetRef = (TreeReference) getAbsRef(new XPathReference(
```
lognaturel: You should be able to just use `getAbsoluteRef(nodesetExpr, contextRef)`, here. Did that not work?

dcbriccetti:
I didn’t completely follow the noop comment (we can move on), but I’ll trust you that it works. I made the change. Thanks.

lognaturel:
Trust is a dangerous thing!

The original expression was `new XPathReference(((XPathPathExpr) ((XPathConditional) nodesetExpr).getExpr()).getReference())`. `XPathPathExpr.getReference()` does a whole bunch of stuff to go from a path expression to a reference. Then the `XPathReference(XPathReference)` constructor was called with it and that basically makes a copy of the parameter (no-op). 

For the others, `.getReference()` wasn't called explicitly. That means the `XPathReference(XPathPathExpr)` constructor was being called. That constructor calls `.getReference()` on its parameter. That's the same thing that was happening above.

PR #350, opened 2018-07-24
Issue 348 test parent and anchor

### LINKED ISSUE #348: Add unit tests for TreeReference.anchor and TreeReference.parent
#341 added some more details to the comments for `anchor` and `parent`. It would be good to support that language with unit tests. 

An example from @dcbriccetti of verifying preconditions:
```
@Test(expected=XPathException.class) public void anchor_base_must_be_absolute() {
    TreeReference trBase = new TreeReference();
    TreeReference tr = new TreeReference();
    tr.anchor(trBase);
}
```

This will also be helpful for validating the change in 507f56de64874f51257829f6825beeffd717b3b6 where `parent` was replaced by `anchor`. For this aspect, the important thing will be to confirm that the two methods have identical behavior in cases with parameter values that have no parent steps (../).

ggalmazor:
Closes #348

This PR adds new test classes to cover `TreeReference.anchor()` and `TreeReference.parent()`, and, hopefully, give enough context to understand what they do.

This PR also adds a new test dependency with hamcrest to be able to write more expressive tests and composable custom Matchers adapted to JR's domain.

Run automated tests

There are no implementation code changes.

None whatsoever.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ dependencies {
     testCompile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'net.sf.kxml', name: 'kxml2', version: '2.3.0'
+    testCompile group: 'org.hamcrest', name: 'hamcrest-all', version: '1.3'
 }
 
 jacocoTestReport {
--- a/test/org/javarosa/core/model/instance/TestHelpers.java
+++ b/test/org/javarosa/core/model/instance/TestHelpers.java
@@ -1,0 +1,38 @@
+package org.javarosa.core.model.instance;
+
+import static org.javarosa.xpath.expr.XPathPathExpr.INIT_CONTEXT_RELATIVE;
+import static org.javarosa.xpath.expr.XPathPathExpr.INIT_CONTEXT_ROOT;
+import static org.javarosa.xpath.expr.XPathStep.ABBR_PARENT;
+import static org.javarosa.xpath.expr.XPathStep.ABBR_SELF;
+import static org.javarosa.xpath.expr.XPathStep.AXIS_CHILD;
+
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.javarosa.xpath.expr.XPathPathExpr;
+import org.javarosa.xpath.expr.XPathQName;
+import org.javarosa.xpath.expr.XPathStep;
+
+public class TestHelpers {
+    static TreeReference buildRef(String xpath) {
+        String[] parts = xpath.split("/");
+        XPathStep[] steps = Stream.of(parts)
+            .map(part -> {
+                switch (part) {
+                    case "..":
+                        return ABBR_PARENT();
+                    case "":
+                        return ABBR_SELF();
+                    default:
+                        return new XPathStep(AXIS_CHILD, new XPathQName(null, part));
+                }
+            })
+            .collect(Collectors.toList())
+            .toArray(new XPathStep[parts.length]);
+
+        XPathPathExpr xpathExpression = new XPathPathExpr(
+            xpath.startsWith("/") ? INIT_CONTEXT_ROOT : INIT_CONTEXT_RELATIVE,
+            steps
+        );
+        return xpathExpression.getReference();
+    }
+}
--- a/test/org/javarosa/core/model/instance/TestHelpers.java
+++ b/test/org/javarosa/core/model/instance/TestHelpers.java
@@ -1,38 +1,16 @@
 package org.javarosa.core.model.instance;
 
 import static org.javarosa.xpath.expr.XPathPathExpr.INIT_CONTEXT_RELATIVE;
-import static org.javarosa.xpath.expr.XPathPathExpr.INIT_CONTEXT_ROOT;
-import static org.javarosa.xpath.expr.XPathStep.ABBR_PARENT;
-import static org.javarosa.xpath.expr.XPathStep.ABBR_SELF;
-import static org.javarosa.xpath.expr.XPathStep.AXIS_CHILD;
 
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import org.javarosa.model.xform.XPathReference;
 import org.javarosa.xpath.expr.XPathPathExpr;
-import org.javarosa.xpath.expr.XPathQName;
 import org.javarosa.xpath.expr.XPathStep;
 
 public class TestHelpers {
     static TreeReference buildRef(String xpath) {
-        String[] parts = xpath.split("/");
-        XPathStep[] steps = Stream.of(parts)
-            .map(part -> {
-                switch (part) {
-                    case "..":
-                        return ABBR_PARENT();
-                    case "":
-                        return ABBR_SELF();
-                    default:
-                        return new XPathStep(AXIS_CHILD, new XPathQName(null, part));
-                }
-            })
-            .collect(Collectors.toList())
-            .toArray(new XPathStep[parts.length]);
-
-        XPathPathExpr xpathExpression = new XPathPathExpr(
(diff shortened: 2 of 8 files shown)


ggalmazor:
On Maven it's failing a `LocalizerTest.testLinearSub()`, which I haven't changed... any ideas?

lognaturel:
I think CircleCI just got resource constrained during that test for some reason. Can you please re-run, I bet that will work.

test/org/javarosa/core/model/instance/TestHelpers.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.javarosa.xpath.expr.XPathStep.AXIS_CHILD;
+
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.javarosa.xpath.expr.XPathPathExpr;
+import org.javarosa.xpath.expr.XPathQName;
+import org.javarosa.xpath.expr.XPathStep;
+
+public class TestHelpers {
+    static TreeReference buildRef(String xpath) {
```
lognaturel: I was wondering whether we might already have a way to do this and I think `XPathReference(String)` might be just that.

test/org/javarosa/core/model/instance/TreeReferenceParentTest.java
```diff
@@ -0,0 +1,52 @@
...
+    public String expectedRef;
+
+    @Parameterized.Parameters(name = "{0}")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][]{
+            {"Parenting absolute refs doesn't change them", "/foo", "/bar", "/foo"},
+            {"Parenting to an empty ref doesn't change them", "../foo", "", "../foo"},
+            {"foo.parent(bar) gives bar/foo", "foo", "bar", "bar/foo"},
+            {"foo.parent(../bar) gives ../bar/foo", "foo", "../bar", "../bar/foo"},
+            // TODO review this. Shouldn't this resolve to "bar/foo"?
```
lognaturel: Did you base these off the latest master? I think that's the bug that was fixed in another recent PR.

build.gradle
```diff
@@ -43,6 +43,7 @@ dependencies {
     testCompile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'net.sf.kxml', name: 'kxml2', version: '2.3.0'
+    testCompile group: 'org.hamcrest', name: 'hamcrest-all', version: '1.3'
```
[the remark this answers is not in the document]
ggalmazor: Junit4 includes `hamcrest-core` which is a very basic library with `assertThat` and some basic matchers only. I'm adding `hamcrest-all`, which includes all sorts of useful matchers that we would have to implement ourselves otherwise.

test/org/javarosa/core/model/instance/TestHelpers.java
ggalmazor: Ugh! I'm a dummy! That's much better, thanks!

ggalmazor: Using `new XPathReference(String nodeset)` works correctly but to cover all the paths of `parent()`, I need to produce a base xpath that would pass this check in `TreeReference.java` line 250:
```java
!baseReference.isAbsolute() && baseReference.size() == 0
```

This is consistent with an empty `""` xpath, right?

In my first implementation of `TestHelpers.buildRef(String xpath)` this xpath turned to a relative `TreeReference` with an empty list of steps, which would pass the check above.

If I do `new XPathReference("").getReference()` I get an `org.javarosa.xpath.XPathException: XPath evaluation: Parse error in XPath path: [].` error.

What now?
- I can remove the test case and leave that line not covered by any test
- We can come up with another xpath that won't fail with the `XPathPreference` class' constructor that passes the check on line 250. **I'd need some help here.**
- I can conditionally run my old `TestHelpers.buildRef(String xpath)` just for the `""` xpath case

test/org/javarosa/core/model/instance/TreeReferenceParentTest.java
ggalmazor: I've rebased the PR. Let's see if that solved this.

ggalmazor:
Thanks, @dcbriccetti!

test/org/javarosa/core/model/instance/TreeReferenceParentTest.java
lognaturel: It looks like there's still a problem here, right? Would be good to file an issue because it should not be `null`.

ggalmazor: OK, I've created #362

test/org/javarosa/core/model/instance/TestHelpers.java
ggalmazor: After discussing this with @lognaturel, we have confirmed that the interpretation of what an empty xpath should be as a `TreeReference` is correct: a relative TreeReference with an empty list of steps.

We also have agreed on conditionally running the old `TestHelpers.buildRef(String xpath)` to avoid the `XPathException` when we get an empty xpath.

This means that there's no need to add more changes regarding this.

test/org/javarosa/core/model/instance/TreeReferenceParentTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Parameterized.Parameter(value = 3)
+    public String expectedRef;
+
+    @Parameterized.Parameters(name = "{0}")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][]{
+            {"Parenting absolute refs doesn't change them", "/foo", "/bar", "/foo"},
+            {"Parenting to an empty ref doesn't change them", "../foo", "", "../foo"},
+            {"foo.parent(bar) gives bar/foo", "foo", "bar", "bar/foo"},
```
lognaturel: Hm, this kind of kills my hypothesis that parenting with a relative base that has non-parent steps (`foo.parent(../bar)`) is forbidden because the result isn't useful. This has the same shape.

I still think for now a comment in the tests that says we don't quite understand why the limitation exists is enough. Maybe it's vestigial and we can get rid of it at some point. Maybe there is a check against `null` that it's important for somewhere. We don't really need to figure that out now.

lognaturel:
Apologies for taking so long to get back to this!

For `anchor`, I would add just one case: anchoring a ref without any parent steps. That is, something like `'baz'.anchor('/foo/bar') -> '/foo/bar/baz'`

It doesn't increase coverage but it feels like a semantically different case to me and it demonstrates that the level copy loop works independently from the ref resolving loop. 

For `parent`, we discussed that the `null` is because of the limitation described in "However, if this ref has 'parent' steps (..), it can only be anchored if the base ref is a relative ref consisting only of other 'parent' steps. For example, '../../a'.parent('..') is valid and yields '../../../a'.". We can't figure out if that limitation is important but that can be investigated some other time.

The one case I see missing there is two relative references with parent steps (`'../../a'.parent('..') -> '../../../a'`). So #362 can either be closed or changed to make it about investigating whether the limitation can be lifted (which would make `parent` and `anchor` even more similar).

lognaturel:
For extra bonus points, it might be helpful if the cases that work exactly the same way for `parent` and `anchor` were highlighted in some way. Maybe just a block with a comment?

ggalmazor:
> For extra bonus points, it might be helpful if the cases that work exactly the same way for parent and anchor were highlighted in some way. Maybe just a block with a comment?

We have an extra test class just for that `TreeReferenceAnchorHarnessTest`. Do you still think it's necessary to highlight them in the other tests?

ggalmazor:
I've updated #362 with your feedback. I agree that we can use that issue to explore the limitation in `parent()`. I've also created a new PR #369 with the extra test cases.

PR #353, opened 2018-07-25
Fixed problem with area() and distance() functions.

### LINKED ISSUE #352: Problem with calculate expressions for area
#### Software versions 
JavaRosa v1.11.0

#### Problem description
I can not load a form that has the calculate function area(). An error is thrown when I try to open the form for filling.

https://forum.opendatakit.org/t/problem-with-calculate-expressions-for-area-in-v1-16-0/14219

1. Try to open the attached form.

grzesiek2010:
Closes #352 

I tested the solution with Collect and this form: 
[area_distance.xml.txt](https://github.com/opendatakit/javarosa/files/2228203/area_distance.xml.txt)

We shouldn't throw any exception if less than three points provided for area() func and less than two for distance() func since it causes problems on a form launch when none points are added but a calculation is performed. 

No.


lognaturel:
Ahh, how frustrating that we didn't think of this. https://forum.opendatakit.org/t/problem-with-calculate-expressions-for-area-in-v1-16-0/14219 is a user running into it. Probably we should have done this check in the first place and asked user to do what @grzesiek2010 suggested and check for blank values but `area` has been out for a long time so it's too late to make that kind of change.

I think the process we need to change is that all test forms should represent what a user wants to do. These test forms pre-populate values which is not representative.

PR #354, opened 2018-07-25
Issue 351 controltype enum

### LINKED ISSUE #351: Replace static members with a new ControlType enum
Control types are defined as static int members in the `Constants` class. Instead of that, we should have a `ControlType` like this:

```java
enum ControlType {
    UNTYPED(-1),
    INPUT(1),
    SELECT_ONE(2),
    SELECT_MULTI(3),
    TEXTAREA(4),
    SECRET(5),
    RANGE(6),
    UPLOAD(7),
    SUBMIT(8),
    TRIGGER(9),
    IMAGE_CHOOSE(10),
    LABEL(11),
    AUDIO_CAPTURE(12),
    VIDEO_CAPTURE(13),
    OSM_CAPTURE(14),
    FILE_CAPTURE(15);

    private int value;

    ControlType(int value) {
      this.value = value;
}
```

ggalmazor:
Closes #351

This PR extracts to an enum the static members about control type codes in `model.Constants`.

It also binds the values in that class to values from the `ControlType` and `DataType` enums to reduce duplication and the risk of accidental de-synchronization of duplicated values.

Run the automated tests.

These changes ensure there's only one definition for the integer codes of JR data types and control types.

They also make the `ControlType` available to be used by third parties like Briefcase

None that I can think of.

### DIFF
--- a/src/org/javarosa/core/model/Constants.java
+++ b/src/org/javarosa/core/model/Constants.java
@@ -46,26 +46,26 @@ public class Constants {
     /** Over The Air or HTTP Connection */
     public static final int CONNECTION_OTA = 4;
 
-    public static final int DATATYPE_UNSUPPORTED    = -1;
-    public static final int DATATYPE_NULL           = 0;
-    public static final int DATATYPE_TEXT           = 1;
-    public static final int DATATYPE_INTEGER        = 2;
-    public static final int DATATYPE_DECIMAL        = 3;
-    public static final int DATATYPE_DATE           = 4;
-    public static final int DATATYPE_TIME           = 5;
-    public static final int DATATYPE_DATE_TIME      = 6;
-    public static final int DATATYPE_CHOICE         = 7;
+    public static final int DATATYPE_UNSUPPORTED    = DataType.UNSUPPORTED.value;
+    public static final int DATATYPE_NULL           = DataType.NULL.value;
+    public static final int DATATYPE_TEXT           = DataType.TEXT.value;
+    public static final int DATATYPE_INTEGER        = DataType.INTEGER.value;
+    public static final int DATATYPE_DECIMAL        = DataType.DECIMAL.value;
+    public static final int DATATYPE_DATE           = DataType.DATE.value;
+    public static final int DATATYPE_TIME           = DataType.TIME.value;
+    public static final int DATATYPE_DATE_TIME      = DataType.DATE_TIME.value;
+    public static final int DATATYPE_CHOICE         = DataType.CHOICE.value;
     /** A list of items used for selecting multiple answers or ordering them */
-    public static final int DATATYPE_MULTIPLE_ITEMS = 8;
+    public static final int DATATYPE_MULTIPLE_ITEMS = DataType.MULTIPLE_ITEMS.value;
     /** The same as {@link Constants#DATATYPE_MULTIPLE_ITEMS} (for backwards compatibility) */
-    public static final int DATATYPE_CHOICE_LIST    = 8;
-    public static final int DATATYPE_BOOLEAN        = 9;
-    public static final int DATATYPE_GEOPOINT       = 10;
-    public static final int DATATYPE_BARCODE        = 11;
-    public static final int DATATYPE_BINARY         = 12;
-    public static final int DATATYPE_LONG           = 13;
-    public static final int DATATYPE_GEOSHAPE       = 14;
-    public static final int DATATYPE_GEOTRACE       = 15;
+    public static final int DATATYPE_CHOICE_LIST    = DataType.MULTIPLE_ITEMS.value;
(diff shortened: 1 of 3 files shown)


src/org/javarosa/core/model/util/restorable/RestoreUtils.java
```diff
@@ -97,27 +97,39 @@ public static void addData (FormInstance dm, String xpath, Object data, int data
...
+        IAnswerData val = buildAnswerData(data, dataType);
 
         TreeReference ref = absRef(xpath, dm);
         if (dm.addNode(ref, val, dataType) == null) {
             throw new RuntimeException("error setting value during object backup [" + xpath + "]");
         }
     }
 
+    private static IAnswerData buildAnswerData(Object data, int dataType) {
+        // Can't use a switch because these static members aren't considered constant expressions anymore
```
dcbriccetti: See AnswerDataFactory.templateByDataType for how you can/should use switch.

src/org/javarosa/core/model/Constants.java
```diff
@@ -46,45 +46,45 @@
...
-    public static final int DATATYPE_UNSUPPORTED    = -1;
-    public static final int DATATYPE_NULL           = 0;
-    public static final int DATATYPE_TEXT           = 1;
-    public static final int DATATYPE_INTEGER        = 2;
-    public static final int DATATYPE_DECIMAL        = 3;
-    public static final int DATATYPE_DATE           = 4;
-    public static final int DATATYPE_TIME           = 5;
-    public static final int DATATYPE_DATE_TIME      = 6;
-    public static final int DATATYPE_CHOICE         = 7;
+    public static final int DATATYPE_UNSUPPORTED    = DataType.UNSUPPORTED.value;
```
dcbriccetti: This is a nice improvement.

src/org/javarosa/core/model/util/restorable/RestoreUtils.java
ggalmazor: Good one, thanks!

src/org/javarosa/core/model/util/restorable/RestoreUtils.java
```diff
@@ -98,17 +100,17 @@ public static void addData (FormInstance dm, String xpath, Object data, int data
...
-        case Constants.DATATYPE_INTEGER: val = new IntegerData((Integer)data); break;
-        case Constants.DATATYPE_LONG: val = new LongData((Long)data); break;
-        case Constants.DATATYPE_DECIMAL: val = new DecimalData((Double)data); break;
-        case Constants.DATATYPE_BOOLEAN: val = new StringData(((Boolean)data).booleanValue() ? "t" : "f"); break;
-        case Constants.DATATYPE_DATE: val = new DateData((Date)data); break;
-        case Constants.DATATYPE_DATE_TIME: val = new DateTimeData((Date)data); break;
-        case Constants.DATATYPE_TIME: val = new TimeData((Date)data); break;
-        case Constants.DATATYPE_MULTIPLE_ITEMS: val = (MultipleItemsData)data; break;
+        switch (from(dataType)) {
+        case UNSUPPORTED: val = null; break;
```
dcbriccetti: Nice and clean!

lognaturel:
Since this breaks downstream tools (see #363), I think we should probably revert it or add the prior interface back in alongside this one. We should make sure that changes to public interfaces are always documented as risks!

I tried to upgrade Briefcase to jr 2.11.2 and that broke for related reasons so I think there was another change that we didn't consider compatibility implications for.

ggalmazor:
Good catch @grzesiek2010!

I've created a PR fixing this and I'll create issues on Collect and Briefcase to migrate code to the enum values.

ggalmazor:
> I tried to upgrade Briefcase to jr 2.11.2 and that broke for related reasons so I think there was another change that we didn't consider compatibility implications for.

In this case, Briefcase fails to build because the value `DataType.CHOICE_LIST` was removed and replaced by `DataType.MULTIPLE_ITEMS`. I think this is correct because according to comments in `model.Constants`, `CHOICE_LIST` is there just for backward compatibility. I think that `DataType` should have only currently valid values.

We can also take advantage of the factory `DataType.from(int)` which will normalize  the int value `8` to `DataType.MULTIPLE_ITEMS` and will help us effectively deprecate the `CHOICE_LIST` concept from the code.

dcbriccetti:
I’m working on ODK today. Can I help, @ggalmazor?

lognaturel:
> In this case, Briefcase fails to build because the value DataType.CHOICE_LIST was removed and replaced by DataType.MULTIPLE_ITEMS

Ahh, yes, that makes sense. I probably should have actually read the error message... 😄 (sorry!)

PR #361, opened 2018-08-09
Issue 349 augment current tests

### LINKED ISSUE #349: Improve and augment tests around relative paths, especially in choice filters
#335 introduced some tests on using `current()` in an itemset context. These are pretty low-level and not entirely targeted (e.g. they always do a "round trip" with verification that values are properly written back out).

On that PR, @ggalmazor discussed a few directions that could be taken to make these tests easier to read. For example, a method `answerQuestion(string, string)` could be introduced. The first string would be interpreted as an XPath path that a reference can be built from. Then this could be converted to a formindex to jump to that desired question. The second string would be the answer to set.

Similarly, `answerOf(string)` could take an XPath path as a string and return the answer to that question / value of the node.

To augment these tests, introduce more cases with relative paths (from https://github.com/opendatakit/javarosa/pull/335#issuecomment-405188083):
- `submission` inside the model of a form  [XFormParser.java#755](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L755)
- `setvalue` action [XFormParser.java#690](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L690)
- A form field with a binding with a relative path [XFormParser.java#969](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L969)
  - Specifically, a group with a binding with a relative path [XFormParser.java#1468](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L1468) and [XFormParser.java#1477](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/xform/parse/XFormParser.java#L1477)
- A `<label>` that shows an `<itext>` related to the selected value from a previous field

ggalmazor:
Closes #349

~This PR is based on #350. Commit d07de54e is the first commit after that.~ (PR rebased to master)

This PR adds new cases to the test suite around the changes done for `current()`, `parent()`, and `anchor()`.

This PR improves the `Scenario` class for testing purposes which lets test JavaRosa with a high-level testing API. This API offers methods to statically declare the state of a form or to dynamically drive the form filling like a real user would do.

Tests developed with the `Scenario` class should give easy to understand example use cases by describing the state of the form or by sequencing user actions, and by defining high level result expectations.

Run the tests

These changes have been worked on top of other lower level test, which are still relevant.

No, this is just test code and maybe some harmless visibility changes and new utility methods in the implementation codebase.

### DIFF
--- a/src/org/javarosa/core/model/instance/TreeReference.java
+++ b/src/org/javarosa/core/model/instance/TreeReference.java
@@ -474,6 +474,10 @@ public class TreeReference implements Externalizable, Serializable {
     }
 
     public String toString (boolean includePredicates) {
+        return toString(includePredicates, false);
+    }
+
+    public String toString (boolean includePredicates, boolean zeroIndexMult) {
         StringBuilder sb = new StringBuilder();
         if(instanceName != null) {
             sb.append("instance("+instanceName+")");
@@ -499,13 +503,14 @@ public class TreeReference implements Externalizable, Serializable {
 
             if (includePredicates) {
                 switch (mult) {
-                case INDEX_UNBOUND: break;
-                case INDEX_TEMPLATE: sb.append("[@template]"); break;
-                case INDEX_REPEAT_JUNCTURE: sb.append("[@juncture]"); break;
-                default:
-                    if ((i > 0 || mult != 0) && mult !=-4)
-                        sb.append("[" + (mult + 1) + "]");
-                    break;
+                    case INDEX_UNBOUND: break;
+                    case INDEX_TEMPLATE: sb.append("[@template]"); break;
+                    case INDEX_REPEAT_JUNCTURE: sb.append("[@juncture]"); break;
+                    default:
+                        if ((i > 0 || mult != 0) && mult !=-4) {
+                            sb.append("[").append(mult + (zeroIndexMult ? 0 : 1)).append("]");
+                        }
+                        break;
                 }
             }
 
--- a/test/org/javarosa/xpath/expr/Scenario.java
+++ b/test/org/javarosa/xpath/expr/Scenario.java
@@ -37,10 +37,48 @@ class Scenario {
         return new Scenario(formDef, formEntryController);
     }
 
-    Scenario answer(String xpath, String value) {
-        FormIndex index = getIndexOf(xpath);
-        formEntryController.answerQuestion(index, new StringData(value), true);
-        return this;
+    void answer(String xpath, String value) {
+        createMissingRepeats(xpath);
+        TreeElement resolve = resolve(xpath);
+        formDef.setValue(new StringData(value), resolve.getRef(), true);
+    }
+
+    @SuppressWarnings("unchecked")
+    <T extends IAnswerData> T answerOf(String xpath) {
+        TreeElement root = (TreeElement) formDef.getMainInstance().getRoot().getParent();
+        String relativeXPath = xpath.startsWith("/") ? xpath.substring(1) : xpath;
+        TreeElement resolve = resolve(root, relativeXPath);
+        return resolve != null ? (T) resolve.getValue() : null;
+    }
+
+    List<SelectChoice> choicesOf(String xpath) {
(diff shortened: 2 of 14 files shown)


test/org/javarosa/xpath/expr/Scenario.java
```diff
@@ -0,0 +1,490 @@
...
+            Optional<TreeElement> result = getFirstDescendantWithName(child, name);
+            if (result.isPresent())
+                return result;
+        }
+
+        return Optional.empty();
+    }
+
+    private List<TreeElement> childrenOf(TreeElement node) {
+        List<TreeElement> children = new ArrayList<>();
```
dcbriccetti: It probably doesn’t matter for these tests, but it might set a better example if you gave an `initialCapacity`, since you know how many elements there will be.

test/org/javarosa/xpath/expr/Scenario.java
```diff
@@ -0,0 +1,490 @@
...
+     * This method is part of a set of methods that address form navigation in a dynamic way,
+     * imitating real user interaction with the form:
+     * <ul>
+     * <li>{@link #jumpToFirst(String)}
+     * <li>{@link #answer(String)}
+     * <li>{@link #next()}
+     * <li>{@link #next(String)}
+     * <li>{@link #atTheEndOfForm()}
+     * </ul>
+     */
```
dcbriccetti: This documentation is really nice, but I’d prefer to avoid the duplication. Maybe just have those 9 lines in one place? An ordinary comment above the group of methods, perhaps?

src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -499,13 +503,14 @@ public String toString (boolean includePredicates) {
...
-                default:
-                    if ((i > 0 || mult != 0) && mult !=-4)
-                        sb.append("[" + (mult + 1) + "]");
-                    break;
+                    case INDEX_UNBOUND: break;
+                    case INDEX_TEMPLATE: sb.append("[@template]"); break;
+                    case INDEX_REPEAT_JUNCTURE: sb.append("[@juncture]"); break;
+                    default:
+                        if ((i > 0 || mult != 0) && mult !=-4) {
+                            sb.append("[").append(zeroIndexMult ? mult : mult + 1).append("]");
```
dcbriccetti: It’s fine how it is, but in case you didn’t think of it and prefer:

append(mult + (zeroIndexMult ? 0 : 1))

ggalmazor:
Thanks, @dcbriccetti! I've made the suggested changes.

test/org/javarosa/xpath/expr/XPathPathExprCurrentTest.java
```diff
@@ -15,86 +15,49 @@
...
+        // The binding of /data/my_group/name_relative is:
+        //   <bind calculate="current()/../name" nodeset="/data/my_group/name_relative" type="string"/>
+        // That will copy the value of our previous answer to /data/my_group/name
+        assertThat(
+            scenario.answerOf("/data/my_group/name_relative"),
+            is(answer(scenario.answerOf("/data/my_group/name")))
+        );
     }
 
-    /**
```
lognaturel: @ggalmazor Can you say more about why you decided to remove these comments? I don't think the reasoning for why these tests exist is so clear without them.

ggalmazor: I was trying to replace the abstract description of how that works with a good enough concrete example & assertion combination. There's also a comment explaining how the itemset is defined in the form, above the assertion.

Since you've asked, I feel like this is not enough :) If you want, I can bring back those comments.

Anyway, after reviewing the test, I think the method name is not very good... It should  say `current_as_itemset_choice_filter_root_should_refer_to_the_select_node()`. The previous test method in the same test class could be renamed to `current_as_calculate_root_should_refer_to_its_bound_nodeset()` as well.

I think you were great at explaining the behavior of `current`. Maybe we could also add them where the `current` is actually processed at `XPathPathExpr:129`

PR #366, opened 2018-08-25
Issue 363 fix breaking changes

### LINKED ISSUE #363: Next release will break ODK Collect and probably other tools
#### Problem description
It's because of https://github.com/opendatakit/javarosa/pull/354
We can no longer use those edited values as constants in switch statements:

screenshot from 2018-08-23 23-03-27

ggalmazor:
Closes #363

Run automated tests.

This solves code in users of JR that depend on values from the `model.Constants` class. They often use `switch` statements against `int` static members of that class.

No. This changes public API members but their effective value is the same.

### DIFF
--- a/src/org/javarosa/core/model/Constants.java
+++ b/src/org/javarosa/core/model/Constants.java
@@ -46,45 +46,45 @@ public class Constants {
     /** Over The Air or HTTP Connection */
     public static final int CONNECTION_OTA = 4;
 
-    public static final int DATATYPE_UNSUPPORTED    = DataType.UNSUPPORTED.value;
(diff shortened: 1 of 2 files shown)


dcbriccetti:
Checking this now. Won’t be long.

ggalmazor:
Thanks, @dcbriccetti!

lognaturel:
I'm afraid I am deeply confused. Why wasn't it ok to use the values from the `DataType` enum? Seems to me that they would be identical to the literal values this PR goes back to using.

dcbriccetti:
>confused

Join the club (G and I). They are identical but they can’t be determined at *compile* time, which is what switch requires. I’ll see if I can find the section in the JLS that explains this.

PR #368, opened 2018-08-27
Issue 349 Enhance tests about current()

ggalmazor:
Addresses #349 and https://github.com/opendatakit/javarosa/pull/361#pullrequestreview-147894956

This PR improves test and implementation files with the following changes:
- Add a comment block in the implementation file
- Add comments and further examples in the test file
- Improve test method names to better capture the testing context

Run automated tests

This PR adds useful documentation and context only

Nope.


lognaturel:
🙏

PR #372, opened 2018-09-17
In FormDef, fix comments, and let IDEA “Replace with <>”.

dcbriccetti:
Tests still pass. I changed no code other than comments, and what IDEA did to “replace with <>”.


ggalmazor:
Should we merge this one directly? They seem to be pretty safe changes.

PR #375, opened 2018-10-16
Documentation comments and gitignore update

joeflack4:
#### Change summary
1. Updated gitignore to ignore common OSX file.
2. Some documentation comment updates.

This is a really small PR.

#### Rationale
**1. Updated gitignore to ignore common OSX file.**  
I wasn't sure why `.DS_Store` was missing. Maybe there is a reason?

**2. Some documentation comment updates.**
I could have probably edited a lot more documentation comments (in Python they're called docstrings; not sure what they're typically called in Java; docco?). I probably will do more in the future.

n/a

n/a

n/a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ secrets.xml
 # logging
 logback.xml
 logback-test.xml
+
+# macOS
+.DS_Store
--- a/src/org/javarosa/core/model/FormDef.java
+++ b/src/org/javarosa/core/model/FormDef.java
@@ -973,15 +973,16 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
     }
 
     /**
-     * Identifies the itemset in the backend model, and creates a set of
-     * SelectChoice objects at the current question reference based on the data
+     * Identifies the <code>itemset</code> in the backend model, and creates a set of
+     * <code>SelectChoice</code> objects at the current question reference based on the data
      * in the model.
      * <p/>
-     * Will modify the itemset binding to contain the relevant choices
+     * Modifies the <code>itemset</code> object passed by clearing any choices currently on it, and populating with
+     * choices based on the question reference <code>curQRef</code> that was passed.
      *
-     * @param itemset The binding for an itemset, where the choices will be populated
-     * @param curQRef A reference to the current question's element, which will be
-     *                used to determine the values to be chosen from.
+     * @param itemset the binding for an itemset, where the choices will be populated
+     * @param curQRef a reference to the current question's element, which will be
+     *                used to determine the values to be chosen from
      */
     public void populateDynamicChoices(ItemsetBinding itemset, TreeReference curQRef) {
         getEventNotifier().publishEvent(new Event("Dynamic choices", new EvaluationResult(curQRef, null)));
@@ -1117,7 +1118,7 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
     /**
      * Iterates over the form's data bindings, and evaluates all post-processing calls.
      *
-     * @return true if the instance was modified in any way. false otherwise.
+     * @return true if the instance was modified in any way, otherwise false
      */
     private boolean postProcessInstance(TreeElement node) {
         // we might have issues with ordering, for example, a handler that writes
@@ -1163,7 +1164,7 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
      * Requires that the instance has been set to a prototype of the instance
      * that should be used for deserialization.
      *
-     * @param dis - the stream to read from.
+     * @param dis - the stream to read from
      * @throws IOException
      * @throws InstantiationException
      * @throws IllegalAccessException
@@ -1237,7 +1238,7 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
     /**
      * Writes the form definition object to the supplied stream.
      *
-     * @param dos - the stream to write to.
+     * @param dos - the stream to write to
      * @throws IOException
      */
     @Override
@@ -1525,6 +1526,17 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
         }
     }
 
+    /**
+    * Traverses recursively the given {@link IFormElement} node tree and returns
+    * the first {@link QuestionDef} that matches the binding defined in the
+    * given {@link TreeReference}.
+    *
+    * @param ref reference to a form binding
+    * @param fe  the recursive {@link IFormElement}. The starting call would
+    *            typically receive a {@link FormDef}
+    * @return the question definition, {@code null} if no {@link QuestionDef}
+    *         matches the given binding
+    */
     public static QuestionDef findQuestionByRef(TreeReference ref, IFormElement fe) {
         if (fe instanceof FormDef) {
             ref = ref.genericize();
--- a/src/org/javarosa/core/model/instance/TreeElement.java
+++ b/src/org/javarosa/core/model/instance/TreeElement.java
@@ -52,19 +52,18 @@ import org.javarosa.xpath.expr.XPathPathExpr;
 import org.javarosa.xpath.expr.XPathStringLiteral;
 
 /**
- * An element of a FormInstance.
+ * <p>An element of a FormInstance.</p>
  *
- * TreeElements represent an XML node in the instance. It may either have a value (e.g., <name>Drew</name>),
- * a number of TreeElement children (e.g., <meta><device /><timestamp /><user_id /></meta>), or neither (e.g.,
- * <empty_node />)
+ * <p>TreeElements represent an XML node in the instance. It may either have a value (e.g., <name>Drew</name>),
+ * a number of TreeElement children (e.g., <meta><device /><timestamp /><user_id /></meta>), or neither (e.g.,<empty_node />)</p>
  *
- * TreeElements can also represent attributes. Attributes are unique from normal elements in that they are
- * not "children" of their parent, and are always leaf nodes: IE cannot have children.
+ * <p>TreeElements can also represent attributes. Attributes are unique from normal elements in that they are
+ * not "children" of their parent, and are always leaf nodes: IE cannot have children.</p>
  *
- * TODO: Split out the bind-able session data from this class and leave only the mandatory values to speed up
- * new DOM-like models
- *  * @author Clayton Sims
- *
+ * <p>TODO: Split out the bind-able session data from this class and leave only the mandatory values to speed up
+ * new DOM-like models</p>
+ 
+ * @author Clayton Sims
  */
 
  public class TreeElement implements Externalizable, AbstractTreeElement<TreeElement> {
@@ -144,11 +143,11 @@ import org.javarosa.xpath.expr.XPathStringLiteral;
     }
 
     /**
-     * Retrieves the TreeElement representing the attribute for
-     * the provided namespace and name, or null if none exists.
+     * <p>Retrieves the TreeElement representing the attribute for
+     * the provided namespace and name, or null if none exists.</p>
      *
-     * If 'null' is provided for the namespace, it will match the first
-     * attribute with the matching name.
+     * <p>If 'null' is provided for the namespace, it will match the first
+     * attribute with the matching name.</p>
      *
      * @param attributes - list of attributes to search
      * @param namespace
@@ -423,11 +422,11 @@ import org.javarosa.xpath.expr.XPathStringLiteral;
     }
 
     /**
-     * Retrieves the TreeElement representing an arbitrary bind attribute
-     * for this element at the provided namespace and name, or null if none exists.
+     * <p>Retrieves the TreeElement representing an arbitrary bind attribute
+     * for this element at the provided namespace and name, or null if none exists.</p>
      *
-     * If 'null' is provided for the namespace, it will match the first
-     * attribute with the matching name.
+     * <p>If 'null' is provided for the namespace, it will match the first
+     * attribute with the matching name.</p>
      *
      * @return TreeElement
      */
@@ -435,11 +434,19 @@ import org.javarosa.xpath.expr.XPathStringLiteral;
         return getAttribute(bindAttributes, namespace, name);
     }
 
-    /**
-     * get value of the bind attribute with namespace:name' in the list
-     *
-     * @return String
-     */
+   /**
+    * Returns the value of the given bind attribute from the given namespace.
+    * <p>
+    * Example:
+    * <ul>
+    * <li>Given a binding {@code <bind assertion="blah"/>}, calling {@code getBindAttributeValue(null, "assertion")} will return {@code "blah"}
+    * <li>Given a binding {@code <bind odk:assertion="blah"/>}, calling {@code getBindAttributeValue("odk", "assertion")} will return {@code "blah"}
+    * </ul>
+    * </p>
+    *
+    * @param namespace the namespace of the attribute. It can be null
+    * @param name      the name of the attribute
+    */
     public String getBindAttributeValue(String namespace, String name) {
         TreeElement element = getBindAttribute(namespace,name);
         return element == null ? null: getAttributeValue(element);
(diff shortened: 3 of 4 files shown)


joeflack4:
Hmm. Wondering how this failed CI? I only changed .gitignore and some comments.

src/org/javarosa/core/model/FormDef.java
```diff
@@ -982,6 +982,9 @@ public String fillTemplateString(String template, TreeReference contextRef,
      * @param itemset The binding for an itemset, where the choices will be populated
      * @param curQRef A reference to the current question's element, which will be
      *                used to determine the values to be chosen from.
+     *
+     * Side effects: Mutates the itemset object passed by clearing any choices currently on it, and populating with
+     * choices based on the question reference curQRef that was passed.
```
dcbriccetti: Hi!

1. Since the return type is void, by definition this method does its work via side effects. So, no need to mention side effects.
2. I don’t think a note like this should _follow_ the @param section.
3. Above, we already have “Will modify the itemset ...”. Maybe you just want to improve that.

.gitignore
```diff
@@ -17,3 +17,6 @@ secrets.xml
 # logging
 logback.xml
 logback-test.xml
+
+# OS X
```
dcbriccetti: I think we call it `macOS` now.
https://www.apple.com/macos/mojave/

src/org/javarosa/core/model/FormDef.java
```diff
@@ -1525,6 +1528,12 @@ private void attachControlsToInstanceData(TreeElement node) {
         }
     }
 
+    /**
+     * Returns the given question definition related to a form node reference.
```
dcbriccetti: Using the word “given” here seems wrong. “given” suggests something given by the caller in method arguments.

src/org/javarosa/core/model/FormDef.java
```diff
@@ -1525,6 +1528,12 @@ private void attachControlsToInstanceData(TreeElement node) {
...
+     * @param ref Reference to a form node.
+     * @param fe IFormElement, typically a FormDef, often 'this'.
```
dcbriccetti: “often 'this'”? What is “this”?

src/org/javarosa/core/model/FormDef.java
```diff
@@ -1525,6 +1528,12 @@ private void attachControlsToInstanceData(TreeElement node) {
         }
     }
 
+    /**
+     * Returns the given question definition related to a form node reference.
+     * @param ref Reference to a form node.
```
dcbriccetti: Looking here, https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html, I see that `@param` descriptions start with lower case and don’t have periods if there’s just one phrase.

src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -55,8 +55,7 @@
  * An element of a FormInstance.
  *
  * TreeElements represent an XML node in the instance. It may either have a value (e.g., <name>Drew</name>),
- * a number of TreeElement children (e.g., <meta><device /><timestamp /><user_id /></meta>), or neither (e.g.,
- * <empty_node />)
+ * a number of TreeElement (e.g., <meta><device /><timestamp /><user_id /></meta>), or neither (e.g.,<empty_node />)
```
dcbriccetti: “a number of” should be followed by a plural noun.

src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -438,6 +437,10 @@ public TreeElement getBindAttribute(String namespace, String name) {
     /**
      * get value of the bind attribute with namespace:name' in the list
      *
+     * Example
```
dcbriccetti: Better to add `@param`s and say that namespace may be null if there is no namespace.

src/org/javarosa/form/api/FormEntryController.java
```diff
@@ -72,12 +72,13 @@ public int answerQuestion(IAnswerData data, boolean midSurvey) {
...
+     * Attempts to save the answer at the specified FormIndex into the datamodel.
      *
-     * @param index
-     * @param data
-     * @return OK if save was successful, error if a constraint was violated.
+     * @param index The current question/prompt that is being evaluated.
+     * @param data The data to attempt to answer the question with.
+     * @return int value of ANSWER_REQUIRED_BUT_EMPTY || ANSWER_CONSTRAINT_VIOLATED || ANSWER_OK
+     *
+     * Side effects: Mutates 'index' by setting its value property to the value of 'data'.
```
dcbriccetti: Worth noting, but should be above the `@param`s, yes?

src/org/javarosa/core/model/FormDef.java
joeflack4: @dcbriccetti Thanks for the response!

Regarding (1): 

**dcbriccetti wrote:**
>1. Since the return type is void, by definition this method does its work via side effects. So, no need to mention side effects.

Regarding this, what of Josh Bloch's recommendation regarding documentation [at this point in this video](https://youtu.be/heh4OeB9A-c?t=2036) (not sure where in the vid the specific example is)?

**Josh Bloch wrote:**
>  ...and any side effects. Those are particularly important. If you don't document side effects, people will shoot themselves in the foot. I'll give you an example of this later

2. I agree.

3. That makes sense to me as well; I'll do that. I didn't notice it.

Regarding "`Side effects: `", I tried putting that in the documentation at work, and people didn't like it either. For me, I try to write as many pure functions as I can, and I like it because it's explicit (even though "`void`" is also explicit). It's probably better to lean towards this in looser languages, like Python, which don't declare a void function.

src/org/javarosa/core/model/FormDef.java
joeflack4: Agree w/ that. Let me edit it.

src/org/javarosa/core/model/FormDef.java
joeflack4: Thanks! Will update

joeflack4:
@dcbriccetti Alrighty! I think I got everything. I responded to all your comments and marked resolved.

Still wondering why this is failing CI.

Wondering what your comment is on https://youtu.be/heh4OeB9A-c?t=2032 =P

Thanks for your helpful comments. Let me know if you want me to tweak anything else or squash these commits I made for you.

Kinda wish I took the time to update a lot more of these documentation strings while I was at it. Also, what's the official term for "documentation string" in Java? Docstring, docco, javadoc?

src/org/javarosa/form/api/FormEntryController.java
```diff
@@ -72,12 +72,13 @@ public int answerQuestion(IAnswerData data, boolean midSurvey) {
...
 
     /**
-     * Attempts to save the answer at the specified FormIndex into the
-     * datamodel.
+     * Attempts to save the answer at the specified FormIndex into the datamodel.
      *
-     * @param index
-     * @param data
-     * @return OK if save was successful, error if a constraint was violated.
+     * Side effects: Mutates `index` by setting its value property to the value of `data`.
```
lognaturel: This is not quite right -- `index` is not mutated. You could say that it sets the answer on the TreeElement corresponding to the provided index, perhaps.

joeflack4: Updated! Thanks for pointing out both of those corrections.

src/org/javarosa/core/model/FormDef.java
```diff
@@ -973,15 +973,16 @@ public String fillTemplateString(String template, TreeReference contextRef,
     }
 
     /**
-     * Identifies the itemset in the backend model, and creates a set of
-     * SelectChoice objects at the current question reference based on the data
+     * Identifies the `itemset` in the backend model, and creates a set of
```
lognaturel: Markdown here and in the rest of the comment!

src/org/javarosa/core/model/FormDef.java
```diff
@@ -1525,6 +1526,12 @@ private void attachControlsToInstanceData(TreeElement node) {
         }
     }
 
+    /**
+     * Return the matching question definition related to a form node reference `ref`.
```
lognaturel: Markdown here!

src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -438,6 +437,11 @@ public TreeElement getBindAttribute(String namespace, String name) {
...
+     *   Accessing a bind simply called "assertion" (has no namespace); can be called as follows:
+     *   `getBindAttributeValue(null, "assertion");`
```
lognaturel: Markdown

src/org/javarosa/core/model/FormDef.java
joeflack4: @lognaturel Eep! Let me take care of these.

src/org/javarosa/core/model/FormDef.java
joeflack4: Fixed!

src/org/javarosa/core/model/instance/TreeElement.java
joeflack4: Fixed!

src/org/javarosa/core/model/FormDef.java
joeflack4: I believe I have taken care of the issues pointed out. Marking resolved!

.gitignore
joeflack4: Corrected. Marking this one resolved too.

src/org/javarosa/core/model/FormDef.java
joeflack4: Corrected this one too. Marking resolved.

src/org/javarosa/core/model/instance/TreeElement.java
joeflack4: Corrected this one too. Marking resolved.

src/org/javarosa/core/model/FormDef.java
joeflack4: Corrected this one too. Marking resolved.

src/org/javarosa/core/model/FormDef.java
joeflack4: Corrected this one too. Marking resolved.

joeflack4:
@dcbriccetti @lognaturel Thanks to both of you to taking the time to look over this. I left the "resolve conversation" buttons open to you guys, but given that they were still open after some time, I marked them all resolved after having fixed.

If you guys want me to make a new PR to squash commits or make things a bit clearer since this one is full of comments, let me know and I can do that.

ggalmazor:
It seems like all @dcbriccetti's concerns have been addressed. Can we approve this PR and merge?

src/org/javarosa/form/api/FormEntryController.java
```diff
@@ -72,12 +72,13 @@ public int answerQuestion(IAnswerData data, boolean midSurvey) {
 
 
     /**
-     * Attempts to save the answer at the specified FormIndex into the
-     * datamodel.
+     * Attempts to save the answer at the specified FormIndex into the datamodel.
```
ggalmazor: I'm missing information about the return values here. The explanation on the `@param` can be missed out and we could give a bit more information.

It doesn't explain when the side effects happen (they don't always take place), and there are another two undocumented side effects (when the question is complex, and the `RuntimeException`) that should be included:
```java
/**
 * Attempts to save the answer at the given {@link FormIndex} into the instance
 * and returns one of three possible {@code int} attempt result codes:
 * <ul>
 * <li>{@link #ANSWER_OK}
 * <li>{@link #ANSWER_REQUIRED_BUT_EMPTY}
 * <li>{@link #ANSWER_CONSTRAINT_VIOLATED}
 * </ul>
 * <p>
 * Side effects: When it returns {@link #ANSWER_OK}, it mutates
 * the {@link TreeElement} corresponding to the given {@link FormIndex} by
 * setting its value to the given {@link IAnswerData} or by copying an
 * itemset answer if the question is complex.
 *
 * @param index The index of the question/prompt that is being currently evaluated
 * @param data  The data to attempt to answer the question with.
 * @return the attempt's {@code int} result code
 * @throws RuntimeException when the question is complex and it has constraints.
 *                          See inline comments.
 * @see QuestionDef#isComplex()
 */
```
(I've also changed `dataModel` with `instance`, which is more consistently used to refer to the "tree of answers filled by the user")

src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -438,6 +437,11 @@ public TreeElement getBindAttribute(String namespace, String name) {
     /**
      * get value of the bind attribute with namespace:name' in the list
      *
```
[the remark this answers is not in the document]
joeflack4: I honestly didn't know this was needed for paragraph spacing in Javadocs. Sorry! Fixed for that method, as well as several other locations in that file. I have a feeling that this is also prevalent elsewhere in the codebase.

src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -438,6 +437,11 @@ public TreeElement getBindAttribute(String namespace, String name) {
...
+     * Example
+     *   Accessing a bind simply called "assertion" (has no namespace); can be called as follows:
+     *   <code>getBindAttributeValue(null, "assertion");</code>
+     * @param namespace a namespace string if provided, else null
+     * @param name the name of an attribute of a <code><bind/></code> node, e.g. "myAttr" if the bind node were <code><bind myAttr="myVal"/></code>
      * @return String
```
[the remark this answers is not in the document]
joeflack4: For me when I was using JavaRosa as a library in XFormTest, this actually would have been helpful to me. For me, `@return String` and the line on `@param namespace` are redundant. But if I already have a Javadoc with a leading paragraph (and maybe also the `@param name`, I think it would probably be bad styling to leave out the other params / returns?

If you have a specific suggestion for how to rephrase, could you write it here and I will update with a commit?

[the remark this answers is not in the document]
joeflack4: Never mind, I see you did that in the next comment.

src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -438,6 +437,11 @@ public TreeElement getBindAttribute(String namespace, String name) {
     /**
      * get value of the bind attribute with namespace:name' in the list
      *
+     * Example
+     *   Accessing a bind simply called "assertion" (has no namespace); can be called as follows:
+     *   <code>getBindAttributeValue(null, "assertion");</code>
+     * @param namespace a namespace string if provided, else null
+     * @param name the name of an attribute of a <code><bind/></code> node, e.g. "myAttr" if the bind node were <code><bind myAttr="myVal"/></code>
```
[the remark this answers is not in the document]
joeflack4: Had no idea lol >,< We need a linter

joeflack4:
@dcbriccetti @yanokwa @ggalmazor @lognaturel Is there a Javadoc linter we could use? I think that would make this process much easier ;_;

src/org/javarosa/core/model/instance/TreeElement.java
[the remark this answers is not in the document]
joeflack4: This is a great help though, by the way. I'm adding this now.

src/org/javarosa/form/api/FormEntryController.java
joeflack4: This is way better. Thanks particularly for your attention to side effects. I have updated in a new commit.

ggalmazor:
Thanks for considering the changes, @joeflack4!

Regarding a linter, I use IntelliJ's dynamic inspections. They highlight on red/yellow stuff that should/could be changed as I go through the code.

ggalmazor:
Apart from the `.DS_Store` I think the changes are OK :)

You can squash the changes as you've suggested and ensure there's no `.DS_Store` or you can rebase that commit + force push, and then I can squash merge everything into master.

joeflack4:
@ggalmazor Thanks for the update. Squashing will come soon.

Back on the subject of linter, I mean more of like a "documentation style" tool. In PMA2020 projects, we use 3 tools for our Python projects: pylint for linting, and pycodestyle and pydocstyle for more style related tools. We add version specific "dev dependencies" to our projects to make sure that all of our contributors are following the same guidelines. If we disagree with a particular rule, we discuss it and then we edit the shared style config, and commit that to the repo.

joeflack4:
@dcbriccetti @yanokwa @ggalmazor @lognaturel Squashed!

ggalmazor:
Oh, we're currently using [checkstyle](https://github.com/checkstyle/checkstyle) for that. I guess we could discuss new rules to address doc block issues. Once that's done, I'd suggest we add an IntelliJ code style definition file to make it easier for people to reformat code automatically accordingly.

Would you be interested in researching new checkstyle rules that would be useful to add to the project? The rules we're currently using are defined in the `config/checkstyle/checkstyle.xml` path and the Gradle tasks that run the linter are `checkstyleMain` and `checkstyleTest`.

PR #379, opened 2018-10-25
Make uid preload generate an RFC 4122 UUID

### LINKED ISSUE #378: Match uid preload behavior to uuid() function (RFC 4122 UUID with uuid: prefix)
See discussion at https://forum.opendatakit.org/t/match-the-uid-preload-structure-to-the-uuid-function/15908

There is a desire to change what form builders add to blank forms to generate an `instanceID` so it looks like `<bind nodeset="/readonly/meta/instanceID" jr:preload="uid" type="string"/>`

See XLSForm issue at https://github.com/XLSForm/pyxform/issues/94

This would mean a change in the `instanceID` format for JavaRosa/Collect users. As far as I know, the `uid` preload is not likely to be in much use now so changing the format should be safe.

lognaturel:
Changed in preparation for form builders to use a preload instead of call on uuid() to generate instanceID.

Closes #378

Built a jar, put it in Collect and verified that [uid-preload.xml.txt](https://github.com/opendatakit/javarosa/files/2516146/uid-preload.xml.txt) generates a UUID in the expected format (e.g. `uuid:c58ea42f-f361-444e-8fb2-ed2f689d2094`).

If we do want to make this change, I don't see any alternative. It's minimally invasive and uses existing functionality.

If users use the `uid` preload and for some reason depend on it having a certain format, this could be disruptive. I think that's pretty unlikely.

### DIFF
--- a/src/org/javarosa/core/model/utils/QuestionPreloader.java
+++ b/src/org/javarosa/core/model/utils/QuestionPreloader.java
@@ -109,8 +109,10 @@ public class QuestionPreloader {
             public String preloadHandled() {
                 return "uid";
             }
+
+            // Generates an RFC 4122 UUID with a "uuid:" prefix
             public IAnswerData handlePreload(String preloadParams) {
-                return new StringData(PropertyUtils.genGUID(25));
+                return new StringData("uuid:" + PropertyUtils.genUUID());
             }
 
             public boolean handlePostProcess(TreeElement node, String params) {
--- a/src/org/javarosa/core/util/PropertyUtils.java
+++ b/src/org/javarosa/core/util/PropertyUtils.java
@@ -42,7 +42,7 @@ public class PropertyUtils {
 
 
     /**
-     * Generate an RFC 1422 Version 4 UUID.
+     * Generate an RFC 4122 Version 4 UUID.
      *
      * @return a uuid
      */


src/org/javarosa/core/model/utils/QuestionPreloader.java
```diff
@@ -109,8 +109,10 @@ public boolean handlePostProcess(TreeElement node, String params) {
             public String preloadHandled() {
                 return "uid";
             }
+
+            // Generates an RFC 4122 UUID with a "uuid:" prefix
             public IAnswerData handlePreload(String preloadParams) {
-                return new StringData(PropertyUtils.genGUID(25));
+                return new StringData("uuid:" + PropertyUtils.genUUID());
```
ggalmazor: Excuse my probably naive question but, why not `UUID.randomUUID().toString()`?

lognaturel: Good question! I was just matching the `uuid()` implementation. I think that it predates Java 5 so that's likely the original reason for a custom method. I can change both if you prefer.

ggalmazor: Don't worry. I was supposing something on that line.

It looks like a good first issue for someone else ;)

PR #382, opened 2018-11-08
[WIP] Document some of the core model classes

lognaturel:
Continuation of https://github.com/opendatakit/collect/pull/2730 for @smoyth and @cooperka to provide some context on existing abstractions that may be helpful for the hierarchy redesign work on Collect.

More broadly, I have been wanting to put the understanding that I've gained of core model classes into JavaDoc so it others can possibly benefit (and/or correct!). Will do this as I can. Feel free to comment and ask questions and I will remove the `WIP` tag when it's ready for a deeper review.

### DIFF
--- a/src/org/javarosa/core/model/FormIndex.java
+++ b/src/org/javarosa/core/model/FormIndex.java
@@ -20,18 +20,68 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.javarosa.core.model.instance.FormInstance;
+import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
 
 /**
- * A Form Index is an immutable index into a specific question definition that
- * will appear in an interaction with a user.
+ * {@code FormIndex} is an immutable index which is structured to provide quick access to a specific node in a
+ * {@link FormDef} (tree of {@link IFormElement} representing a blank form) or a {@link FormInstance}
+ * (tree of {@link TreeElement} representing a form instance that can be filled). It also includes a
+ * {@link TreeReference} representing the XPath path of that node.
  *
- * An index is represented by different levels into hierarchical groups.
+ * The XML node referred to by a FormIndex can correspond to either a question or a group. The index is constructed as a
+ * linked list of levels.
  *
- * Indices can represent both questions and groups.
+ * A FormIndex can refer to:
+ * - a node at the top level of the form (only {@link #localIndex} is used)
+ * - a node nested in a group hierarchy ({@link #localIndex} and {@link #nextLevel} are used)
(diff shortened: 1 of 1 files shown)


src/org/javarosa/core/model/FormIndex.java
```diff
@@ -20,18 +20,68 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import org.javarosa.core.model.instance.FormInstance;
+import org.javarosa.core.model.instance.TreeElement;
```
ggalmazor: It looks like these imports are only required for the JavaDoc comments. Now, this class is coupled with (depends on) `FormInstance` and `TreeElement`, which makes future changes harder to perform.

The value we will get from these imports is having a link from this class' docs to another class, which in my opinion it doesn't justify the tradeoff of having a more coupled/complex design.

In my experience with Aggregate, Briefcase, and JavaRosa, these `{@link}` directives have been a real problem while refactoring and generally working the code. Classes not used by any code line will still be considered as being used (therefore, unsafe to be removed) if they're used in any `{@link}`, which requires them to be imported. This is confusing and really gets in the way of understanding how things work and are related to each other.

I'd strongly advise removing the related `{@link FormInstance}`, and `{@link TreeElement}` directives to avoid importing these classes. If we want to highlight these names in the docs, we can use a `{@code}` directive instead.

cooperka: This is a good point @ggalmazor; since lognaturel is out I made this change here: https://github.com/cooperka/javarosa/commit/bf6257dbfadb080fd9e15f088aeb138cbee4bae5

I can't push to her branch but I assume you can as a Member.

```console
git remote add cooperka https://github.com/cooperka/javarosa.git
git fetch cooperka
git cherry-pick bf6257d
```

Let me know if you'd like any changes.

ggalmazor: Thanks, @cooperka! The change you're proposing doesn't solve the core problem of the class being coupled to those classes. Static code inspection tools will still mark the linked classes as being used by `FormIndex`.

ggalmazor: Would it be acceptable to change those `{@link}` to `{@code}`?

cooperka: @ggalmazor I personally think that would significantly hurt the documentation; I follow `@link`s daily when reading docs both within my IDE and online. It's a simple change if you're sure you want it, but I wish there were a better way :thinking: 

Do you have an example of the impact on a particular static inspection tool? Often they're configurable to ignore javadoc like this. Are you referring to Codecov?

ggalmazor: Hey, @cooperka! Just to be really sure, I've done a smaller scale test with IntelliJ and I've learned that `{@link}` directives **don't actually confuse the unused class inspection tool** :O 

Sorry about that! I'll merge the commit that used qualified names in the `{@link}`.

PR #388, opened 2018-11-27
add new number-or-zero() XPath function

tiritea:
Returns 0 instead of NaN if argument is not a valid numeric expression, eg number-or-zero(${foo})

 Specifically intended to replace the more long-winded coalesce(${foo},0) or if(${foo}!='',${foo},0)

Closes #

this solution is intended to make more 'complicated' solutions less necessary or prone to user mistakes.

Shouldnt be - this new XPath function wont ever get used unless explicitly added to XForm expression by form writer. So will not affect any existing forms.


tiritea:
First ODK PR! Please be gentle... ;-)

tiritea:
FYI would imply corresponding function added to Enketo XPath (@MartijnR)

tiritea:
Cancelled per discussion on ODK Forum; https://forum.opendatakit.org/t/new-number-or-zero-xpath-function/15975

PR #393, opened 2018-12-27
Add substring-before() and substring-after() XPath functions

tiritea:
Closes https://forum.opendatakit.org/t/add-xpath-substring-before-and-substring-after-functions/16863

###What has been done to verify that this works as intended?

tested under Android Studio simulator running ODK Collect v1.18.0 (dirty), using test examples described on http://www.xsltfunctions.com/xsl/fn_substring-before.html and http://www.xsltfunctions.com/xsl/fn_substring-after.html, and compared results running same against Enketo.

###Why is this the best possible solution? Were any other approaches considered?

Implements XPath standardized functions (which is arguably best solution). Implements identical XPath behavior as exists already in Enketo and libxml2.

###Are there any risks to merging this code? If so, what are they?

Introduces entirely new XPath functions and does not touch or modify any existing javaRosa function or behavior. Can only be used by new forms that exploit these new functions in their XPath expressions, so no risk to existing forms.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -345,17 +345,17 @@ public class XPathFuncExpr extends XPathExpression {
         } else if (name.equals("substr") && (args.length == 2 || args.length == 3)) {
             return substring(argVals[0], argVals[1], args.length == 3 ? argVals[2] : null);
         } else if (name.equals("substring-before") && args.length == 2) {
-            String haystack = toString(argVals[0]);
-            String needle = toString(argVals[1]);
-            int needleIndex = haystack.indexOf(needle);
-            // XPath reference states that we should return the empty string when we don't find the needle in the haystack
-            return needleIndex >= 0 ? haystack.substring(0, needleIndex) : "";
+            String str = toString(argVals[0]);
+            String substr = toString(argVals[1]);
+            int pos = str.indexOf(substr);
+            // XPath reference states that we should return the empty string when we don't find the substring
+            return pos >= 0 ? str.substring(0, pos) : "";
         } else if (name.equals("substring-after") && args.length == 2) {
-            String haystack = toString(argVals[0]);
-            String needle = toString(argVals[1]);
-            int needleIndex = haystack.indexOf(needle);
-            // XPath reference states that we should return the empty string when we don't find the needle in the haystack
-            return needleIndex >= 0 ? haystack.substring(needleIndex + needle.length()) : "";
+            String str = toString(argVals[0]);
+            String substr = toString(argVals[1]);
+            int pos = str.indexOf(substr);
+            // XPath reference states that we should return the empty string when we don't find the substring
+            return pos >= 0 ? str.substring(pos + substr.length()) : "";
         } else if (name.equals("contains") && args.length == 2) {
             return toString(argVals[0]).contains(toString(argVals[1]));
         } else if (name.equals("starts-with") && args.length == 2) {
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -345,15 +345,17 @@ public class XPathFuncExpr extends XPathExpression {
         } else if (name.equals("substr") && (args.length == 2 || args.length == 3)) {
             return substring(argVals[0], argVals[1], args.length == 3 ? argVals[2] : null);
         } else if (name.equals("substring-before") && args.length == 2) {
-            String str = toString(argVals[0]);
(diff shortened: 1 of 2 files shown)


src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -344,6 +344,16 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
             }
         } else if (name.equals("substr") && (args.length == 2 || args.length == 3)) {
             return substring(argVals[0], argVals[1], args.length == 3 ? argVals[2] : null);
+        } else if (name.equals("substring-before") && args.length == 2) {
+            String str = toString(argVals[0]);
```
ggalmazor: I'd like to suggest some changes:
- [MDN](https://developer.mozilla.org/en-US/docs/Web/XPath/Functions/substring-before) suggests `haystack` and `needle` argument names. I think those names do a great work of explaining what's going on.
- Use `needleIndex` instead of `pos`. It's not abbreviated, it's more expressive (you're calling an `indexOf` method, therefore, the result should be an index), and it's explicit about its relationship with the `needle`.
- Remove the parenthesis (redundant) in the `return` line.
- I think the comment should be moved to a line before the return and explicitly say that the XPath reference states that an empty string should be returned when the `needle` isn't found. Otherwise, it could look like some arbitrary decision on our side.

```java
String haystack = toString(argVals[0]);
String needle = toString(argVals[1]);
int needleIndex = haystack.indexOf(needle);
// XPath reference states that we should return the empty string when we don't find the needle in the haystack
return needleIndex >= 0 ? haystack.substring(0, needleIndex) : "";
```

tiritea: Thanks for your suggestions. Done.

tiritea:
I believe I've made the necessary changes and its ready, but as its my first ODK PR please let me know if I've omitted anything github-wise. Thnx.

src/org/javarosa/xpath/expr/XPathFuncExpr.java
dcbriccetti: I think the needle in a haystack metaphor is dumb, and phooey on MDN for using it.

https://dictionary.cambridge.org/us/dictionary/english/a-needle-in-a-haystack

tiritea: I have no opinion either way... But I do notice that javaRosa does sometime errs on the side of 'verbosity', so its not completely out of character. eg

return GeoUtils.calculateAreaOfGPSPolygonOnEarthInSquareMeters(latLongs);

;-)

dcbriccetti: Ha ha.
Verbosity is a different dimension from frivolity or suitability, thought. 
Was it a coincidence that you picked something *I* named? ;-)

tiritea: >Was it a coincidence that you picked something I named?

D'oh! :-)

ggalmazor: > I think the needle in a haystack metaphor is dumb, and phooey on MDN for using it.

Huh, I didn't know about that. This seemed to be an improvement over `arg0` and `arg1`, but now I'm hesitating. Do you have any better alternatives, @dcbriccetti?

ggalmazor:
@tiritea, @dcbriccetti, I've approved the PR, although maybe we should give a last thought to the arg0/arg1/haystack/needle conundrum before merging this.

tiritea:
@ggalmazor , @dcbriccetti - it'll let you guys call the shots wrt javaRosa variable name best-practices. I'm happy with either str/substr/pos or haystack/needle/needleIndex

dcbriccetti:
I much prefer str/substr/pos to the haystack business. And both, of course, to arg0....

tiritea:
I'm happy to change it back. But it will mean @ggalmazor owes me a beer... ;-)

dcbriccetti:
Heh heh. Although his English is excellent—his writing, especially—we can forgive @ggalmazor  for not knowing the connotations of “needle in a haystack”. I wonder what the nutty https://developer.mozilla.org/en-US/docs/Web/XPath/Functions/substring-before developer/documenter was thinking, though. One should not have such a negative attitude about a search. ;-)

tiritea:
@ggalmazor - Given that finding said 'needle' is actually incredibly simple when every piece of straw happens to lined up adjacently, perhaps we should adopt a more appropriate idiom here, say "Ser pan comido"?

:-)

dcbriccetti:
Ha ha!

ggalmazor:
Ha ha indeed! In Spanish we say ja ja, by the way.

PR #394, opened 2018-12-28
Issue 390 external instances

### LINKED ISSUE #390: External secondary instances don’t work
#### Software versions 
JavaRosa v2.12.1, Collect v1.18.0

#### Problem description
External secondary instances (see [docs](https://opendatakit.github.io/xforms-spec/#secondary-instances---external)) no longer work in Collect, seemingly due to JavaRosa issues.

1. There are errors parsing external XML
2. The external data location is extremely hard to access

1. Download the [form XML and data XML](https://drive.google.com/drive/folders/1l8LgfP0DXzTlH5YyjPIqNUNmvcOox8Og/)
1. Install an emulator using "Google APIs" (NOT "Google Play" because those don't support root access)
    - This is necessary because of a different regression; see notes below
1. `adb push internal_select_10.xml /sdcard/odk/forms/`
1. `adb push external_select_10.xml /sdcard/odk/forms/`
1. `adb root`
1. `adb shell`
1. As root:
    - `mount -o rw,remount rootfs /`
    - `mkdir /resources`
    - `exit`
1. `adb push external_data_10.xml /resources/`
1. Open the **external select 10** form

#### Expected behavior
The form should open and work normally, identically to **internal select 10**.

Additionally, external data should be stored at `/sdcard/odk/forms/external_select_10-media/`, not root `/resources/`. See [slack discussion](https://opendatakit.slack.com/archives/C35ENHA64/p1544186315008500?thread_ts=1544116019.006900&cid=C35ENHA64) with @grzesiek2010 for more context.

>I think we implemented changes on the javarosa side https://github.com/opendatakit/javarosa/pull/157/ but it doesn't work well in Collect.
>
> From logs, I can see that javarosa tries to read that file from /resources/external_data_10.xml where it should be  /storage/emulated/0/odk/forms/external_select_10-media/external_data_10.xml

#### Actual behavior
Form fails to parse:

```console
XFormParseException: <label> node for itemset doesn't exist! [instance(second)/root/item/itextId]
        at org.javarosa.xform.parse.FormInstanceParser.verifyItemsetBindings(FormInstanceParser.java:344)
```

#### Other information 
Here's the inspector value of `DataInstance fi` for internal vs. external forms:

Internal:

fi internal

External:

fi external

Breakpoint location: https://github.com/opendatakit/javarosa/blob/master@{2018-12-06}/src/org/javarosa/xform/parse/FormInstanceParser.java#L343

dcbriccetti:
Closes #390

New tests using @cooperka’s forms are passing. Editing, saving, and re-editing Kevin’s forms in Collect, with this change, https://github.com/opendatakit/collect/pull/2771, appears to be working.

I’ve filled in the missing pieces, now that I better understand how things are supposed to work.

This changes where JavaRosa looks for external secondary instance files, so any client relying on the previous behavior will need to be changed.


src/org/javarosa/xform/util/ParserCustomizer.java
```diff
@@ -0,0 +1,17 @@
+package org.javarosa.xform.util;
+
+import org.javarosa.xform.parse.XFormParser;
+
+/**
+ * Provides a hook to allow customizing a parser before its parse method is called, allowing the tests
+ * to use the same {@link org.javarosa.xform.util.XFormUtils#getFormFromInputStream(java.io.InputStream)}
+ * call that OpenDataKit Collect uses.
+ */
+public interface ParserCustomizer {
```
ggalmazor: Is this strictly necessary? It doesn't look like it's key for the fix and I don't know if I feel comfortable with adding more (accidental) complexity to the design: it adds a new way to do something that we could do before, which can be confusing for users.

I would consider a new abstraction aimed to remove the `onWarning` and `onError` methods from the parser instance, which would be a change in break even the way I see it, but I don't know how we could push the needed changes to Collect and how that would be justified.

src/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -36,26 +36,27 @@ public ExternalDataInstance(String path, String instanceId, TreeElement root) {
...
     }
 
-    public static ExternalDataInstance buildFromPath(String path, String instanceId)
-            throws IOException, UnfullfilledRequirementsException, XmlPullParserException, InvalidStructureException {
-        String absolutePath = getPathPrefix() + path;
-        KXmlParser xmlParser = ElementParser.instantiateParser(new FileInputStream(absolutePath));
+    /**
+     * Builds an ExternaldataInstance
+     *
+     * @param path       the path to the XML file
```
ggalmazor: Maybe add "absolute" to the wording to make it explicit. It also would explain why line 41 was removed.

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2257,6 +2259,11 @@ private void triggerError(String message) {
             callback.accept(message);
     }
 
+    /** Sets where to locate external instance data: A path to be placed in front of what follows `jr://file/` in the src attribute of the instance element */
```
ggalmazor: It's not clear how to use this method. Should users provide a String with or without the `jr://file` prefix? Maybe add an example to the doc block?

src/org/javarosa/xform/util/XFormUtils.java
```diff
@@ -78,7 +103,12 @@ public static FormDef getFormFromInputStream(InputStream is) throws XFormParseEx
                 throw new XFormParseException("IO Exception during parse! " + uee.getMessage());
             }
 
-            return _factory.getXFormParser(isr).parse();
+            XFormParser xFormParser = _factory.getXFormParser(isr);
+            xFormParser.setExternalInstancePathPrefix(externalInstancePathPrefix);
```
ggalmazor: I understand that this line (107) is the only one needed to fix the issue, right? (see comments about the `ParserCustomizer` class)

test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -108,13 +109,18 @@ public void parsesPreloadForm() throws IOException {
...
     }
 
-    @Test public void parsesSecondaryInstanceForm() throws IOException, XPathSyntaxException {
+    @Test public void parsesSecondaryInstanceForm() throws IOException {
         FormDef formDef = parse(SECONDARY_INSTANCE_XML).formDef;
         assertEquals("Form with secondary instance", formDef.getTitle());
     }
 
+    private String allButLastSubpath(Path formName) {
+        return formName.subpath(0, formName.getNameCount() - 1).toString();
```
ggalmazor: This can be `formName.getParent().toString()`, and I think it could be inlined (no need for a method for that, I think)

test/org/javarosa/xform/parse/FormParserHelper.java
```diff
@@ -14,21 +15,17 @@ private FormParserHelper() {
     }
 
     public static ParseResult parse(Path formName) throws IOException {
-        XFormParser parser = new XFormParser(new FileReader(formName.toString()));
+        return parseWithPrefix(formName, "");
+    }
+
+    public static ParseResult parseWithPrefix(Path formName, String externalInstancePathPrefix) throws IOException {
```
ggalmazor: Two comments on this one:
- Why not call it `parse`?
- I would argue that it's easier to have a second argument `externalInstances` of type `Path`
  - This would let the `parse(Path)` to be implemented as:

    ```java
    public static ParseResult parse(Path formName) throws IOException {
      return parse(formName, formName.getParent());
    }
    ```
  - This would let you avoid some of the changes in `XFormParserTest.java`

test/org/javarosa/test/utils/ResourcePathHelper.java
```diff
@@ -6,16 +6,34 @@
...
+    /**
+     * Makes a Path for a resource file either in a directory corresponding to the test class’s package, or
+     * in the resources directory.
+     *
+     * @param filename the file name for which to create a path
+     * @param fallBack whether to “fall back” to the resources directory if the file is not in the
+     *                 class’s corresponding directory
+     * @return a Path for the resource file
+     */
+    public static Path r(String filename, boolean fallBack) {
```
ggalmazor: I think we should throw an error when a test asks for a file that isn't there. In fact, this implementation in masking an error on this PR: the file `towns-large.xml`is not present (I'm guessing you forgot to commit the file?), but the timing test still passes.

I like the fallback mechanism though, and I think we should have it as the default behavior, which would avoid some changes on the test classes. I've tried this implementation, which works:

```java
public static Path r(String filename) {
    String resourceFileParentPath = inferResourceFileParentPath();
    Path resourceFilePath = Paths.get("resources", resourceFileParentPath, filename);

    if (Files.exists(resourceFilePath))
        return resourceFilePath;

    Path resources = Paths.get("resources", filename);
    if (Files.exists(resources))
        return resources;

    throw new RuntimeException("File " + filename + " not found");
}
```

ggalmazor: Now I think I'd like to change the implementation to recursively search for the file with `Files.walk`, but I'm not super crazy about it either.

test/org/javarosa/xform/parse/FormParserHelper.java
dcbriccetti: Hey, thanks for all the great feedback. To overload parse would mean I could no longer use static imports of parse and would have to change all the callers. I was trying to avoid that.

test/org/javarosa/test/utils/ResourcePathHelper.java
dcbriccetti: `towns-large.xml` is dynamically generated, hence it might not exist when `r` is called.

test/org/javarosa/xform/parse/FormParserHelper.java
dcbriccetti: Oh, I see I can leave that static import alone and call the new two-args `parse` via `FormParserHelper.parse`.

test/org/javarosa/xform/parse/XFormParserTest.java
dcbriccetti: I missed that when I looked up `subpath`. Thanks!

src/org/javarosa/xform/util/ParserCustomizer.java
dcbriccetti: I found a way! :-)

src/org/javarosa/xform/parse/XFormParser.java
dcbriccetti: Good suggestion. Thanks.

dcbriccetti:
I have @cooperka’s forms (see the issue) in Collect, and I’m trying to get them to work. We fail to generate the dynamic options for the second page after pushing a radio button on the first page and swiping to advance. The predicates aren’t matching in `XPathEqExpr.eval`. I’m digging into why.

dcbriccetti:
OK, *that* problem is solved. We weren’t setting the instance ID in the ExternalDataInstance’s root nodes. The next problem to solve may be opening a saved form with external data. Here’s a bit of the stack trace I’ll be examining:
```
12-30 12:00:33.120 21795-21795/org.odk.collect.android E/FormEntryActivity: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.javarosa.core.model.instance.TreeElement.getNumChildren()' on a null object reference
        at org.javarosa.core.model.instance.ExternalDataInstance.getRoot(ExternalDataInstance.java:68)
        at org.javarosa.xpath.expr.XPathPathExprEval.getDataInstance(XPathPathExprEval.java:70)
        at org.javarosa.xpath.expr.XPathPathExprEval.eval(XPathPathExprEval.java:33)
        at org.javarosa.xpath.expr.XPathPathExpr.eval(XPathPathExpr.java:210)
        at org.javarosa.xpath.XPathConditional.evalNodeset(XPathConditional.java:93)
        at org.javarosa.core.model.FormDef.populateDynamicChoices(FormDef.java:993)
```

dcbriccetti:
I just realized we probably don’t want to serialize the contents of the external instance. Rather, we should re-parse it upon deserialization of the form it’s in.

test/org/javarosa/test/utils/ResourcePathHelper.java
ggalmazor: Ugh, I didn't realize that... 

In any case, what do you think about making it the default behavior and avoid having two `parse` methods?

This implementation works on all the test scenarios without polluting unknown paths:
```java
public static Path r(String filename) {
    String resourceFileParentPath = inferResourceFileParentPath();
    Path resourceFilePath = Paths.get("resources", resourceFileParentPath, filename);

    if (Files.exists(resourceFilePath))
        return resourceFilePath;

    Path resources = Paths.get("resources", filename);
    if (Files.exists(resources))
        return resources;

    // File doesn't exist. Returning a temp file
    String name = filename.substring(0, filename.lastIndexOf("."));
    String extension = filename.substring(filename.lastIndexOf("."));
    try {
        return Files.createTempFile(name + "-", extension);
    } catch (IOException e) {
        throw new UncheckedIOException(e);
    }
}
```

This code would create a `/tmp/towns-large-8296528288103120809.xml`, for example.

src/org/javarosa/xform/util/ParserCustomizer.java
ggalmazor: I still think there's an issue with this. Honestly, passing an output argument feels worse than the customizer class, especially for a method that's exposed as part of JR's API.

I now realize that the core reason for this change was the need of checking the parse errors and warnings on the tests (and, potentially on clients), which wasn't possible to do with the `XFormParser.onWarning()` and `XFormParser.onError()` methods since the `XFormUtils.getFormFromInputStream()` doesn't give us a chance to subscribe before actually parsing the form.

I think that a workaround for that could be to save those parse messages in the `FormDef` itself. This makes sense to me:
- Clients can't rely 100% on the `onWarning()` and `onError()` methods, as proven by this change we're discussing.
- JR doesn't provide any other means to know if something goes wrong with the parsing of forms (it logs them, but that's not really providing a way for clients to get them).
- It's reasonable that clients want to check/display parse messages after the fact.

I've tried this here: https://github.com/ggalmazor/javarosa/commit/8c6ac1089187df738fb242787090a185abe0d414

dcbriccetti: I like this! I’ll merge in your change. Thanks.

src/org/javarosa/core/reference/RootTranslator.java
```diff
@@ -52,29 +52,16 @@ public RootTranslator(String prefix, String translatedPrefix) {
...
-    /* (non-Javadoc)
-     * @see org.commcare.reference.Root#derives(java.lang.String)
-     */
     public boolean derives(String URI) {
-        if(URI.startsWith(prefix) && !URI.startsWith(translatedPrefix)) {
-            return true;
-        } else{
-            return false;
-        }
+        return URI.startsWith(prefix) && !URI.startsWith(translatedPrefix);
```
ggalmazor: Thanks!

dcbriccetti:
That’s the way I first wrote it. But:
/Users/daveb/devel/opendatakit/javarosa/src/org/javarosa/core/reference/ReferenceManager.java:207: error: incompatible types: ArrayList<? extends ReferenceFactory> cannot be converted to List<ReferenceFactory>
        for (List<ReferenceFactory> rfs : Arrays.asList(sessionTranslators, translators, factories)) {

src/org/javarosa/core/reference/ReferenceManager.java
```diff
@@ -203,28 +191,14 @@ public void clearSession() {
...
-            }
-        }
-
-        //Now try all of the raw connectors available
-        for (ReferenceFactory root : factories) {
-            if (root.derives(uri)) {
-                return root;
-            }
-        }
+        for (List<? extends ReferenceFactory> rfs : Arrays.asList(sessionTranslators, translators, factories))
```
ggalmazor: You could do this:
```java
private ReferenceFactory derivingRoot(String uri) throws InvalidReferenceException {
    for (ReferenceFactory rf : getAllReferenceFactoriesInOrder())
        if (rf.derives(uri))
            return rf;

    throw new InvalidReferenceException(getPrettyPrintException(uri), uri);
}

private List<ReferenceFactory> getAllReferenceFactoriesInOrder() {
    List<ReferenceFactory> refFactories = new ArrayList<>();
    refFactories.addAll(sessionTranslators);
    refFactories.addAll(translators);
    refFactories.addAll(factories);
    return refFactories;
}
```

What do you think?

ggalmazor: No problem! :dagger: 

I can't help to not try to fix design problems. Those subtypes are leaking because that hierarchy is not sealed tight. The code I'm suggesting is more verbose but deals with that design flaw. It also flattens the lists, which is kind of nice too.

I won't cry if you don't change it :)

dcbriccetti: Thank you for not crying. 

Sometime I’d like to understand the design flaw this would deal with.

ggalmazor: (I don't know why a dagger was showing up in my last comment!)

This is something super common not only in the ODK codebase, but in many other projects as well, and I think we can blame Java and how Java has been historically teached for that.

The flaw is that the hierarchy formed by `ReferenceFactory` (supertype) and its subtypes `RootTranslator`, `PrefixedRootFactory`, `ResourceReferenceFactory`, ..., is leaking the subtypes.

You can see this leak in many places. Related to this change we're talking about:
- You can't change the type of the `ReferenceManager` `translators` and `sessionTranslators` to a `List<ReferenceFactory>` because it's coupled with a method on `RootTranslator`.
- Because of this, you can't directly get a `List<List<ReferenceFactory>>` as a result of an `Array.asList(sessionTranslators, translators, factories)`.

If you're a SOLID design fan, this is breaking the L (Liskov's Substitution Principle), but in a more general way, this is wrong because it's a misuse of inheritance.

Inheritance is a tool we use to implement polymorphism, and here we don't have polymorphism, which is weird because we could benefit from it (the `for` block we're discussing would benefit from it).

In a polymorphic type, you can't have the subtypes leaked outside the type hierarchy (no one can be coupled to them, even implicitly, like with the `List<? extends ReferenceFactory>`).

If this is not a polymorphic type, this is telling me that:
- Maybe we have used inheritance for code reuse, which would be much simpler to get by composing stuff together
- Maybe we have used inheritance systematically without taking into account the impact this has in evolving software and the cost of introducing new changes.

This is even worse because the `ReferenceFactory` is an extension point that clients use to adapt JR to their particular environment. It's really confusing that JR is coupled both to the supertype (OK) and the subtypes (wrong).

Since this is all across the codebase, it's not super critical to fix in this particular instance but you know the Boy Scout's rule: leave the place cleaner than you found it.

ggalmazor:
Even though there are two standing discussions (the one about `ResourcePathHelper`, and the one about `ReferenceFactory`), I think this PR is solid work and solves the issue. 

I've also tested this in Collect, so I'm confident this is a step in the right direction.

@dcbriccetti, if you feel the same, I'd suggest we merge so that it's easier for you to continue with #397

src/org/javarosa/core/reference/ReferenceManager.java
dcbriccetti: This is very educational!

PR #395, opened 2019-01-03
Add XPath translate() function

tiritea:
Closes # https://forum.opendatakit.org/t/add-xpath-translate-function/17017

tested under Android Studio simulator running ODK Collect v1.18.0 (dirty), using test examples described on http://www.xsltfunctions.com/xsl/fn_translate.html (and those added to testEval testcases), and compared results running same against Enketo.

Implements XPath standardized function (which is arguably best solution). Implements identical XPath behavior as exists already in Enketo and libxml2.

BTW, I also tried an implementation that avoided using a StringBuilder, and instead tried to exploit existing java String replace()/replaceAll() functions, but this ended up much messier, required multiple re-instantiations of String objects, and ultimately did not handle simultaneous interchanging translations correctly (eg 'x' --> 'y' and 'y' --> 'x').

Introduces entirely new XPath function and does not touch or modify any existing javaRosa function or behavior. Can only be used by new forms that exploit this new function in their XPath expressions, so no risk to existing forms.

### DIFF
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -356,6 +356,26 @@ public class XPathFuncExpr extends XPathExpression {
             int pos = str.indexOf(substr);
             // XPath reference states that we should return the empty string when we don't find the substring
             return pos >= 0 ? str.substring(pos + substr.length()) : "";
+        } else if (name.equals("translate") && args.length == 3) {
+            String str = toString(argVals[0]);
+            String fromChars = toString(argVals[1]);
+            String toChars = toString(argVals[2]);
+            StringBuilder result = new StringBuilder();
+            // iterate thru each char in the original string
+            for (int i = 0; i < str.length(); i++) {
+                char from = str.charAt(i);
+                // determine if there is a mapping for this char
+                int fromPos = fromChars.indexOf(from);
+                if (fromPos == -1) {
+                    // no mapping for this char, so just add it to result unchanged
+                    result.append(from);
+                } else if (fromPos < toChars.length()) {
+                    // replace with the corresponding char it is mapped to
+                    result.append(toChars.charAt(fromPos));
+                }
+                // else the char is mapped to nothing, so per XPath definition we 'delete' it from the string by simply not appending it
+            }
+            return result.toString();
         } else if (name.equals("contains") && args.length == 2) {
             return toString(argVals[0]).contains(toString(argVals[1]));
         } else if (name.equals("starts-with") && args.length == 2) {
--- a/src/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -360,6 +360,7 @@ public class XPathFuncExpr extends XPathExpression {
             String str = toString(argVals[0]);
             String fromChars = toString(argVals[1]);
             String toChars = toString(argVals[2]);
+            int toNumChars = toChars.length();
             StringBuilder result = new StringBuilder();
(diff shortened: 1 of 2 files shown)


src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -356,6 +356,26 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
...
+            StringBuilder result = new StringBuilder();
+            // iterate thru each char in the original string
+            for (int i = 0; i < str.length(); i++) {
+                char from = str.charAt(i);
+                // determine if there is a mapping for this char
+                int fromPos = fromChars.indexOf(from);
+                if (fromPos == -1) {
+                    // no mapping for this char, so just add it to result unchanged
+                    result.append(from);
+                } else if (fromPos < toChars.length()) {
```
ggalmazor: `toChars.length()` could be pre-computed before the `for` block to avoid computing it each loop iteration.

test/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -260,6 +260,15 @@ public void doTests () {
...
         testEval("substring-after('','q')", "");
+        testEval("translate('hello','l','L')", "heLLo");
+        testEval("translate('hello','q','Q')", "hello");
+        testEval("translate('hello','','L')", "hello");
+        testEval("translate('hello','l','')", "heo");
+        testEval("translate('','l','L')", "");
+        testEval("translate('hello','lo','LO')", "heLLO");
+        testEval("translate('hello world','hello','')", "wrd");
+        testEval("translate('hello wor,ld',' ,',', ')", "hello,wor ld");
+        testEval("translate('2019/01/02','/','-')", "2019-01-02");
```
ggalmazor: Neat! Thanks for thinking about the tests! :)

src/org/javarosa/xpath/expr/XPathFuncExpr.java
tiritea: Done. Good catch. Thnx.

BTW I left the other potential candidate for this optimization - namely the for loop terminating expression - unchanged, for internal consistency with existing toNumeric().

yanokwa:
It'd be good for this to be documented in the docs. The bare minimum should be an issue is filed at https://github.com/opendatakit/docs/issues/new!

tiritea:
My plan is to doc this translate() together with substring-before() and substring-after() in one PR. Hence I was waiting for this one to be approved before proceeding with updating docs. But I can open a doc issue in mean time.

tiritea:
Are we waiting on me for anything before merging this? If we can get it in before 16th we might be able to rev javaRosa to including this [and the new substring-before() & substring-after() functions] for next Collect release.

ggalmazor:
Normally we would try the PR in Collect as a sort of QA phase before merging. Do you have any form we could use to test this?

In any case, I'd say we merge it because It's a narrow changeset and it has automated tests. If no one complains by the end of day I can merge this.

tiritea:
> Do you have any form we could use to test this?

Only the one I posted to the Forum: https://forum.opendatakit.org/t/add-xpath-translate-function/17017/3

Sorry. I didn't mean to appear impatient. I just wanted to make sure there wasn't something waiting on _me_ to move it along...  I think I covered all same bases as my other PR PR that got merged, but then I've already violated numerous procedural 'protocols' on both! :-)

ggalmazor:
Sure, don't worry! I think you're right. I didn't remember the previous PR with the substring functions. Let's merge this one.

tiritea:
> I didn't remember the previous PR with the substring functions.

You mean you don't remember the beer you owe me!!!??? Hurmph.

;-)

PR #396, opened 2019-01-04
PRs should have information to help with testing and docs

yanokwa:
This matches the questions we ask for Collect.
https://github.com/opendatakit/collect/blob/master/.github/PULL_REQUEST_TEMPLATE.md

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,4 +4,8 @@ Closes #
 
 #### Why is this the best possible solution? Were any other approaches considered?
 
-#### Are there any risks to merging this code? If so, what are they?
+#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
+
+#### Do we need any specific form for testing your changes? If so, please attach one.
+
+#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/opendatakit/docs/issues/new) and include the link below.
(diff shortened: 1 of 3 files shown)


tiritea:
Somewhat related, do all javaRosa PR's require an github issue, or is referencing Forum feature request sufficient? There seems to be some duplication here, especially wrt approval new feature requests... should new features pass thru Forum Features first, then *after* 'approved' open a github issue (?). Then what should trigger submitting actual change PR?

Just for my future reference; I'm not sure what the correct procedure is (whatever it is, I'm sure I've violated it in some manner... :-)

ggalmazor:
The only nuance here is that "users" are developers of other apps that import JavaRosa. I don't know how you feel about changing the language to address this difference.

yanokwa:
@ggalmazor I appreciate your attention to detail here because it's the kind of think I would complain about 😂, but in this case I think it's fine.

@tiritea The official process is documented at https://github.com/opendatakit/governance/tree/master/TSC1#roadmapping. Our governance suggests (and I prefer) an issue that the PR is closing. We start on the forum to give non-devs who aren't on GitHub a chance to chime in. Once that discussion is done, then an issue is filed. Changes to JR doesn't always have a user-facing impact, but it might and so we want to give folks a chance to comment. Also, it's good practice.

I hear you that there is some duplication and it's a wee bit onerous, but I think the consistency of a process that works for most of the features we do outweighs the annoyance of creating an issue. If it's truly a barrier to contribution, then I'm open to tweaking it.

tiritea:
No barrier, and I certainly wouldn't call cut-n-pasting a few lines of text from the Forum onerous... :-) I shalt endeavor to follow the official process in the future, thnx much.

PR #397, opened 2019-01-06
WIP:  Support CSV external data

dcbriccetti:
Closes #

Trying out the idea here: https://github.com/opendatakit/xforms-spec/issues/88, with an optimization from @ggalmazor to skip the XML parsing altogether.

### DIFF
--- a/test/org/javarosa/xform/parse/XFormParserTest.java
+++ b/test/org/javarosa/xform/parse/XFormParserTest.java
@@ -41,6 +41,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.text.NumberFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -54,7 +55,8 @@ import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsInstanceOf.instanceOf;
 import static org.javarosa.core.model.Constants.CONTROL_RANGE;
 import static org.javarosa.core.model.Constants.CONTROL_RANK;
-import static org.javarosa.core.reference.ReferenceManagerTest.buildReferenceFactory;
+import static org.javarosa.core.reference.ReferenceManagerTestUtils.buildReferenceFactory;
+import static org.javarosa.core.reference.ReferenceManagerTestUtils.setUpSimpleReferenceManager;
 import static org.javarosa.core.util.externalizable.ExtUtil.defaultPrototypes;
 import static org.javarosa.test.utils.ResourcePathHelper.r;
 import static org.javarosa.xform.parse.FormParserHelper.parse;
@@ -105,7 +107,11 @@ public class XFormParserTest {
 
     @Test
     public void parsesPreloadForm() throws IOException {
-        FormDef formDef = parse(r("Sample-Preloading.xml"));
+        // The form on this test uses a jr://file-csv resource.
+        // We need to prime the ReferenceManager to deal with those
+        Path form = r("Sample-Preloading.xml");
+        setUpSimpleReferenceManager("file-csv", form.getParent());
+        FormDef formDef = parse(form);
         assertEquals("Sample Form - Preloading", formDef.getTitle());
     }
 
@@ -140,6 +146,13 @@ public class XFormParserTest {
 
     @Test public void parsesExternalSecondaryInstanceForm2() throws IOException {
         Path formName = r("external_select_10.xml");
+        mapFileToResourcePath(formName);
+        FormDef formDef = parse(formName);
+        assertEquals("external select 10", formDef.getTitle());
+    }
+
+    @Test public void parsesExternalSecondaryInstanceCsvForm() throws IOException {
+        Path formName = r("external-select-csv.xml");
         mapFileToResourcePath(formName);
         FormDef formDef = parse(formName);
         assertEquals("external select 10", formDef.getTitle());
@@ -437,7 +450,8 @@ public class XFormParserTest {
     private void mapFileToResourcePath(Path formPath) {
         ReferenceManager rm = ReferenceManager.instance();
         rm.reset();
-        rm.addReferenceFactory(buildReferenceFactory("file", formPath.getParent().toString()));
+        for (String t : Arrays.asList("file", "file-csv"))
+            rm.addReferenceFactory(buildReferenceFactory(t, formPath.getParent().toString()));
     }
 
     /** Generates large versions of a secondary instance */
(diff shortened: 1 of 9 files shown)


src/org/javarosa/core/model/instance/utils/CsvAsXmlInputStream.java
```diff
@@ -0,0 +1,46 @@
...
+                for (int i = 0; i < fieldNames.length; ++i) {
+                    sb.append(String.format("\t\t<%s>%s</%s>\n", fieldNames[i], data[i], fieldNames[i]));
+                }
+                sb.append("\t</item>\n");
+            }
+        }
+
+        sb.append("</root>\n");
+
+        stream = new ByteArrayInputStream(sb.toString().getBytes(StandardCharsets.UTF_8));
```
dcbriccetti: If we like this approach, we can change  this class so it doesn’t read the whole file into memory.

ggalmazor: My only concern is that this is reading all bytes from the original input stream to produce an XML document and then, that XML document is parsed (effectively reading the contents twice).

Since each line represents an item, maybe we could manually produce one `TreeElement` for each row which could be added to a root element, and then return it. I think that this could be more memory and time efficient. It could be even faster than the XML loader.

ggalmazor: This is what I was thinking about: https://github.com/ggalmazor/javarosa/commit/cd7c991f51e2cf5d56264fd004c30554a7b104d2

dcbriccetti: You’re a clever one!

dcbriccetti:
@ggalmazor, thanks for that excellent code. It’s here now.

We now have a conflict, given that file-csv seems to have special meaning in Collect, and that previously JavaRosa ignored it. 

Now we have this test failing:
`public void parsesPreloadForm() throws IOException`

ggalmazor:
The commits so far are looking good ;)

To solve the issue with that test, one would have to set up the ReferenceManager before parsing the form. I've prepared two commits on top of your PR to show how I'd go about it:
- https://github.com/ggalmazor/javarosa/commit/9a843f9f9d37a0e9cb49625c6cf55a3779529d49
- https://github.com/ggalmazor/javarosa/commit/36b97be5dff2c779fd9b55ca8f5b2d1b18add8e5

dcbriccetti:
@ggalmazor thanks for the coded solution. I cherry picked, and made a small correction (setUp instead of setup—phrasal verb, two words, to camel case) to your second commit. Was the first commit just formatting changes? I didn’t take it because it loses all my hand-beautifications.

What’s the next step for this project?

dcbriccetti:
Thanks. Perhaps we’d like to meet again to discuss next steps.

ggalmazor:
Sure, let's talk it over Slack and touch base.

test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -105,7 +107,11 @@ public void parsesForm2() throws IOException {
 
     @Test
     public void parsesPreloadForm() throws IOException {
-        FormDef formDef = parse(r("Sample-Preloading.xml"));
+        // The form on this test uses a jr://file-csv resource.
+        // We need to prime the ReferenceManager to deal with those
+        Path form = r("Sample-Preloading.xml");
```
lognaturel: @dcbriccetti @ggalmazor This particular form doesn't actually use the instance loaded by `file-csv` because it relies on the Collect-only database-backed implementation of the `search()` appearance/function: https://github.com/opendatakit/collect/blob/d52ce0bfc63a11fb7f3fc62e0c0fd7c58684527a/collect_app/src/main/java/org/odk/collect/android/external/handler/ExternalDataHandlerSearch.java. You'll see the form doesn't actually use the instance. It's not a big deal since this test only verifies that the form can be parsed but it would probably be worth changing to avoid confusion.

I would also recommend avoiding the "preload' word unless referring to https://opendatakit.github.io/xforms-spec/#preload-attributes

lognaturel:
It looks like this was never tried in Collect, right? It interferes with one of the existing client-only database-backed external CSV implementations as discovered in #415. It also looks like to get this to work, a Collect change would be needed, at the very least to add a translator for `jr://file-csv`.

Since we definitely do not want to parse a CSV at the JR level if one of the other database-backed implementations is used (for performance), I think we should disable this support for now. I'm not really coming up with a great way to only parse the CSV if the instance is actually referred to in the form definition. That is, https://github.com/opendatakit/javarosa/blob/1547174f521c47335cc4d87dc5c6638ef93ea570/resources/Sample-Preloading.xml defines `<instance id="hhplotdetails" src="jr://file-csv/hhplotdetails.csv"/>` but actually never uses `instance("hhplotdetails")`. Instead, it uses `pulldata` and `search`. [This forum post](https://forum.opendatakit.org/t/external-data-current-state-of-affairs/14564/3?u=ln) has more context.

test/org/javarosa/xform/parse/XFormParserTest.java
ggalmazor: We should file an issue for this change

ggalmazor:
There's something I'm missing for sure, but what's the relation between `jr://file-csv` and the client-only database-backed external CSV implementations? I mean:

> Since we definitely do not want to parse a CSV at the JR level if one of the other database-backed implementations is used (for performance), I think we should disable this support for now

I believe this would be a mistake in the form's design. Using the pulldata/search solution in xlsforms doesn't create an xform that includes a secondary external instance using `file-csv`. I don't see what's the relation here (I'm probably missing something).

Do these recent discoveries challenge our decision to progressively deprecate other ways to load external data, such as pulldata/search, and to implement native CSV external data parsing in JR?

If I understand correctly, a fix to the way `jr://file/` and `jr://file-csv/` would some all the issues detected so far, right?

Regarding QA testing, we decided to merge this into master to release a SNAPSHOT and then eventually test it on Collect. We deemed that to be a safe move because the new "hostname" `file-csv` acts effectively as a feature toggle. Unfortunately, it looks like we didn't detect the bug our changes introduced in the existing "hostname" `jr://file` :(.

lognaturel:
I think I addressed your questions/concerns in https://github.com/opendatakit/javarosa/issues/417, @ggalmazor! We can continue the discussion there when it comes to figuring out what to actually do about it.

test/org/javarosa/xform/parse/XFormParserTest.java
lognaturel: I propose removing that test in https://github.com/opendatakit/javarosa/pull/452

lognaturel:
https://github.com/opendatakit/javarosa/pull/452 has a proposed fix for #417 and other various little issues related to CSV data files.

PR #401, opened 2019-01-23
Add methods: isPresentationGroup, isLogicalGroup

cooperka:
Changes:
- Adds 2 new methods to `FormEntryCaption` in order to determine if a particular group is a "presentation group" and/or a "logical group" (see [docs](https://opendatakit.github.io/xforms-spec/#groups) for definitions).

Verified with https://github.com/opendatakit/collect/pull/2815.

This simplifies the logic in Collect previously used to determine group status.

I previously tried using `getAdditionalAttribute(null, "ref")` to see if a group has a `ref`, but `ref` is not available in the list of "additional" attributes because it's used by JR internally (and thus removed from the list).

It adds 2 new methods. `GroupDef` serialization has also been updated accordingly.

❓ Due to serialization, existing forms may **not** be recognized as having "logical groups" until they're re-parsed. I got conflicting results locally, but it seems like `isLogicalGroup` will only be set on re-parse, which will only happen **after the form XML is updated**. Can anyone confirm if that's accurate?

See linked Collect PR.

Not that I know of.

### DIFF
--- a/src/org/javarosa/form/api/FormEntryCaption.java
+++ b/src/org/javarosa/form/api/FormEntryCaption.java
@@ -379,6 +379,30 @@ public class FormEntryCaption implements FormElementStateListener {
         }
     }
 
+    /**
+     * Returns true if the group has a displayable label,
+     * i.e. it's a "presentation group".
+     */
+    public boolean isPresentationGroup() {
+        if (!(element instanceof GroupDef)) {
+            return false;
+        }
+
+        return getShortText() != null;
+    }
+
+    /**
+     * Returns true if the group has an XML `ref` attribute,
+     * i.e. it's a "logical group".
+     */
+    public boolean isLogicalGroup() {
+        if (!(element instanceof GroupDef)) {
+            return false;
+        }
+
+        return ((GroupDef) element).isLogicalGroup;
+    }
+
     public FormIndex getIndex() {
         return index;
     }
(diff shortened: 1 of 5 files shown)


cooperka:
Closing per discussion with @yanokwa. This is likely not the direction we want to go.

yanokwa:
And just so future @yanokwa has the rationale past @yanokwa used.
- This PR changes the serialization and results in old (or non-updated forms) having different behavior than new forms. Re-serializing everything could be expensive.
- There is already a fair bit of fragile code in Collect. If we are concerned about breakage, I'd prefer a unit test that mirrors the code in Collect so we know when things have failed.

src/org/javarosa/form/api/FormEntryCaption.java
```diff
@@ -379,6 +379,30 @@ public boolean repeats() {
...
+
+        return getShortText() != null;
+    }
+
+    /**
+     * Returns true if the group has an XML `ref` attribute,
+     * i.e. it's a "logical group".
+     */
+    public boolean isLogicalGroup() {
+        if (!(element instanceof GroupDef)) {
```
lognaturel: Possibly naïve question from a passerby who is very out of the loop: why not use the same logic as you did at https://github.com/opendatakit/collect/pull/2815/files#diff-586e0a142c0ade6fb5d285ebc8da0965L710? It's still ugly but at least it's hidden in JavaRosa. That way serialization isn't affected. Since you are adding this to `FormEntryCaption`, you should have access to all the same objects, I believe.

PR #406, opened 2019-02-15
Remembering previously entered values (auto-fill)

cooperka:
Companion to https://github.com/opendatakit/collect/pull/2882. See that PR for a full description.

This change basically just maps an instance's `src` value of `"jr://instance/last-saved"` to whatever is passed in as `lastSavedSrc`.

Note: A `null` value for `lastSavedSrc` is fine and means no instance `src` is provided, meaning the instance will be treated as **internal** instead of **external**. The internal `<instance id="last-saved"/>` has no children and thus will provide empty values for anything referring to it.


yanokwa:
@cooperka This will likely require a PR to https://github.com/opendatakit/xforms-spec as well. Can you please file an issue there so we don't forget?

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -348,15 +348,15 @@ public XFormParser(Document form, Document instance) {
         _instDoc = instance;
     }
 
-    public FormDef parse() throws IOException {
+    public FormDef parse(String lastSavedSrc) throws IOException {
```
ggalmazor: ~Could you restore the old `parse()` method to maintain backwards compatibility? It can call to `parse(String lastSavedSrc)` with a null.~ I see that you added overrides in another commit ;)

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -439,7 +439,7 @@ public static Document getXMLDocument(Reader reader, CacheTable<String> stringCa
         return doc;
     }
 
-    private void parseDoc(Map<String, String> namespacePrefixesByUri) {
+    private void parseDoc(Map<String, String> namespacePrefixesByUri, String lastSavedSrc) {
```
ggalmazor: ~Here there's no need to maintain backwards compatibility because this method was already private.~ Don't mind me ;)

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -461,10 +461,23 @@ private void parseDoc(Map<String, String> namespacePrefixesByUri) {
                 final String instanceId = instanceNodeIdStrs.get(instanceIndex);
                 final String instanceSrc = instance.getAttributeValue(null, "src");
 
-                if (instanceSrc != null && instanceSrc.toLowerCase().startsWith("jr://file")) { // file or file-csv
+                final String externalSrc;
```
ggalmazor: This change is super helpful to understand what's going on. Thanks!

I'd suggest you extract it to a method, so that we can have a nice method name that improves semantics and externalSrc can be immutable in the context of `parseDoc()`

cooperka:
Thanks @ggalmazor, I added one basic test and I'd love help with the second one. It's probably a simple thing; I have a file I want to read from, but I get this error:

```console
XFormParseException: Unable to parse external secondary instance: org.javarosa.core.reference.InvalidReferenceException:
The reference "jr://file/resources/org/javarosa/xform/parse/last-saved-filled.xml" was invalid and couldn't be understood. The javarosa jr:// reference root "file" is not available on this system and may have been mis-typed.
```

ggalmazor:
Hi, @cooperka! Sorry for the delay :P

You could try this:

```java
Path lastSavedSubmissionDirectory = r("last-saved-filled.xml").toAbsolutePath().getParent();
ReferenceManagerTestUtils.setUpSimpleReferenceManager("file", lastSavedSubmissionDirectory);
FormDef formDef = parse(formName, "jr://file/last-saved-filled.xml");
```

I've tried it in your branch and it passes the test. Here's how that works:
- We configure JR to believe that the `jr://file` schema has to be resolved to the directory holding the last saved submission in your computer.
- We tell JR to parse the form assuming the last saved submission is at the root of `jr://file`

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -461,10 +461,23 @@ private void parseDoc(Map<String, String> namespacePrefixesByUri) {
...
+                if (instanceSrc == null) {
+                    // It's internal, not external.
+                    externalSrc = null;
+                } else if (instanceSrc.toLowerCase().startsWith("jr://file")) { // file or file-csv
```
zestyping: Strictly speaking, shouldn't this check for `startsWith("jr://file/")`?

ggalmazor: I believe this is to support both the last-saved instance from this PR, and the external csv itemsets feature, which uses a `file-csv` "host". 

Now that you mention it, maybe we should separate both cases to improve semantics at the expense of some duplication? Is this what you had in mind, @zestyping?

zestyping: That seems like a good idea.  Certainly, reading the code as it is now, it's not evident that it's intended to accept `jr://file/abcd` and `jr://file-csv/abcd`.

src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -461,10 +461,23 @@ private void parseDoc(Map<String, String> namespacePrefixesByUri) {
...
+                    externalSrc = instanceSrc;
+                } else if (instanceSrc.toLowerCase().startsWith("jr://instance/last-saved")) {
```
zestyping: Should this be `equals` instead of `startsWith`?

ggalmazor: I think you're right, @zestyping

cooperka:
All feedback addressed, thank you both!

PR #407, opened 2019-03-06
Ensure formatted datetimes follow the ISO8601 #404

### LINKED ISSUE #404: Ensure formatted datetimes follow the spec's ISO8601 format
#### Problem description
Following the directions of the [ISO8601 adaptations doc](https://docs.google.com/document/d/1vPuEgapHvYWINnoa9QiG1R0HWMiCS4_8vtP_Zx3bGec/edit#heading=h.17wejfqu7qg9), we should fix how JR formats datetimes. Specifically, JR uses a wrong format when the time offset falls into whole hours:

- Wrong: `2019-02-11T16:04:45.638+07`
- Correct: `2019-02-11T16:04:45.638+07:00`

From the spec doc:
> JavaRosa doesn’t always use the correct offset format
> When the local timezone offset is of multiples of a full hour, it outputs this format:
> 
> [HH]:[MM]:[SS].[SSS]±[hh] as in 20:04:16.669+09
> 
> Only when the local timezone offset is of odd multiples of half an hour, it uses the format we want to use:
> 
> [HH]:[MM]:[SS].[SSS]±[hh]:[mm] as in 20:04:16.669+09:30
> 
> We should change JavaRosa to ensure it always adds the minutes fragment in the timezone offset text.

#### Other information 
Forum discussion bits:

[Clint says:](https://forum.opendatakit.org/t/odk-central-v0-1-alpha/13437/31?u=ggalmazor)
> it looks like the problem here is that whichever version of Collect you are using isn't outputting the correct date format. compare:
> 
> `2019-02-11T16:04:45.638+07`
> `2019-02-11T16:04:45.638+07:00`

[Gareth says:](https://forum.opendatakit.org/t/odk-central-v0-1-alpha/13437/35?u=ggalmazor)
> FYI I think it may be here: https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/core/model/utils/DateUtils.java
>
> ```java
> private static String formatTimeISO8601 (DateFields f) {
>   ...
>   String hrs = intPad(value / 60, 2);
>   String mins = value % 60 != 0 ? ":" + intPad(value % 60, 2) :"";
>   time += offsetSign + hrs + mins;
> ```
> ie if on hour it drops the mins off the (date)time suffix. On the surface it appears it could be a quick fix, in of itself, but given its potential impact elsewhere it could require some regression testing...?

tiritea:
fix for https://github.com/opendatakit/javarosa/issues/404

- always display timezone offset minutes, instead of dropping them if 00

Closes #404 

Tested using form containing dateTime user input, now() calculation, and date(decimal-date-time(now)) as different means which generate dateTime results.

Ran 'before' test to confirm all the resulting dateTime strings in the submission instance XML had dropped the minutes:

`<?xml version='1.0' ?><data id="snapshot_xml" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:orx="http://openrosa.org/xforms" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:xsd="http://www.w3.org/2001/XMLSchema"><datetime>2019-03-06T12:32:00.000+13</datetime><show_datetime /><now>2019-03-06T12:32:32.387+13</now><show_now /><calc>2019-03-05T00:00:00.000+13</calc><show_calc /><meta><instanceID>uuid:49d1cf07-2644-4653-a2f2-0bac47acde7a</instanceID></meta></data>
`

Then re-ran test using same form after making change to confirm all the resulting dateTime strings in the submission instance XML now correctly showed 00 minutes:

`<?xml version='1.0' ?><data id="snapshot_xml" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:orx="http://openrosa.org/xforms" xmlns:odk="http://www.opendatakit.org/xforms"><datetime>2019-03-06T12:40:00.000+13:00</datetime><show_datetime /><now>2019-03-06T12:40:31.132+13:00</now><show_now /><calc>2019-03-05T00:00:00.000+13:00</calc><show_calc /><meta><instanceID>uuid:79cdb162-077f-419b-861a-631ae4fa008b</instanceID></meta></data>
`

Only obvious place to fix. No other approach considered

This change will result in submitted dateTime being fully iso8601 conformant. Given that tools consuming these dateTime must already have been able to handle timezone offset minutes - because they would previously have been present should the offset have been, say, 30min, then they should already be able to handle 00 minutes.

I dont believe the dropping of 00 minutes feature was previously documented anywhere, so there should be no doc changes necessary. [@ggalmazor can you confirm?]

### DIFF
--- a/src/org/javarosa/core/model/utils/DateUtils.java
+++ b/src/org/javarosa/core/model/utils/DateUtils.java
@@ -249,7 +249,7 @@ public class DateUtils {
             int value = Math.abs(offset) / 1000 / 60;
 
             String hrs = intPad(value / 60, 2);
-            String mins = value % 60 != 0 ? ":" + intPad(value % 60, 2) :"";
+            String mins = ":" + intPad(value % 60, 2);
 
             time += offsetSign + hrs + mins;
         }
--- a/test/org/javarosa/core/model/utils/test/DateUtilsSCTOTests.java
+++ b/test/org/javarosa/core/model/utils/test/DateUtilsSCTOTests.java
@@ -43,7 +43,7 @@ public class DateUtilsSCTOTests {
         Date date = DateUtils.parseDateTime("2014-10-05T00:03:05.244+03");
         String str = DateUtils.formatDateTime(date, DateUtils.FORMAT_ISO8601);
 
-        assertEquals("2014-10-04T23:03:05.244+0200", str);
+        assertEquals("2014-10-04T23:03:05.244+02:00", str);
     }
 
     @Test
@@ -53,7 +53,7 @@ public class DateUtilsSCTOTests {
         Date date = DateUtils.parseDateTime("2014-10-05T00:03:05.244+03");
         String str = DateUtils.formatDateTime(date, DateUtils.FORMAT_ISO8601);
 
-        assertEquals("2014-10-05T00:03:05.244+0300", str);
+        assertEquals("2014-10-05T00:03:05.244+03:00", str);
     }
 
     @Test
@@ -66,7 +66,7 @@ public class DateUtilsSCTOTests {
 
         String formatted = DateUtils.formatTime(date, DateUtils.FORMAT_ISO8601);
 
-        assertEquals("11:03:05.011+0200", formatted);
+        assertEquals("11:03:05.011+02:00", formatted);
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


tiritea:
test form:

[isodate2.xml.txt](https://github.com/opendatakit/javarosa/files/2933673/isodate2.xml.txt)

tiritea:
will update existing testcases which now throw false negative

tiritea:
not sure what I can/should do about this package error to push it along... Any ideas?

yanokwa:
@ggalmazor Please squash and merge whenever this PR is done. The issue with the broken mvn build is fixed at https://github.com/opendatakit/javarosa/pull/408 and that should be merged after this to save @tiritea, who is using the web UI, a rebasing.

tiritea:
Thank you @ggalmazor for cleaning up my mess (yet again...)  :-)

PR #409, opened 2019-03-07
Register range element into CoreModelModule

### LINKED ISSUE #245: DeserializationException with All_widgets.xml form
#### Software versions 
JavaRosa v2.7.0

#### Problem description
org.javarosa.core.util.externalizable.DeserializationException: No datatype registered to serialization code [d8 60 b9 d1]

Add a test to XFormParserTest:
```
@Test public void allWidgetsFormSavesAndRestores() throws IOException, DeserializationException {
    serAndDeserializeForm(r("All_widgets.xml"));
}
```
#### Expected behavior
Not the exception.
#### Other information 
I’m guessing there’s a class being serialized that isn’t registered with JavaRosa, and that it needs to be added here: https://github.com/opendatakit/javarosa/blob/15a825805a3f5976a908d0e8a1b65c7f8642cf96/src/org/javarosa/core/model/CoreModelModule.java#L26

[All_widgets.xml.txt](https://github.com/opendatakit/javarosa/files/1717676/All_widgets.xml.txt)

mdmahendri:
Closes #245 

run unit test method provided by @dcbriccetti 

Insted of adding into `CoreModelModule` we can add `RangeQuestion` into another classes: `JavaRosaCoreModule` or `XFormsModule`, but only `CoreModelModule` seems relevant for me as `QuestionDef` is also placed there

It should not, as mentioned in opendatakit/collect#2731 only throw warning

yes, i already added the form into `resources` dir. FYI i'm using the `all-widgets.xml` from collect assets not from the issue as it is latest version. Both forms passed the unit test

No

### DIFF
--- a/test/org/javarosa/xform/parse/XFormParserTest.java
+++ b/test/org/javarosa/xform/parse/XFormParserTest.java
@@ -202,6 +202,14 @@ public class XFormParserTest {
         serAndDeserializeForm(r("Simpler_Cascading_Select_Form.xml"));
     }
 
+    /**
+     * ensure serializing and deserializing a range form is done without errors
+     * see https://github.com/opendatakit/javarosa/issues/245 why this is needed
+     */
+    @Test public void rangeFormSavesAndRestores() throws IOException, DeserializationException {
+        serAndDeserializeForm(r("range-form.xml"));
+    }
+
     @Test public void externalSecondaryInstanceFormSavesAndRestores() throws IOException, DeserializationException {
         Path formPath = EXTERNAL_SECONDARY_INSTANCE_XML;
         mapFileToResourcePath(formPath);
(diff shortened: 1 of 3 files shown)


test/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -202,6 +202,10 @@ private void timeParsing(LargeInstanceFileGenerator lfg, Path largeDataFilename,
         serAndDeserializeForm(r("Simpler_Cascading_Select_Form.xml"));
     }
 
+    @Test public void allWidgetsFormSavesAndRestores() throws IOException, DeserializationException {
```
ggalmazor: I'm concerned that people reviewing this won't understand why this test is here and why it's important. Using all-widgets is a good idea, but here we need to make sure one can understand that this is a test that will prevent regression if the RangeWidget somehow disappears from the list of classes for JR serialization.

I'd suggest using a new form with just a range widget and adding a comment inside the test linking to the GitHub issue.

mdmahendri: already done but we lose coverage on serialize/deserialize `OSMTag `, `GroupDef`,  and `SelectChoice` though. I squashed the commit, i assume this is okay?
Thanks for the review

PR #411, opened 2019-03-21
Support reading/writing external parameters used in RangeQuestion

### LINKED ISSUE #410: Parameters from RangeQuestion are not read/written during calling readExternal/writeExternal methods
#### Problem description
Parameters from RangeQuestion are not read/written during calling readExternal/writeExternal methods.

We discovered a problem in Collect: https://github.com/opendatakit/collect/issues/2954 but it's caused by changes in javarosa this pr: https://github.com/opendatakit/javarosa/pull/409
The pr itself is fine but it revealed a long-standing issue in [RangeQuestion](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/core/model/RangeQuestion.java) class. 

The long-standing problem I mentioned is that in the implementation of [RangeQuestion](https://github.com/opendatakit/javarosa/blob/master/src/org/javarosa/core/model/RangeQuestion.java) we don't have reading and writing parameters (rangeStart, rangeEnd, rangeStep), that means after reading a form from cache (in Collect) those values are always null.

Why it didn't cause any problems (those null values) in v1.20.0 and previous versions? Because we didn't have this fix https://github.com/opendatakit/javarosa/pull/409 and without this fix, each form that has at least one RangeQuestion (in Collect) doesn't use cache files, such forms are parsed every time because reading from a cache fails.

grzesiek2010:
Closes #410 

I tested this fix with Collect. Here is an apk with this fix:
[collect-debug-v1.20.0-78-gc103044-dirty.apk.zip](https://github.com/opendatakit/javarosa/files/2990550/collect-debug-v1.20.0-78-gc103044-dirty.apk.zip)

We just need to read/write those parameters to use cache files in Collect.

It's not risky since changes are small. Playing with some forms would be enough here. I implemented changes only for RangeQuestion so we can focus only on that question type.

It would be good to test a case where we have a form used at least once using Collect v1.20.0 and then upgrading to a version with this fix and trying to open the same form again. 

Any form with RangeQuestions: 
[range.xml.txt](https://github.com/opendatakit/javarosa/files/2990530/range.xml.txt)

No.

### DIFF
--- a/src/org/javarosa/core/model/RangeQuestion.java
+++ b/src/org/javarosa/core/model/RangeQuestion.java
@@ -1,5 +1,14 @@
 package org.javarosa.core.model;
 
+import org.javarosa.core.util.externalizable.DeserializationException;
+import org.javarosa.core.util.externalizable.ExtUtil;
+import org.javarosa.core.util.externalizable.ExtWrapNullable;
+import org.javarosa.core.util.externalizable.PrototypeFactory;
+import org.javarosa.xform.parse.RangeParser;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
 import java.math.BigDecimal;
 
 /** A Range-type question, with information pulled from attributes of the range form element */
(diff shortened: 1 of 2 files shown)


lognaturel:
I think we haven't typically done a QA pass on JR pull requests but seems it would be a good idea to do, if only to make sure everyone is aware of the changes. @mmarciniak90, what do you think?

mmarciniak90:
@lognaturel I totally agree.

mmarciniak90:
**Tested with success**

I used an apk attached by Grzegorz.
Verified on Android: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1

Verified cases:
* Jump to range question
* Navigate around range questions
* Fill range question in a new form
* Fill range question in a saved form
* Verified `All widgets` form

lognaturel:
@mmarciniak99 The most important check when there are changes to serialization behavior is opening the form with a prior version so that the cached file is created, upgrading, and then opening the form again to make sure the cached file is regenerated. You didn’t explicitly mention doing this. Could you please either confirm you did or verify it quickly? Thanks!

lognaturel:
Or is that what you meant by “saved form?”

mmarciniak90:
@lognaturel I re-tested issue and checked that cached file is regenerated. It works correctly.
I checked that on Android: 4.2, 4.4, 5.1, 6.0, 7.0, 8.1

'saved form' is a form opened from `Edit Saved Form`.

lognaturel:
Thanks, @mmarciniak90!

PR #413, opened 2019-03-27
Add odk-instance-first-load event and deprecate xforms-ready

### LINKED ISSUE #412: Introduce odk-instance-first-load event and mark xforms-ready as deprecated
The `xforms-ready` event is intended to be dispatched any time the xforms engine is ready. JavaRosa dispatches it only the first time that a record is instantiated.

Instead of continuing to use `xforms-ready`, it will be marked as deprecated and eventually removed, perhaps to be re-introduced with the intended behavior if there's a need for it. The `odk-instance-first-load` event will be dispatched the first time a record is instantiated.

See [forum discussion](https://forum.opendatakit.org/t/spec-proposal-add-first-load-event-to-replace-xforms-ready/18662).

lognaturel:
Closes #412

Changed the test forms to use `odk-instance-first-load` event and verified that tests still pass.

I considered adding a `FormEvent` enum to do some cleanup while I was here but since we'd like to release this very soon I went for the easiest diff to review and verify. I searched the codebase for `"xforms-ready"` and `XFORMS_READY` and made additions and substitutions as needed.

This doesn't change any existing behavior and only adds a new event. I'm not seeing regression risk.

https://github.com/lognaturel/javarosa/blob/b30ce18c876045b3508716ba5f7359ede303af91/resources/form-with-setvalue-action.xml and https://github.com/lognaturel/javarosa/blob/b30ce18c876045b3508716ba5f7359ede303af91/resources/org/javarosa/xform/parse/last-saved-filled.xml are verified by the test suite but it wouldn't be a bad idea to verify in Collect as well.

This only affects XML form definitions so doesn't need to be part of user-facing docs. https://github.com/opendatakit/xforms-spec/issues/223 has been filed.

CC @cooperka

### DIFF
--- a/src/org/javarosa/core/model/FormDef.java
+++ b/src/org/javarosa/core/model/FormDef.java
@@ -1227,10 +1227,11 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
             getLocalizer().setToDefault();
         }
 
-        // TODO: Hm, not 100% sure that this is right. Maybe we should be
-        // using a slightly different event for "First Load" which doesn't
-        // get fired again, but always fire this one?
         if (newInstance) {
+            actionController.triggerActionsFromEvent(Action.EVENT_ODK_INSTANCE_FIRST_LOAD, this);
+
+            // xforms-ready is marked as deprecated as of JavaRosa 2.14.0 but is still dispatched for compatibility with
+            // old form definitions
             actionController.triggerActionsFromEvent(Action.EVENT_XFORMS_READY, this);
         }
 
--- a/resources/form-with-setvalue-action.xml
+++ b/resources/form-with-setvalue-action.xml
@@ -9,7 +9,7 @@
                 </data>
             </instance>
             <bind id="text-data-binding" nodeset="/data/text" type="string"/>
-            <setvalue bind="text-data-binding" event="xforms-ready" >Test Value</setvalue>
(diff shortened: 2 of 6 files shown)


src/org/javarosa/core/model/FormDef.java
```diff
@@ -1227,10 +1227,11 @@ public void initialize(boolean newInstance, InstanceInitializationFactory factor
             getLocalizer().setToDefault();
         }
 
-        // TODO: Hm, not 100% sure that this is right. Maybe we should be
-        // using a slightly different event for "First Load" which doesn't
-        // get fired again, but always fire this one?
         if (newInstance) {
+            actionController.triggerActionsFromEvent(Action.EVENT_ODK_INSTANCE_FIRST_LOAD, this);
```
yanokwa: Are there any performance issues with firing off two events like this?

ggalmazor: The weight of triggering an event grows proportionally with the number of listeners there are subscribed. I don't think this extra event has an impact on JR level.

I'd imagine that if Collect reacts to both, then yes, but since it's being deprecated, we can make sure that Collect only listens to the new event.

lognaturel: The trigger method loops through actions the form defines to be triggered by the event. Presumably a form creator would use one or the other event name in the form definition but not both. If they used both to trigger the same action there'd be some duplication of effort but inexpensive duplication. And that would be very odd form design. So I'd say no.

PR #414, opened 2019-03-27
Upgrade mvn plugins to fix Javadoc errors

yanokwa:
* Ran `mvn clean package` on my local machine to ensure the JavaDoc build works.
* Ran `mvn -s secrets.xml clean deploy` and made sure the snapshot worked (which maybe I should have waited to get this PR approved first, but there's no easy way to test).

We could upgrade to latest point releases, but I figured I'd jump to the latest versions and get that working. If we can't ship to the maven repo for some reason, we'll see this in Collect immediately and roll back.

Useful links: 
* How I knew an upgrade would help: https://github.com/eclipse/xtext-maven/issues/47#issuecomment-416027182
* How I knew about the doclint fix: https://github.com/twitter/hraven/issues/162#issuecomment-433941617

Should have no user facing changes. This is just the scripts that go to the maven repo. Worst case, JavaDoc will be broken, but I'm pretty sure no one uses that.

No.

No.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,7 +86,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.6.1</version>
+        <version>3.8.0</version>
         <configuration>
           <source>1.8</source>
           <target>1.8</target>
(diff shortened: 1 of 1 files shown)


ggalmazor:
I've tried `mvn clean package` and I'm still getting a ton of JavaDoc warnings. The build completes, though. Is this the expected outcome of the changes?

[dpaste with the output of the command](http://dpaste.com/16093DZ)

yanokwa:
Yes, you should get warnings, but not errors. Without the doclint change, the build will fail on those javadoc issues. And without the javadoc plugin upgrades, the build will fail because of "An API incompatibility was encountered while executing". The other plugin upgrades are there because I like to do dangerous things.

PR #416, opened 2019-03-28
Fixed the problem with parsing instances

### LINKED ISSUE #415: External secondary CSV instance support with jr://file-csv connector breaks forms with search() and pulldata()
From: https://github.com/opendatakit/collect/issues/2972

#### Software versions 
JavaRosa - the latest version

[sample_preloading.zip](https://github.com/opendatakit/javarosa/files/3019685/sample_preloading.zip)

#### Problem description
Unexpected error when user opens 'Sample Form - Preloading'

* User opens form 'Sample Form - Preloading'
* Error appears
Screenshot_2019-03-28-17-00-11

#### Expected behavior
* Form opens correctly
* I checked case on v1.20 and it works properly

grzesiek2010:
Closes #415 

I tested the attached form and the form provided by @kkrawczyk123 https://github.com/opendatakit/collect/pull/2882#issuecomment-477233562 (Remembering previously entered values)

The if statement was edited in this pr https://github.com/opendatakit/javarosa/pull/406 and it was a mistake. I think @cooperka can confirm.

This change should just fix the problem with the attached form. I edited the code implemented for remembering previously entered values so it would be good to confirm there is no problem with that functionality as well.

The form attached to the issue and the form provided by @kkrawczyk123 https://github.com/opendatakit/collect/pull/2882#issuecomment-477233562

No.

### DIFF
--- a/src/org/javarosa/xform/parse/XFormParser.java
+++ b/src/org/javarosa/xform/parse/XFormParser.java
@@ -471,7 +471,7 @@ public class XFormParser implements IXFormParserFunctions {
                 final String instanceId = instanceNodeIdStrs.get(instanceIndex);
                 final String instanceSrc = parseInstanceSrc(instance, lastSavedSrc);
 
-                if (instanceSrc != null) {
+                if (instanceSrc != null && instanceSrc.toLowerCase().startsWith("jr://file/")) {
                     final ExternalDataInstance externalDataInstance;
                     try {
                         externalDataInstance = ExternalDataInstance.build(instanceSrc, instanceId);
--- a/src/org/javarosa/xform/parse/XFormParser.java
+++ b/src/org/javarosa/xform/parse/XFormParser.java
@@ -471,6 +471,8 @@ public class XFormParser implements IXFormParserFunctions {
                 final String instanceId = instanceNodeIdStrs.get(instanceIndex);
                 final String instanceSrc = parseInstanceSrc(instance, lastSavedSrc);
 
+                // Disable jr://file-csv/ support by explicitly only supporting jr://file/
+                // until https://github.com/opendatakit/javarosa/issues/417 is addressed
                 if (instanceSrc != null && instanceSrc.toLowerCase().startsWith("jr://file/")) {
                     final ExternalDataInstance externalDataInstance;
                     try {
--- a/test/org/javarosa/xform/parse/XFormParserTest.java
+++ b/test/org/javarosa/xform/parse/XFormParserTest.java
@@ -29,6 +29,7 @@ import org.javarosa.model.xform.XPathReference;
 import org.javarosa.xpath.expr.XPathPathExpr;
 import org.javarosa.xpath.parser.XPathSyntaxException;
 import org.junit.After;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.kxml2.kdom.Element;
 
@@ -180,6 +181,7 @@ public class XFormParserTest {
         assertEquals("external select 10", formDef.getTitle());
     }
 
+    @Ignore("See https://github.com/opendatakit/javarosa/pull/416")
     @Test public void parsesExternalSecondaryInstanceCsvForm() throws IOException {
         Path formName = r("external-select-csv.xml");
         mapFileToResourcePath(formName);


lognaturel:
@grzesiek2010 This looks like it will break last-saved support. Is that something you tried?

I haven't dug very deep yet but the check on a `jr://file` prefix moved to https://github.com/opendatakit/javarosa/pull/406/files#diff-3c2dbe0e3dc62d2d751abeb77add7986R533.

grzesiek2010:
Yes, I tried as I said.
It can't break it because if we use that feature `instanceSrc = jr://file/last-saved.xml` so `startsWith("jr://file/")` retruns TRUE.

>I haven't dug very deep yet but the check on a jr://file prefix moved to 

It was moved there just to return `rawSrc = instance.getAttributeValue(null, "src");` so the same as it was before https://github.com/opendatakit/javarosa/pull/406/files#diff-3c2dbe0e3dc62d2d751abeb77add7986L462

but we still need that additional check https://github.com/opendatakit/javarosa/pull/406/files#diff-3c2dbe0e3dc62d2d751abeb77add7986L464

lognaturel:
Agree with @cooperka. The exception is a deserialization exception. Did you clear the cached file between each trial, @grzesiek2010?

grzesiek2010:
> Did you clear the cached file between each trial

yes, maybe I missed something...

grzesiek2010:
Ok a few last words from me before I go to sleep:
This pr just makes that the code works like it worked before we merged #406
I can confirm that I cleared cache every time and with my change, I can open the form like it was before #406 
I think there is something wrong with that form because I can open it but can's see values from csv files but it works in the same way on older versions, maybe @mmarciniak90 edited the form. I tried the original one I had and then everything is fine https://drive.google.com/file/d/0B3xwpCYFo556eVptNUdyWmRoMmc/view?usp=sharing.

Maybe I missed something I don't really understand your concerns @cooperka but you know the code here better since you are the last editor. 
Unfortunately, I won't be available tomorrow so everything I can do is just sharing these thoughts.

@lognaturel @cooperka if you are 100% sure I was wrong please go ahead and fix it.

cooperka:
The code prior to #406 didn't have the trailing slash that was added here (see [406 diff](https://github.com/opendatakit/javarosa/pull/406/files#diff-3c2dbe0e3dc62d2d751abeb77add7986L464)), so I think this change basically removes support for `file-csv` and the error is fixed because JR doesn't even try to load the CSV.

Thank you for the context @grzesiek2010, I'll see what I can do.

lognaturel:
> so I think this change basically removes support for `file-csv` and the error is fixed because JR doesn't even try to load the CSV

Sort of. This form doesn't actually use the `file-csv` support, it uses the client-only database-driven implementation of `pulldata` so the CSV gets loaded that way. Let's see what else I can figure out!

lognaturel:
I thought that `odk-csv` in the `FileNotFoundException` in the error dialog looked really odd. What's going on is that the `jr://file` prefix is being used to derive the path because no reference factory exists for `file-csv`. When that prefix is removed from `jr://file-csv/hhplotdetails.csv`, we're left with `-csv/hhplotdetails.csv` and a path that doesn't exist.

Regardless, in that test form, we don't actually want to load `hhplotdetails.csv` as an external CSV instance. More details at https://github.com/opendatakit/javarosa/pull/397#issuecomment-477825281. 

All that to say that the problem has nothing to do with #406, it's actually caused by #397. But this change does incidentally work because it disables support for `jr://file-csv`, as @cooperka notes. I've spent a little bit of time trying to think of a quick way to make #397 play nice with the existing external CSV implementations but am coming up short. It seems risky this close to a release anyway.

I'm going to mark this as `needs testing` for @kkrawczyk123 to confirm that it works in all supported external data cases (last saved, external secondary XML instance, search/pulldata like the original form in the issue and http://xlsform.org/en/#external-selects). It won't work with external secondary CSV instances but it looks like those never worked, anyway. Provided all that looks good, I'll add a comment and merge.

kkrawczyk123:
I've checked sample preloading forms, forms with external itemsets, external secondary instances, select one external  - I haven't noticed anything suspicious, any error messages, I was able to finalize forms in all cases. I have also looked for the regression in the last saved submission feature and here nothing suspicious. I hope I didn't miss anything. 
Verified on Androids 6.0 and 4.4.
@opendatakit-bot "behavior verified"

lognaturel:
@ggalmazor, when you are satisfied with this, please merge and re-snapshot from master! 🙏 

Thanks again for the initial legwork, @grzesiek2010. 👍

PR #420, opened 2019-04-02
Prepare version 2.15.0-SNAPSHOT for next dev iteration

yanokwa:
Nothing. Just bumping the version to prepare for the next iteration.


lognaturel:
Shouldn't we wait to merge this until we know we won't need a point release on 2.14?

PR #423, opened 2019-04-12
MicroBenchmarking Contd

JohnTheBeloved:
### What this PR does?
The sole purpose of this PR was to create benchmark for methods which make form loading inefficient and find ways this could be improved

### Major Changes
Added Benchmark for major Flow events when loading and xForm 

- Creating FormDef from Xforms(with **only** **in**ternal secondary instances)
- Creating FormDef from Xforms(with **ex**ternal secondary instances)
- Parsing Secondary Instances (_wards.xml_, _lgas.xml_) sepearately
- Loading External secondary instances
- Answering Questions
- Validating Questions
- Validating Forms
- Traversing Questions on Forms

#### What has been done to verify that this works?
- Dry run benchmark tests to ensure non is failing
- Benchmarks have being run and results attached below
- Spreadsheet results and comparisons can also be found [here](https://docs.google.com/spreadsheets/d/1Qrv8MzGPjm5jTTFBxClJqFxDybM3T1ZI_zg8JdqvKGg/edit#gid=1607988166)

#### Results

Click [here](https://docs.google.com/spreadsheets/d/1Qrv8MzGPjm5jTTFBxClJqFxDybM3T1ZI_zg8JdqvKGg/edit#gid=1607988166) for comparisons 

See below for result snapshot
image

### BenchMark Options Configuration

Warmup Iterations : 10
Measurement Iterations : 10
`BenchmarkTemplate` benchmark excluded


test/org/javarosa/core/TestBenchmark.java
```diff
@@ -0,0 +1,198 @@
...
+    @State(Scope.Thread)
+    public static class ParseWithInternalInstanceState {
+        Path resourcePath = PathConst.getTestResourcePath().toPath();
+
+        Path xFormFilePath = r("nigeria_wards_external_combined.xml");
+        FormDef formDef;
+        @Setup(Level.Trial)
+        public void
+        initialize() {
+            Path resourcePath = PathConst.getTestResourcePath().toPath();
```
ggalmazor: I think it would be more correct and safe here to do:

```java
Path resourcePath = xFormFilePath.getParent();
```

test/org/javarosa/core/TestBenchmark.java
```diff
@@ -0,0 +1,198 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.xmlpull.v1.XmlPullParserException;
+
+
+import static org.javarosa.core.reference.ReferenceManagerTestUtils.buildReferenceFactory;
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.junit.Assert.fail;
+
+public class TestBenchmark {
```
ggalmazor: I'd suggest that we have just one `@Benchmark`. Otherwise, I'm forced to run all the benchmarks in the class, or to change the `OptionsBuilder()`.

test/org/javarosa/core/TestBenchmark.java
```diff
@@ -0,0 +1,198 @@
...
+        state.dataRootNode.populate(state.savedRoot, state.formDef);
+
+  }
+
+    @Benchmark
+    public void
+    benchmarkParseXFormWithExternalInstances(ParseWithExternalInstanceState state, Blackhole bh) {
+        try {
+            bh.consume(FormParserHelper.parse(state.xFormFilePath));
+        } catch (IOException e) {
```
ggalmazor: Could declaring an IOException in the method's signature be enough to make the test fail too? E.g.:

```java
@Benchmark
public void
benchmarkParseXFormWithInternalInstances(ParseWithInternalInstanceState state, Blackhole bh) throws IOException {
    bh.consume(FormParserHelper.parse(state.xFormFilePath));
}
```

test/org/javarosa/core/TestBenchmark.java
```diff
@@ -0,0 +1,198 @@
...
+
+    @State(Scope.Thread)
+    public static class ExternalDataInstanceState {
+        ExternalDataInstance externalDataInstance = null;
+
+        @Setup(Level.Trial)
+        public void
+        initialize() {
+            Path resourcePath = PathConst.getTestResourcePath().toPath();
+            ReferenceManager.instance().addReferenceFactory((buildReferenceFactory("file", resourcePath.toString())));
```
ggalmazor: Since ReferenceManager is a singleton, better use `ReferenceManagerTestUtils.setUpSimpleReferenceManager()` instead to avoid some sources of interference between across different benchmarks.

test/org/javarosa/core/TestBenchmark.java
JohnTheBeloved: @ggalmazor Do you mean on Benchmark File Per use case ?

test/org/javarosa/core/TestBenchmark.java
JohnTheBeloved: I think If the Exception is thrown, The Junit test would pass.

A Runtime Error is thrown in the catch block so as to make the test fail

ggalmazor: No problem. 

I can confirm that it would be enough. Tried this:

```java
@Benchmark
public void
benchmarkPopulateRootNode(BenchmarkState state, Blackhole bh) throws IOException {
    throw new IOException("some error");
}
```

Running this I get a junit test error and the full stacktrace in the output.

test/org/javarosa/core/TestBenchmark.java
ggalmazor: At least we could try this at the beginning. I fear that we might get too many files, but it's really easier to deal and work with benchmarks that way.

ggalmazor:
This is basically right, although I think some things have to be improved yet:
- Use better naming: method names, variables, @State classes must express their intention correctly (what's being benchmarking/testing, what the variable holds, etc.)
- Replace all `PathConst` uses with some relative path to the form we're benchmarking.
  - Use the `r(String filename)` as much as possible.
  - Use `formFile.getParent()` otherwise.
- Don't `try/catch` in tests/benchmarks. Declare exceptions in method signatures instead.
- Separate the benchmarks to different files:
  - One file for the couple of benchmarks around `FormParserHelper.parse()`
  - One file for each of the rest of the benchmarks
  - General rules would be:
    - Ideally, have one @benchmark per file.
    - No more than one @State class.
      - More than one @benchmark could share the same @State class if that makes sense.
- Clean up: remove unused variables/class members/methods and watch the indentation (use the automatic reformat tool of your IDE)

ggalmazor:
I've taken the liberty of applying all the suggestions from my last comment (except the one about having different files) and uploaded a gist with the result: https://gist.github.com/ggalmazor/9448eb30891b963a275a3544499c6162

test/org/javarosa/core/TestBenchmark.java
JohnTheBeloved: Yes, Updated the code

test/org/javarosa/core/benchmark/FormParserHelperParseISIBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+        public void
+        initialize() {
+            Path resourcePath = PathConst.getTestResourcePath().toPath();
+            ReferenceManagerTestUtils.setUpSimpleReferenceManager("file", resourcePath);
+        }
+    }
+
+    @Benchmark
+    public void
+    benchmark_FormParserHelper_parse_internal_secondary_instance(FormParserHelperParseInternalSecondaryInstanceState state, Blackhole bh) throws IOException {
```
ggalmazor: This one doesn't use the @State class. Therefore, we can remove the argument from this method and remove the @State class from this benchmark class.

test/org/javarosa/core/benchmark/FormParserHelperParseISIBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+public class FormParserHelperParseISIBenchmark {
+    private static final Logger logger = LoggerFactory.getLogger(FormParserHelperParseISIBenchmark.class);
+
+    @Test
+    public void
+    launchBenchmark() throws Exception {
+        try{
+            RunResult run = new Runner(BenchmarkUtils.getJVMOptions(this.getClass().getName())).run().iterator().next();
+
+        }catch (NoBenchmarksException nbe){
```
ggalmazor: Why catch this?

test/org/javarosa/core/benchmark/FormParserHelperParseISIBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+
+public class FormParserHelperParseISIBenchmark {
+    private static final Logger logger = LoggerFactory.getLogger(FormParserHelperParseISIBenchmark.class);
```
ggalmazor: Remove the logger since it's not being used.

test/org/javarosa/core/benchmark/FormParserHelperParseISIBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void
+    launchBenchmark() throws Exception {
+        try{
+            RunResult run = new Runner(BenchmarkUtils.getJVMOptions(this.getClass().getName())).run().iterator().next();
```
ggalmazor: The `run` variable is not being used and can be removed.

test/org/javarosa/core/benchmark/FormParserHelperParseESIBenchmark.java
```diff
@@ -0,0 +1,58 @@
...
+            RunResult run = new Runner(BenchmarkUtils.getJVMOptions(this.getClass().getName())).run().iterator().next();
+
+        }catch (NoBenchmarksException nbe){
+
+        }
+    }
+    @State(Scope.Thread)
+    public static class FormParserHelperParseExternalSecondaryInstanceState {
+        Path xFormFilePath = r("nigeria_wards_external.xml");
+        Path resourcePath = xFormFilePath.getParent();
```
ggalmazor: This field is not being used and can be removed

test/org/javarosa/core/benchmark/FormParserHelperParseESIBenchmark.java
```diff
@@ -0,0 +1,58 @@
...
+        FormDef formDef;
```
ggalmazor: This field is not being used and can be removed

test/org/javarosa/core/benchmark/FormParserHelperParseESIBenchmark.java
```diff
@@ -0,0 +1,58 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+
+public class FormParserHelperParseESIBenchmark {
+    private static final Logger logger = LoggerFactory.getLogger(FormParserHelperParseESIBenchmark.class);
```
ggalmazor: The logger is not being used and can be removed

test/org/javarosa/core/benchmark/FormParserHelperParseISIBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+        }catch (NoBenchmarksException nbe){
+
+        }
+    }
+
+
+    @State(Scope.Thread)
+    public static class FormParserHelperParseInternalSecondaryInstanceState {
+        Path xFormFilePath = r("nigeria_wards_external_combined.xml");
+        Path resourcePath = xFormFilePath.getParent();
```
ggalmazor: Not used. Remove.

test/org/javarosa/core/benchmark/FormParserHelperParseISIBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+        FormDef formDef;
```
ggalmazor: Not used. Remove.

test/org/javarosa/core/benchmark/FormParserHelperParseISIBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+        @Setup(Level.Trial)
+        public void
+        initialize() {
+            Path resourcePath = PathConst.getTestResourcePath().toPath();
+            ReferenceManagerTestUtils.setUpSimpleReferenceManager("file", resourcePath);
```
ggalmazor: You should use `resourcePath.getParent()`

test/org/javarosa/core/benchmark/FormParserHelperParseISIBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+import org.openjdk.jmh.runner.Runner;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+
+public class FormParserHelperParseISIBenchmark {
```
ggalmazor: Don't use abbreviations, especially when they hide key aspects of the benchmark i.e. the fact that this benchmark is measuring *internal* secondary instances.

test/org/javarosa/core/benchmark/FormParserHelperParseESIBenchmark.java
```diff
@@ -0,0 +1,58 @@
...
+import org.openjdk.jmh.runner.Runner;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+
+public class FormParserHelperParseESIBenchmark {
```
ggalmazor: Don't use abbreviations, especially when they hide key aspects of the benchmark i.e. the fact that this benchmark is measuring *external* secondary instances.

test/org/javarosa/core/benchmark/FormParserHelperParseESIBenchmark.java
```diff
@@ -0,0 +1,58 @@
...
+    @State(Scope.Thread)
+    public static class FormParserHelperParseExternalSecondaryInstanceState {
+        Path xFormFilePath = r("nigeria_wards_external.xml");
+        Path resourcePath = xFormFilePath.getParent();
+        FormDef formDef;
+        @Setup(Level.Trial)
+        public void
+        initialize() {
+            Path resourcePath = PathConst.getTestResourcePath().toPath();
+            ReferenceManagerTestUtils.setUpSimpleReferenceManager("file", resourcePath);
```
ggalmazor: Use `xFormFilePath.getParent()`  instead of `PathConst`

test/org/javarosa/core/benchmark/ExternalDataInstanceBuildBenchmark.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @State(Scope.Thread)
+    public static class ExternalDataInstanceState {
+        ExternalDataInstance externalDataInstance = null;
+
+        @Setup(Level.Trial)
+        public void
+        initialize() {
+            Path resourcePath = PathConst.getTestResourcePath().toPath();
+            ReferenceManagerTestUtils.setUpSimpleReferenceManager("file", resourcePath);
```
ggalmazor: Use `r("wards.xml").getParent()` instead of the PathConst.

test/org/javarosa/core/benchmark/ExternalDataInstanceBuildBenchmark.java
```diff
@@ -0,0 +1,60 @@
...
+            RunResult run = new Runner(BenchmarkUtils.getJVMOptions(this.getClass().getName())).run().iterator().next();
+
+        }catch (NoBenchmarksException nbe){
+
+        }
+    }
+
+    @State(Scope.Thread)
+    public static class ExternalDataInstanceState {
+        ExternalDataInstance externalDataInstance = null;
```
ggalmazor: Unused. Remove.

test/org/javarosa/core/benchmark/ExternalDataInstanceBuildBenchmark.java
```diff
@@ -0,0 +1,60 @@
...
+import org.openjdk.jmh.runner.Runner;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.xmlpull.v1.XmlPullParserException;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+public class ExternalDataInstanceBuildBenchmark {
+    private static final Logger logger = LoggerFactory.getLogger(ExternalDataInstanceBuildBenchmark.class);
```
ggalmazor: Unused. Remove.

test/org/javarosa/core/benchmark/PopulateTreeNodeBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.junit.Assert.fail;
+
+public class PopulateTreeNodeBenchmark {
+    private static final Logger logger = LoggerFactory.getLogger(FormParserHelperParseESIBenchmark.class);
```
ggalmazor: Unused. Remove.

test/org/javarosa/core/benchmark/PopulateTreeNodeBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    public void
+    launchBenchmark() throws Exception {
+        try{
+            RunResult run = new Runner(BenchmarkUtils.getJVMOptions(this.getClass().getName())).run().iterator().next();
```
ggalmazor: Remove unused variable `run`

test/org/javarosa/core/benchmark/PopulateTreeNodeBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+
+        }catch (NoBenchmarksException nbe){
+
+        }
+
+    }
+
+    @Benchmark
+    public void
+    benchmark_TreeElement_populate(TreeElementPopulateState state, Blackhole bh) throws IOException {
```
ggalmazor: Remove `IOException` from the signature. This method doesn't throw that exception.

test/org/javarosa/core/benchmark/PopulateTreeNodeBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+        private TreeElement savedRoot;
+        private FormDef formDef;
+
+        @Setup(Level.Trial)
+        public void
+        initialize() {
+            FormParseInit formParseInit = new FormParseInit(r("nigeria_wards_external_combined.xml"));
+            FormEntryController formEntryController = formParseInit.getFormEntryController();
+            byte[] formInstanceAsBytes = null;
+            try {
```
ggalmazor: Remove `try/catch` and declare that the method `initialize` throws `IOException`

test/org/javarosa/core/benchmark/PopulateTreeNodeBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+                formInstanceAsBytes = Files.readAllBytes(Paths.get(PathConst.getTestResourcePath().getAbsolutePath(), "populate-nodes-attributes-instance.xml"));
```
ggalmazor: Replace with: 
```java
byte[] formInstanceBytes = r("populate-nodes-attributes-instance.xml");
```

test/org/javarosa/core/benchmark/BenchmarkUtils.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.concurrent.TimeUnit;
+
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.junit.Assert.fail;
+
+
+public class BenchmarkUtils {
```
ggalmazor: Remove all the methods that aren't being used.

test/org/javarosa/core/model/instance/test/SimpleBenchmarkTests.java
```diff
@@ -0,0 +1,104 @@
...
+import java.nio.file.Path;
+import java.util.Collections;
+
+import static org.javarosa.core.reference.ReferenceManagerTestUtils.buildReferenceFactory;
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+
+public class SimpleBenchmarkTests {
```
ggalmazor: Avoid `PathConst` and replace with calls to the `r(String filename)` method.

Also, watch out for the code formatting: some indentations and whitespaces don't follow the code style from the rest of the file.

test/org/javarosa/core/util/PathConst.java
```diff
@@ -1,4 +1,4 @@
-package org.javarosa.core;
+package org.javarosa.core.util;
```
ggalmazor: Why move this class?

ggalmazor:
@JohnTheBeloved, I've removed all the comments that are not really useful at this stage. I forgot this is just an exploration attempt. Apologies!

ggalmazor:
I'm getting some strange results in `FormEntryControllerAnswerQuestion`:
```
Benchmark                                                                     Mode  Cnt      Score   Error  Units
FormEntryControllerAnswerQuestion.benchmark_FormDefValidate_answerAll         avgt    2      5,451          ns/op
FormEntryControllerAnswerQuestion.benchmark_FormDefValidate_answerAndSaveAll  avgt    2      2,554          ns/op
FormEntryControllerAnswerQuestion.benchmark_FormDefValidate_answerOne         avgt    2  72209,750          ns/op
```

Does that make sense to yo, @JohnTheBeloved?

lognaturel:
#429 is now merged but I think it leaves things in a not-ideal state where the benchmarks aren't really reliably runnable because of the configuration. At least, I ran them a handful of times off master and got values different by 1-2 orders of magnitude. On the other hand, I can run them from this PR and get consistent results. I think it would be good to merge this sooner than later. Does that sound right?

Ideally only the new commits would appear but I don't know how cleanly it will rebase on master.

lognaturel:
I've made an executive decision to go ahead and merge this so we can iterate on it through issues and PRs and so that we can be talking about the same thing if we are referring to the results of a particular benchmark.

In my mind, the ideal end state would be benchmarks over different important sections of the code that can be run against a standard set of forms. Ideally when a form is added to the set, all benchmarks are run against it.

PR #429, opened 2019-04-26
Add JMH Gradle plugin

ggalmazor:
I've tried to create self-explaining commits, but here's some extra context:
- make current sourcesets compliant with a maven project structure: src/main/java, src/main/resources, src/test/java, src/test/resources
  - We need this because the JMH plugin creates a new sourceset that complies with the same source/resources directory structure, and working out how to make JMH work with our current sourceset configuration is much harder.
  - Ensure that everything continues working as always
  - It's recommended to recreate the Idea project. **Important** Make sure that the option to "Create separate module per source set" is enabled.
- Update Gradle to 5.1 to use the latest version of the JMH plugin

Here's how we should work now:
- Benchmark classes go to the `src/jmh/main` sourceset
- Benchmark assets (form files, external instance files, submission files, etc.) go to the `src/jmh/resources` sourceset
- Benchmarks that load files must use the `BenchmarkUtils.prepareAssets(String... filenames)` method to prepare a temporary directory where all the provided files will be copied.
  - Paths to individual files can be worked out with `assetsDir.resolve(filename)`
  - The `ResourceManager` can be set up against the `assetsDir`
- `BenchmarkUtils.dryRun(Class clazz)` performs a dry run of the benchmarks defined in a class. No measurement is done. This is just to do quick sanity check to ensure the benchmark payload works correctly without running them through JMH.

Other considerations:
- To run the benchmarks, use the `jmh` Gradle task: `./gradlew jmh`
- **If subsequent runs fail**, stop the Gradle daemon as a workaround with `./gradlew --stop` and try again.
- Disable the IntelliJ Idea JMH Plugin if it's installed

This is a benchmark class template:
```java
package org.javarosa.benchmarks;

import static java.nio.file.Files.readAllBytes;
import static org.javarosa.benchmarks.BenchmarkUtils.dryRun;
import static org.javarosa.benchmarks.BenchmarkUtils.prepareAssets;
import static org.javarosa.core.reference.ReferenceManagerTestUtils.setUpSimpleReferenceManager;

import java.io.IOException;
import java.nio.file.Path;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Level;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.infra.Blackhole;

public class BenchmarkTemplate {
    public static void main(String[] args) {
        // You can dry-run all the benchmarks in this class with BenchmarkUtils.dryRun()
        // This won't take any measure.
        // It'll just run the benchmark methods to ensure everything works as intended.
        // It will let you debug the benchmark code too.
        dryRun(BenchmarkTemplate.class);
    }

    @State(Scope.Thread)
    public static class StateClassTemplate {
        Path formFile;

        @Setup(Level.Trial)
        public void setUp() {
            // When the benchmark needs files from src/jmh/resources, you need to call BenchmarkUtils.prepareAssets()
            // to take them out of the JAR file JMH uses into a temporary directory that JavaRosa can work with.
            Path assetsPath = prepareAssets("some-file.xml");
            // Once you do so, get paths to any individual file using Path.resolve
            formFile = assetsPath.resolve("some-file.xml");
            // If your forms load external files, you'll need to prime the ResourceManager using the assetsDir
            // returned by BenchmarkUtils.prepareAssets(). You need to make sure that all external files in the
            // form are like jr://file/filename (all use the "file" hostname, all are located at the root).
            setUpSimpleReferenceManager("file", assetsPath);
        }
    }

    @Benchmark
    public void benchmark_method_template(StateClassTemplate state, Blackhole bh) throws IOException {
        // Always use Blackhole.consume with the output of the benchmarked action
        bh.consume(readAllBytes(state.formFile));
    }
}
```


lognaturel:
Did you consider rolling your own as in https://blog.morethan.io/jmh-with-gradle-from-easy-to-simple-dc872d57cf7f or an alternate plugin such as https://github.com/blackboard/jmh-gradle-plugin (less popular but could still be better in some way)? Why do you think this plugin is the best approach?

It's really too bad that everything needs to move. In this case the change is not "better" in an absolute sense (it doesn't, for example, provide more clarity) and it makes archaeology harder because Github doesn't follow moves (ARG. https://github.com/isaacs/github/issues/900).

ggalmazor:
We discussed @lognaturel's comment on Slack, and here's the excerpt:

> Did you consider rolling your own <JMH integration blahblah>?

Yes! I followed the guide you've linked and a couple more, but doing everything that the plugin does would bring a lot of complexity to our build file

Side note: What *does* the plugin? I think I only understand half of what's going on, which involves:
- Creating a new jmh sourceset that depends on the JMH library
- Making the jmh sourceset also depend on the main sourceset and, conditionally, on the test sourceset too.
- Setting up magic Gradle tasks that generate benchmark classes/bytecode stuff?, put everything into a new JAR file, and run it << This is the part that I most don't understand.
- And everything in such a way that IntelliJ doesn't go crazy and actually works (editado)

> Did you consider other plugins?

I browsed a few of other options but went to the one my (expert) colleague suggested, which is the one that he uses frequently.

Other plugins/tools don't look as updated as the one I chose, and it has the benefit that I can ask questions about it to my colleague :)

<hr/>

> Yes, makes sense. Just want to make sure we're in sync on the goals. So they (the benchmarks) do need to be runnable independent of the IDE for collecting results and they also need to be runnable in the IDE for development. :+1: 

<hr/>

> It's really too bad that everything needs to move

Yes! We're hoping GitHub implements tracking movements of files :)

ggalmazor:
I've solved the conflicts on this PR. Now we can wait for the feedback from @JohnTheBeloved and others using these changes before merging the PR.

ggalmazor:
It looks like we should merge this PR to make #423 easier to merge and to resolve conflicts while #431 and #437 are still drafts.

src/jmh/resources/nigeria_wards_external_combined.xml
```diff
@@ -0,0 +1,8220 @@
+<?xml version="1.0"?>
```
lognaturel: Calling this `external_combined` seems really odd to me -- seems `internal` would be a better suffix!

PR #431, opened 2019-04-30
Setlocation action

### LINKED ISSUE #430: Add support for odk:setlocation action
As described in https://forum.opendatakit.org/t/auto-gps-implementation-on-odk-collect/10635 and https://forum.opendatakit.org/t/spec-addition-proposal-location-preload/10768

The JavaRosa implementation must make it so that clients that read but don't write forms such as Aggregate and Validate recognize the element. It must leave it up to clients that write form instance data to provide an implementation for fetching location.

lognaturel:
Closes #430 

@ggalmazor would be great to get a sanity check on the approach and the question about tests before I go deeper in the Collect implementation. 🙏 

Ran tests with forms. Started a Collect implementation at https://github.com/opendatakit/collect/compare/master...lognaturel:setlocation-action (the location-getting part works but there's still work to do to notify the user and allow them to override background GPS).

I first considered doing everything in Collect since `registerActionHandler` makes that easily possible. However, that would mean any client that uses JavaRosa would need to register its own handler. This would lead to duplicate code and likelihood that the action wouldn't be recognized somewhere it should.

Providing a stub implementation at the JavaRosa level means Briefcase, Validate, Aggregate and any other client will easily be able to read form definitions that include the action. Clients like Collect that need to actually populate form instances with locations can override the `requestLocationUpdates` method. They can use the `saveLocationValue` method so that they don't need to know about messy JR implementation details. The `getSetLocationAction` factory method lets clients simply build handlers for their custom implementations.

I also considered subclassing `SetValueAction` because arguably location is just a specialized value we're setting. I chose not to do that because I don't understand why `parseSetValueAction` is in `XFormParser` rather than in the action definition and the code in that method for supporting `bind` and predicates looks smelly to me. I decided to go for a simpler (but similar) implementation. That means predicates aren't supported but they won't be in XLSForm anyway and while I can think of uses in the setvalue case, I can't really in this one.

This is the first namespaced action to be introduced. I considered changing the key for the `actionHandlers` map to include namespace information but that ended up feeling like a heavy change. Instead, I throw a parse exception if the incorrect namespace is provided. This approach would only be an issue if there's a future need to introduce actions with matching names in different namespaces. This would be a really unfortunate state of affairs and very unlikely so I think this approach is sufficient.

Clients that get location updates may need to let users know about it in the user interface. That means they need to be able to identify form definitions that include the `odk:setlocation` action. 
10bc803 is the best way I could come up with for doing that.

I'm not sure about how I've structured the tests and have some questions inline for @ggalmazor.

This is an additive change that is isolated in new classes so regression risks should be nil (for real!). The only affect on users should be enabling use of the `odk:setlocation` action. 

In addition to the forms in the tests, I used [setlocation-action-with-repeats.xml.zip](https://github.com/opendatakit/javarosa/files/3129864/setlocation-action-with-repeats.xml.zip)

https://github.com/opendatakit/xforms-spec/issues/172

### DIFF
--- a/src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
+++ b/src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2019 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.javarosa.core.model.actions.setlocation;
+
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.actions.Action;
+import org.javarosa.core.model.condition.EvaluationContext;
+import org.javarosa.core.model.condition.Recalculate;
+import org.javarosa.core.model.data.AnswerDataFactory;
+import org.javarosa.core.model.data.IAnswerData;
+import org.javarosa.core.model.instance.AbstractTreeElement;
+import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.util.externalizable.DeserializationException;
+import org.javarosa.core.util.externalizable.ExtUtil;
+import org.javarosa.core.util.externalizable.ExtWrapNullable;
+import org.javarosa.core.util.externalizable.PrototypeFactory;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+
+/**
+ * Abstract implementation of odk:setlocation action. Concrete implementations must:
+ * - provide a way to request location updates when the action is triggered
+ * - use {@link #saveLocationValue(String)} to write a location to the model
+ * - provide a no-argument constructor with no body for serialization
+ * - get registered by the {@link org.javarosa.core.services.PrototypeManager}
+ */
+public abstract class SetLocationAction extends Action {
+    private TreeReference targetReference;
+    private TreeReference contextualizedTargetReference;
+
+    private FormDef formDef;
+
+    public SetLocationAction() {
+        // empty body for serialization
+    }
+
+    SetLocationAction(TreeReference targetReference) {
+        super(SetLocationActionHandler.ELEMENT_NAME);
+        setTargetReference(targetReference);
+    }
+
+    public TreeReference getTargetReference() {
+        return targetReference;
+    }
+
+    public void setTargetReference(TreeReference targetReference) {
+        this.targetReference = targetReference;
+    }
+
+    @Override
+    public final TreeReference processAction(FormDef model, TreeReference contextRef) {
+        this.formDef = model;
+        contextualizedTargetReference = contextRef == null ? this.targetReference
+            : this.targetReference.contextualize(contextRef);
+
+        requestLocationUpdates();
+
+        return contextualizedTargetReference;
+    }
+
+    /**
+     * Client-specific location request. Implementations could immediately read a location and write it to the model if
+     * available or could initiate a request and asynchronously write the location when available.
+     */
+    public abstract void requestLocationUpdates();
+
+    /**
+     * Save the location to the model.
+     */
+    public final void saveLocationValue(String location) {
+        EvaluationContext context = new EvaluationContext(formDef.getEvaluationContext(), contextualizedTargetReference);
+        AbstractTreeElement node = context.resolveReference(contextualizedTargetReference);
+        if (node != null) {
+            int dataType = node.getDataType();
+            IAnswerData val = Recalculate.wrapData(location, dataType);
+            if (val == null) {
+                formDef.setValue(null, contextualizedTargetReference, true);
+            } else {
+                IAnswerData answer = AnswerDataFactory.templateByDataType(dataType).cast(val.uncast());
+                formDef.setValue(answer, contextualizedTargetReference, true);
+            }
+        }
+    }
+
+    //region serialization
+    @Override
+    public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
+        super.readExternal(in, pf);
+
+        targetReference = (TreeReference) ExtUtil.read(in, new ExtWrapNullable(TreeReference.class), pf);
+    }
+
+    @Override
+    public void writeExternal(DataOutputStream out) throws IOException {
+        super.writeExternal(out);
+
+        ExtUtil.write(out, new ExtWrapNullable(targetReference));
+    }
+    //endregion
+}
--- a/src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
+++ b/src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
@@ -93,12 +93,11 @@ public abstract class SetLocationAction extends Action {
         if (node != null) {
             int dataType = node.getDataType();
             IAnswerData val = Recalculate.wrapData(location, dataType);
-            if (val == null) {
-                formDef.setValue(null, contextualizedTargetReference, true);
-            } else {
-                IAnswerData answer = AnswerDataFactory.templateByDataType(dataType).cast(val.uncast());
-                formDef.setValue(answer, contextualizedTargetReference, true);
-            }
+
+            IAnswerData answer = val != null
+                ? AnswerDataFactory.templateByDataType(dataType).cast(val.uncast())
+                : null;
+            formDef.setValue(answer, contextualizedTargetReference, true);
         }
     }
 
--- a/src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
+++ b/src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
@@ -50,13 +50,17 @@ public abstract class SetLocationAction extends Action {
         // empty body for serialization
     }
 
-    SetLocationAction(TreeReference targetReference) {
+    public SetLocationAction(TreeReference targetReference) {
         super(SetLocationActionHandler.ELEMENT_NAME);
         setTargetReference(targetReference);
     }
 
     public TreeReference getTargetReference() {
         return targetReference;
+    }
+
+    public TreeReference getContextualizedTargetReference() {
+        return contextualizedTargetReference;
     }
 
     public void setTargetReference(TreeReference targetReference) {
--- a/src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationActionHandler.java
+++ b/src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationActionHandler.java
@@ -36,7 +36,6 @@ public abstract class SetLocationActionHandler implements IElementHandler {
             throw new XFormParseException("setlocation action must be in http://www.opendatakit.org/xforms namespace");
         }
 
-        TreeReference target;
         String ref = e.getAttributeValue(null, "ref");
 
         if (ref == null) {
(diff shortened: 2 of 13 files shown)


test/org/javarosa/core/model/actions/SetLocationActionTest.java
```diff
@@ -0,0 +1,140 @@
...
+        final FormDef formDef = parse(r("setlocation-action-instance-load.xml"));
+        formDef.initialize(true, new InstanceInitializationFactory());
+        TreeReference targetRef = createTargetRef(null);
+
+        assertEquals(SetLocationAction.NO_IMPLEMENTATION_MESSAGE, getValueAt(formDef, targetRef));
+    }
+
+    @Test
+    public void when_triggerNodeIsUpdated_locationIsSetAtTarget() throws IOException {
+        // Given
```
ggalmazor: Would you consider doing this?

```java
@Test
public void when_triggerNodeIsUpdated_locationIsSetAtTarget() {
    Scenario scenario = Scenario.init(r("setlocation-action-value-changed.xml"));

    // Answering a question triggers an "xforms-value-changed" event on it
    scenario.answer("/data/text", "some answer");

    assertThat(
        scenario.answerOf("/data/location"),
        is(stringAnswer(SetLocationAction.NO_IMPLEMENTATION_MESSAGE))
    );
}
```
(the `Scenario` class would only require a couple of small changes)

lognaturel:
Aha, thank you, @ggalmazor for all the great suggestions! I think I agree with all. I'm going to push a little further on the Collect implementation because I think I might have to add a little bit more here and then bring in your changes.

Agreed in particular with moving to a standard way of doing tests. You probably saw my todo about that. I'm fine with using `Scenario` but I think it should probably be unburied in addition to made public if that's going to be the default way we write moving forward. Maybe moving it to `javarosa.core.test` alongside `FormParseInit`? Would you be willing to do that in a quick PR when you have a moment (hah haha!).

test/org/javarosa/core/model/actions/SetLocationActionTest.java
```diff
@@ -0,0 +1,140 @@
...
+        // When
+        formEntryController.stepToNextEvent();
+        formEntryController.answerQuestion(new StringData("Foo"), true);
+
+        // Then
+        assertEquals(SetLocationAction.NO_IMPLEMENTATION_MESSAGE, getValueAt(formDef, targetRef));
+    }
+
+    @Test
+    public void testSerializationAndDeserialization() throws IOException, DeserializationException {
```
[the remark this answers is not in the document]
lognaturel: Yeaah, I had some copy pasta problems, sorry. 🍝 😞

ggalmazor:
Working on the testing PR!

src/org/javarosa/core/model/actions/SetLocationAction.java
```diff
@@ -0,0 +1,110 @@
...
+    public SetLocationAction(TreeReference targetReference) {
+        super(ELEMENT_NAME);
+        setTargetReference(targetReference);
+    }
+
+    void setTargetReference(TreeReference targetReference) {
+        this.targetReference = targetReference;
+    }
+
+    public final TreeReference processAction(FormDef model, TreeReference contextRef) {
```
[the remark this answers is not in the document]
lognaturel: Indeed!

src/org/javarosa/core/model/actions/SetLocationActionHandler.java
```diff
@@ -0,0 +1,66 @@
...
+import org.javarosa.model.xform.XPathReference;
+import org.javarosa.xform.parse.IElementHandler;
+import org.javarosa.xform.parse.XFormParseException;
+import org.javarosa.xform.parse.XFormParser;
+import org.kxml2.kdom.Element;
+
+public class SetLocationActionHandler implements IElementHandler {
+    @Override
+    public final void handle(XFormParser p, Element e, Object parent) {
+        if (!e.getNamespace().equals(XFormParser.NAMESPACE_ODK)) {
```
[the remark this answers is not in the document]
lognaturel: 🎉

test/org/javarosa/core/model/actions/SetLocationActionTest.java
```diff
@@ -0,0 +1,140 @@
...
+    public void when_instanceIsLoaded_locationIsSetAtTarget() throws IOException {
+        final FormDef formDef = parse(r("setlocation-action-instance-load.xml"));
+        formDef.initialize(true, new InstanceInitializationFactory());
+        TreeReference targetRef = createTargetRef(null);
+
+        assertEquals(SetLocationAction.NO_IMPLEMENTATION_MESSAGE, getValueAt(formDef, targetRef));
+    }
+
+    @Test
+    public void when_triggerNodeIsUpdated_locationIsSetAtTarget() throws IOException {
```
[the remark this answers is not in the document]
lognaturel: Thanks, I brought your `Scenario` commit in and then pulled most of the changes from your demo branch in.

src/org/javarosa/core/model/actions/SetLocationActionHandler.java
```diff
@@ -0,0 +1,66 @@
...
+            throw new XFormParseException("setlocation action must be in http://www.opendatakit.org/xforms namespace");
+        }
+
+        String event = e.getAttributeValue(null, "event");
+
+        TreeReference target;
+        String ref = e.getAttributeValue(null, "ref");
+        if (ref != null) {
```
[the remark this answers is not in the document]
lognaturel: Yes, that's definitely nicer!

src/org/javarosa/core/model/actions/SetLocationAction.java
```diff
@@ -0,0 +1,110 @@
...
+        this.formDef = model;
+        contextualizedTargetReference = contextRef == null ? this.targetReference
+            : this.targetReference.contextualize(contextRef);
+
+        requestLocationUpdates();
+
+        return contextualizedTargetReference;
+    }
+
+    public void requestLocationUpdates() {
```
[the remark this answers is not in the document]
lognaturel: This is what clients will need to implement. Using the abstract classes you've suggested hopefully makes that clearer and I've also added some clarifying comments.

src/org/javarosa/core/model/actions/SetLocationActionHandler.java
```diff
@@ -0,0 +1,66 @@
...
+import org.kxml2.kdom.Element;
+
+public class SetLocationActionHandler implements IElementHandler {
+    @Override
+    public final void handle(XFormParser p, Element e, Object parent) {
+        if (!e.getNamespace().equals(XFormParser.NAMESPACE_ODK)) {
+            throw new XFormParseException("setlocation action must be in http://www.opendatakit.org/xforms namespace");
+        }
+
+        String event = e.getAttributeValue(null, "event");
```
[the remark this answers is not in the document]
lognaturel: Done, thanks.

ggalmazor:
@lognaturel, I see you added some new changes. Let me know if you want me to do a new review ;)

lognaturel:
👀 You see me! I want to do a little more on the Collect side before calling this ready. Will let you know when it's ready! 😊

lognaturel:
Ready when you are, @ggalmazor.

src/test/java/org/javarosa/core/test/Scenario.java
```diff
@@ -97,8 +98,12 @@ private Scenario(FormDef formDef, FormEntryController formEntryController) {
      * Creates and prepares the test scenario loading and parsing the given form
      */
     public static Scenario init(String formFileName) {
+        return init(r(formFileName));
+    }
+
+    public static Scenario init(Path formFile) {
```
ggalmazor: Thanks! This class is much better now!

src/test/java/org/javarosa/core/model/actions/SetLocationActionTest.java
```diff
@@ -0,0 +1,93 @@
...
+    @Test
+    public void testSerializationAndDeserialization() throws IOException, DeserializationException {
+        StubSetLocationAction originalAction = new StubSetLocationAction(absoluteRef("/data/text"));
+
+        Path ser = Files.createTempFile("serialized-object", null);
+        try (DataOutputStream dos = new DataOutputStream(Files.newOutputStream(ser))) {
+            originalAction.writeExternal(dos);
+        }
+
+        StubSetLocationAction deserializedAction = new StubSetLocationAction(null);
```
ggalmazor: Silly detail: `deserializedAction` could be of `SetLocationAction` type

src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationActionHandler.java
```diff
@@ -0,0 +1,62 @@
...
+public abstract class SetLocationActionHandler implements IElementHandler {
+    public static final String ELEMENT_NAME = "setlocation";
+
+    @Override
+    public final void handle(XFormParser p, Element e, Object parent) {
+        if (!e.getNamespace().equals(XFormParser.NAMESPACE_ODK)) {
+            throw new XFormParseException("setlocation action must be in http://www.opendatakit.org/xforms namespace");
+        }
+
+        TreeReference target;
```
ggalmazor: Silly detail: You could inline this declaration at line 47, where the variable gets a value because no one is using it before that.

src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Save the location to the model.
+     */
+    public final void saveLocationValue(String location) {
+        EvaluationContext context = new EvaluationContext(formDef.getEvaluationContext(), contextualizedTargetReference);
+        AbstractTreeElement node = context.resolveReference(contextualizedTargetReference);
+        if (node != null) {
+            int dataType = node.getDataType();
+            IAnswerData val = Recalculate.wrapData(location, dataType);
+            if (val == null) {
```
ggalmazor: To avoid conditional logic, you could do this instead:
```java
IAnswerData answer = val != null
    ? AnswerDataFactory.templateByDataType(dataType).cast(val.uncast())
    : null;
formDef.setValue(answer, contextualizedTargetReference, true);
```

This way we explicitly separate the moving part (conditionally obtaining the answer) from the fixed part (setting the value in the form def)

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -696,6 +702,11 @@ private void parseAction(Element e, Object parent, IElementHandler specificHandl
             throw new XFormParseException("An action element occurred in an invalid location. " +
                     "Must be either a child of a control element, or a child of the <model>");
         }
+        
+        if (e.getName().equals(SetLocationActionHandler.ELEMENT_NAME)) {
```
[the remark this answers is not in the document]
lognaturel: Ooh, yes! I was uncomfortable with that too. 7da74bb

src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
lognaturel: You and @dcbriccetti really love your ternary! 😊 I see the appeal in this case. Done.

src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationActionHandler.java
lognaturel: Whoops, things moved around. Thanks.

src/test/java/org/javarosa/core/model/actions/SetLocationActionTest.java
lognaturel: 👍

src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
dcbriccetti: Though I think the ? belongs on the same line as the conditional expression, as in English. But if everybody were like me the world would be great^h^h^h^h^hboring.
:-)

lognaturel: 😂Nice to see some `^H^H^H`.

lognaturel:
Thanks so much for the great tips, @ggalmazor. I've made the changes and verified this in Collect. I'll let you take a last look and do the merge if satisfied.

src/main/java/org/javarosa/core/model/actions/setlocation/SetLocationAction.java
ggalmazor: Today I learned something new :) `^h`

PR #432, opened 2019-05-01
Add SelectChoices equals and hashCode

### LINKED ISSUE #426: Implement equals for SelectChoice
See https://github.com/opendatakit/collect/pull/3023/files#diff-31b495e34e6fbe9464cec0d2e0025770R120

lognaturel:
Closes #426

Added the https://jqno.nl/equalsverifier/ dependency and ran the verifier. It does flag two issues: the fields are not final and the class is not final. Note in particular that `index` does get changed.

Now that I've done it, I don't know if this is a good idea, honestly. The fact that `index` mutates makes things very odd comparison-wise. It's ok for the use I want to make of it at https://github.com/opendatakit/collect/pull/3023/files#diff-31b495e34e6fbe9464cec0d2e0025770R120 but it would be bad if these were put in sets or used as indexes for maps. Then again, that's not terribly helpful anyway. A better design might have been immutable options with another layer that keeps track of order.

What do reviewers think -- should I just keep comparison in Collect for now?

That aside, the approach is idiomatic, leverages tools the language makes available and is simple to understand. I could probably do something fancier with `hashCode` but it's not necessary. I hesitated between `other == null || getClass() != other.getClass()` and using `instanceOf`. Doing it this way may help remind us there may be issues if ever we subclass.

Clients will be able to identify equal `SelectChoices`. It's imperfect for the reasons listed above.

No.

No.

### DIFF
--- a/src/org/javarosa/core/model/SelectChoice.java
+++ b/src/org/javarosa/core/model/SelectChoice.java
@@ -3,6 +3,7 @@ package org.javarosa.core.model;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.Objects;
 
 import org.javarosa.core.model.data.helper.Selection;
 import org.javarosa.core.model.instance.TreeElement;
@@ -132,4 +133,27 @@ public class SelectChoice implements Externalizable, Localizable {
         this.textID = textID;
     }
 
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
+
+        if (other == null || getClass() != other.getClass()) {
+            return false;
+        }
+        SelectChoice otherChoice = (SelectChoice) other;
+
+        return Objects.equals(labelInnerText, otherChoice.labelInnerText)
+            && Objects.equals(textID, otherChoice.textID)
+            && isLocalizable == otherChoice.isLocalizable
+            && Objects.equals(value, otherChoice.value)
+            && index == otherChoice.index
+            && Objects.equals(copyNode, otherChoice.copyNode);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(labelInnerText, textID, isLocalizable, value, index, copyNode);
+    }
 }


src/org/javarosa/core/model/SelectChoice.java
```diff
@@ -132,4 +133,27 @@ public void setTextID(String textID) {
...
+            return false;
+        }
+        SelectChoice otherChoice = (SelectChoice) other;
+
+        return Objects.equals(labelInnerText, otherChoice.labelInnerText)
+            && Objects.equals(textID, otherChoice.textID)
+            && isLocalizable == otherChoice.isLocalizable
+            && Objects.equals(value, otherChoice.value)
+            && index == otherChoice.index
+            && copyNode == otherChoice.copyNode;
```
ggalmazor: I always use IntelliJ's automatic generator of equals and hash methods. The only difference I see is that IntelliJ suggests `Objects.equals(copyNode, otherChoice.copyNode)`. Are you sure you want to compare the `copyNode` by reference and not by value?

lognaturel: I made the change and pushed it but I don't even know, really. `TreeElement`s aren't meaningfully comparable because again, they're mutable.

lognaturel:
Thanks, @ggalmazor! What do you think about my high-level question about whether we should even add this functionality at all given that the class is mutable? To be explicit, let's say `SelectChoice` is used as the key to a map, if choices get moved around, then the `index` values change which means the `hashCode` changes and it may no longer be possible to retrieve certain things from a map. Equality is somewhat less bad -- it makes sense to only consider two choices equal if they are at the same location.

src/org/javarosa/core/model/SelectChoice.java
ggalmazor: I think it's OK: using Object.equals is safer regarding null values (not super important in this particular case), and equality rules are what we need unless we implement `TreeElement.equals()` (which should happen only if we make it immutable or implement an `equals` method we're happy about)

ggalmazor:
Ugh! Free mutable state is a huge pain! :P

Would it make sense to ignore the `index` member in `equals` and `hash`?

If not (I'm guessing not), I agree with you and I think we shouldn't implement `equals` and `hash` for this one while it's mutable.

ggalmazor:
Please, see my other comment https://github.com/opendatakit/javarosa/pull/433#issuecomment-488607027

lognaturel:
> Would it make sense to ignore the index member in equals and hash?

I don't think so. At some point we can look at separating the choice concept from its position.

I think this one shouldn't be done at all.

PR #433, opened 2019-05-01
Add FormIndex hashCode method

### LINKED ISSUE #425: Add hashCode to FormIndex
`FormIndex` is used as a key in some maps in Collect. It doesn't have a `hashCode` method so that means keys only match if they are references to the same object.

lognaturel:
Closes #425 

Added the https://jqno.nl/equalsverifier/ dependency and ran the verifier. There are issues with fields being mutable. It's not ideal but the intended usage is for these indexes to be built once and then used. Also, beginningOfForm and endOfForm are used independently of the other fields to determine equality. That is, you could have two `FormIndex` objects with different localIndex values that are considered the same because they both have `beginningOfForm` as true. In practice, as long as clients always use `createBeginningOfFormIndex` and `createEndOfFormIndex`, everything should be consistent.

It's simple to understand and verify.

I considered removing the `beginningOfForm` and `endOfForm` fields and replacing them with special values for `localIndex` but then this change becomes a lot more risky with no real benefit in the short term.

Clients can more correctly use these as map indexes, as Collect already does. There would be problems if they get mutated, though.

No.

No.

### DIFF
--- a/src/org/javarosa/core/model/FormIndex.java
+++ b/src/org/javarosa/core/model/FormIndex.java
@@ -19,6 +19,7 @@ package org.javarosa.core.model;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import org.javarosa.core.model.instance.TreeReference;
 
@@ -113,7 +114,8 @@ public class FormIndex implements Serializable {
     private FormIndex nextLevel;
 
     /**
-     * The XPath reference this index refers to.
+     * The XPath reference this index refers to. Warning: these are mutable and could conceivably get out of sync with
+     * the index.
      */
     private TreeReference reference;
 
@@ -292,6 +294,7 @@ public class FormIndex implements Serializable {
         return beginningOfForm;
     }
 
+    @Override
     public boolean equals(Object o) {
         if(!(o instanceof FormIndex))
             return false;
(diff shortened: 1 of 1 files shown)


src/org/javarosa/core/model/FormIndex.java
```diff
@@ -90,7 +91,7 @@
  * @author Clayton Sims
  *
  */
-public class FormIndex implements Serializable {
+public final class FormIndex implements Serializable {
```
ggalmazor: Now I realize that this change could break downstream client code cause this class is public. I'm OK to set it `final`, but I just wanted to point it out.

lognaturel: Oh yes, I should have mentioned that. It does make `equals` a little more correct in the sense that it doesn't have to deal with subclasses. But really, let's not bother doing that here. I've removed it and will let you merge if you're otherwise satisfied.

ggalmazor:
The more I dig into how the state mutates in SelectChoice and FormIndex, the more I get blocked about these PRs... 

I don't see a way to remove state mutation from JavaRosa and it doesn't make (semantic) sense to me to have equals and hash methods on mutable state objects. I think we should either:
- Not add them, and let third parties decide how they understand equality
  Collect might be OK to ignore  `FormIndex.reference` mutation, but other apps could require other criteria.
- Add them, but include a comment to warn users about the ins&outs for future reference

lognaturel:
> The more I dig into how the state mutates in SelectChoice and FormIndex, the more I get blocked about these PRs...

Yeah, seriously. I had left these for when I had a quick moment and not too much brain power left. Hah. Good thing I didn't mark them as `good first issue`! `equals` and `hashCode` are never as simple as they look but this is particularly annoying.

Collect already uses `FormIndex` as a map key. It incidentally works as used because for that usage the same objects are always the same (unless there is sometimes incorrect behavior we don't know about). But I got bit when I tried to do the same in a context where different objects representing the same index had to be used. This makes me want to add `hashCode`, especially since `equals` is there and in use.

How about excluding `reference` from `hashCode`? Then at least it matches the `equals` implementation and better matches what these indexes are supposed to represent. I've pushed yet another change for you to consider. In particular, please let me know if any of the wording is unclear or imprecise.

ggalmazor:
I think we have covered our bases enough and we can merge this one now :)

PR #434, opened 2019-05-03
Change the test sourceset so any test can use Scenario

ggalmazor:
This PR only changes the test sourceset following conversation [here](https://github.com/opendatakit/javarosa/pull/433#issuecomment-488750229). I've verified that all the tests continue passing after making the changes.

The changes to `ResourcePathHelper` could be controversial. Check the commit message at 920a8c5. A non-caching solution could be done as well, but it would probably require changing tests.

This PR should not affect users at all.

Nope.

No.


lognaturel:
I think there's a comment that needs updating but other than that, looks good to me! I can't think of a great alternative to the cache-based solution so let's roll with it for now. It's not a big deal to change in the future.

Issue #436, opened 2019-05-10
Changes to cascading select only propagate one level


lognaturel:
#### Software versions 
JavaRosa v2.14.0 and previous

#### Problem description
"Cascading selects" are a series of selects in which the previous select's answer is used to filter the choices in the next select. If the value of a select part of such a cascade is updated, it only updates the choices for selects that directly depend on its value and not for selects that are two or more layers away. Ironically, the change doesn't cascade. 😅

#### Steps to reproduce the problem
See https://github.com/opendatakit/javarosa/compare/master...lognaturel:multilevel-cascading-select#diff-679ff01315a35ab7db877c8cb9a86f32R69 and note that there's nothing special about "clearing", setting any value fails to propagate.

#### Expected behavior
The choices of all dependent selects should update.

lognaturel:
The root of the issue is that the next level's value doesn't change. That is, if there are three levels with values a, a1 and a1a, when the first level gets cleared, a1 and a1a stay as answers. They get cleared incidentally later when they're not part of the available choices.

lognaturel:
I think the first step is something like https://github.com/opendatakit/javarosa/commit/cac1a2b3ee91a18992df118958b07f0cb3682a7d -- if the question's value is no longer part of the choices, clear the value. However, that's not enough because the choice filters are not part of the recomputation of dependent calculations at all.

The change I've made makes the tests pass but that's because `choicesOf` rebuilds the choice list no matter what. What we'd want is for a change at the Nth level of the cascade to trigger an update at the (N+1)st level and in the absence of a change at the (N+1)st level, for the Nth level not to recompute the choices during a form filling session. The last part potentially has significant performance ramifications. That is, if there is a question that needs to filter down a big list of selections, that filter needs to be applied each time the question is revisited rather than just once when a parent select's value changes.

`IDag` and implementing classes along with `Triggerable` govern these updates. I don't think I can prioritize this now but hopefully we can come back to it soon.

lognaturel:
See additional context and ideas at https://github.com/opendatakit/javarosa/pull/437#discussion_r284443873 and https://github.com/opendatakit/javarosa/pull/437#discussion_r284447606

lognaturel:
I was thinking about this again in the context of #486 and want to make a note that when declaring an `itemset`, `value` and `label` `ref`s must be references (NOT expressions) and must be children of the `nodeset` expression. That simplifies things somewhat.

[...]

PR #437, opened 2019-05-15
Clear select answers that are no longer part of the choice list

lognaturel:
Partially addresses #436

I have added a series of tests that ensure the answer is only cleared when the choice is no longer available.

This is not a complete fix for the long term but I think it's valuable and should be relatively low-risk and quick to review. I considered doing nothing for now but decided this was worth doing because:
- without this change, it's possible for data collectors to accidentally save incoherent values. For example, imagine a cascading select that first asks for country and then for city. The user could select "France", then "Paris", then change the country to "Spain", jump to the end and unintentionally submit Paris, Spain. 
- we are now supporting cascading selects on a single screen in Collect. Without this change, clearing the first level after answering subsequent levels would leave choices and selections in those subsequent levels. We believe cascading selects on a single screen will be used commonly and so this is a pretty annoying bug.

I don't love adding a side-effect to `populateDynamicChoices` but I think it is expected that the choice selected should always be part of the available choices.

I spent about an hour trying to figure out what a more complete fix might look like and concluded that it would be a fairly big lift for me given that I don't know the details of the DAG/Triggerable implementation. This kind of clearing would need to happen anyway.

@ggalmazor I'll definitely want your feedback on the tests. Specifically:
- I recognize there's a lot of redundancy between them. I wrote them to be independent from each other and to verify each assumption I might make. Do you think I should cut them down to just the bit that is not verified in other tests?
- I use a different naming convention than you do. You use `some_precondition_should_lead_to_this` and I use `somePrecondition_ShouldLeadToThis`. I have a slight preference for the later but I don't know where I picked up that convention so maybe it's a weird one. Should we enforce a convention? I don't mind adopting yours or being loose about it as long as the naming is clear.
- The tests "cheat" since this solution is incomplete. Do the `TODO`s make that clear enough?
- Does moving the Matchers to core.test make sense to you or is there another place you think we should put them?

It clears out a user's answer if that answer is no longer available to be chosen. If I got the logic wrong, there could be a risk of data loss when using selects with choice filters.

See test.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -30,6 +30,7 @@ import org.javarosa.core.model.condition.Triggerable;
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.MultipleItemsData;
 import org.javarosa.core.model.data.SelectOneData;
+import org.javarosa.core.model.data.StringData;
 import org.javarosa.core.model.data.helper.Selection;
 import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.model.instance.FormInstance;
@@ -995,32 +996,33 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
         List<TreeReference> matches = itemset.nodesetExpr.evalNodeset(this.getMainInstance(),
                 new EvaluationContext(exprEvalContext, itemset.contextRef.contextualize(curQRef)));
 
-        DataInstance fi = null;
-        if (itemset.nodesetRef.getInstanceName() != null) // We're not dealing
-        // with the default
-        // instance
-        {
-            fi = getNonMainInstance(itemset.nodesetRef.getInstanceName());
-            if (fi == null) {
-                throw new XPathException("Instance " + itemset.nodesetRef.getInstanceName()
-                        + " not found");
+        DataInstance formInstance;
+        if (itemset.nodesetRef.getInstanceName() != null) { // a secondary instance is specified
+            formInstance = getNonMainInstance(itemset.nodesetRef.getInstanceName());
+            if (formInstance == null) {
+                throw new XPathException("Instance " + itemset.nodesetRef.getInstanceName() + " not found");
             }
         } else {
-            fi = getMainInstance();
+            formInstance = getMainInstance();
         }
 
         if (matches == null) {
-            throw new XPathException("Could not find references depended on by"
-                    + itemset.nodesetRef.getInstanceName());
+            throw new XPathException("Could not find references depended on by" + itemset.nodesetRef.getInstanceName());
         }
 
+        // Get the answer to the current question so that it can be compared with the new choice list. If the answer
+        // to the question is no longer in the choice list, it will be cleared.
+        String currentQuestionAnswer = null;
+        IAnswerData rawValue = getMainInstance().resolveReference(curQRef).getValue();
+        if (rawValue != null) {
+            currentQuestionAnswer = rawValue.getDisplayText();
+        }
+
+        boolean currentAnswerIsInNewChoices = false;
         for (int i = 0; i < matches.size(); i++) {
             TreeReference item = matches.get(i);
 
-            // String label =
-            // itemset.labelExpr.evalReadable(this.getMainInstance(), new
-            // EvaluationContext(exprEvalContext, item));
-            String label = itemset.labelExpr.evalReadable(fi, new EvaluationContext(exprEvalContext,
+            String label = itemset.labelExpr.evalReadable(formInstance, new EvaluationContext(exprEvalContext,
                     item));
             String value = null;
             TreeElement copyNode = null;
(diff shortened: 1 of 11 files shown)


src/org/javarosa/core/model/FormDef.java
```diff
@@ -1029,15 +1031,18 @@ public void populateDynamicChoices(ItemsetBinding itemset, TreeReference curQRef
...
             if (itemset.valueRef != null) {
-                // value = itemset.valueExpr.evalReadable(this.getMainInstance(),
-                // new EvaluationContext(exprEvalContext, item));
                 value = itemset.valueExpr
-                        .evalReadable(fi, new EvaluationContext(exprEvalContext, item));
+                        .evalReadable(formInstance, new EvaluationContext(exprEvalContext, item));
+            }
+
+            // Provide a default value if none is specified
+            value = value != null ? value : "dynamic:" + i;
```
ggalmazor: I'm curious: can this really happen? If I understand this correctly, a form could define choices without a value?

test/org/javarosa/core/model/ItemsetBindingTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+        scenario.answer("/data/level1", "a");
+        scenario.answer("/data/level2", "aa");
+        assertThat(scenario.choicesOf("/data/level3"), containsInAnyOrder(
+            choice("aaa"),
+            choice("aab")));
+
+        scenario.answer("/data/level1", "");
+        assertThat(scenario.choicesOf("/data/level2"), empty());
+        // this next assertion is only true because the one before called populateDynamic choices
```
ggalmazor: This leads me to think that we should change how the Scenario works. Instead of surgically writing a value into the main instance, I think it should follow the same workflow as Collect, which triggers all the side-effects this test relies upon.

test/org/javarosa/core/model/ItemsetBindingTest.java
```diff
@@ -0,0 +1,147 @@
...
+        // TODO: make clearing out answers that are no longer available choices part of the form re-evaluation
```
ggalmazor: Does this have to do with the new way Collect can show many selects on the same page?

test/org/javarosa/core/model/ItemsetBindingTest.java
lognaturel: `Scenario` works great! It writes the value such that the form is re-evaluated with the new value. I mean that all of the expressions (in constraints, calculates, etc) with values that might have changed are re-evaluated. In code, that means `IDag.triggerTriggerables` is called.

The problem is with the implementation -- `populateDynamicChoices` should be called if and only if values that are part of the expression that defines the itemset are changed. For the first level select of the first cascade in the test form, the expression would be `instance('level2')/root/item[level1 = /data/level1]` and only a change to `/data/level1` should trigger a re-evaluation of the itemset expression.

Without the DAG-based optimization, I'd expect all expressions to be re-evaluated every time any value in the form instance changes. That's not what happens in the case of these choice lists and that's good because it would slow answering any question down a lot. Instead it's this weird in-between -- the itemset declaration is not part of the normal form evaluation path and it's only when the choice list is requested that it's recomputed. That's much better than every time a value is saved but not as good as only when a dependent value is evaluated.

I'm evidently having trouble explaining this. Please ask more questions if it's still not clear.

test/org/javarosa/core/model/ItemsetBindingTest.java
lognaturel: Kind of. 😝 The underlying issue is what I described at https://github.com/opendatakit/javarosa/pull/437#discussion_r284443873 and that affects any client whether questions are on one screen or not. Ideally we could remove that `choicesOf` assertion and the following two assertions would pass. That's not the case right now because we need to force a re-computation of the second level's choices and `choicesOf` does that (as it should and as any client would do to display choices).

With the fix I'm proposing here, clients can get correct behavior *but only if they request the choices of all selects in a cascade in order*. With selects on the same screen, some questions might get removed when the screen is refreshed so the current implementation loops backwards over questions. So the slightly irritating thing is that I need another loop that runs forwards to recompute the choices. It's not a very big deal, though, it's the broader performance implication of having to re-compute the same choice lists over and over again that is a problem.

lognaturel:
> Reusing stuff is good, but it creates indirection, which comes with cognitive load.

➕Thanks, that's a really good way to put it.

> I for one welcome the absence of indirection in JavaRosa :)

Agreed. And as we've run into recently, some of the not-totally-thought-through indirection is really painful to work with.

> The problem is that I don't know what's "form re-evaluation" and how or when that happens. 

Hopefully https://github.com/opendatakit/javarosa/pull/437#discussion_r284443873 clarifies!

> I don't think this is much of a problem, though, because it feels like this TODO would become an issue which would be properly documented with extended context about the change.

I'd like that issue to be #436. Once you feel like I've adequately explained things in this PR, I'll summarize in the issue and rework the comments in the code.

test/org/javarosa/core/model/ItemsetBindingTest.java
ggalmazor: I think you nailed the explanation :)

I didn't consider that a DAG-based optimization could perform worse. Maybe combining both is what makes more sense? i.e. rebuilding choice lists when requesting it, only when a dependent value has changed. QuestionDefs could save a snapshot of their last set of choices for fast access to them.

test/org/javarosa/core/model/ItemsetBindingTest.java
ggalmazor: Thanks for the clarification! I think I have the whole context now :)

ggalmazor:
Thanks for your answers, @lognaturel! I think I have all the information now :)

> I'd like that issue to be #436. Once you feel like I've adequately explained things in this PR, I'll summarize in the issue and rework the comments in the code.

That sounds good.

test/org/javarosa/core/model/ItemsetBindingTest.java
lognaturel: > I didn't consider that a DAG-based optimization could perform worse.

I hadn't either, hah. I see what you mean -- it's possible that the choice list won't be requested and in that case triggering the update on dependent value change would mean a lot of work being done for nothing. I think that's not going to be a very common scenario because generally if the choices change, the question has to be answered again.

That said, I like the way that you are thinking. I was thinking of treating the expression that gets evaluated and putting it in the DAG with its dependent values triggering a re-evaluation but perhaps there's a way for `ItemsetBinding` to know its dependents and find out whether any have changed when choices are requested. I'm not sure that's simpler but it's something to explore.

> QuestionDefs could save a snapshot of their last set of choices for fast access to them.

`ItemsetBinding` does that, right? It's just that the choices are currently always cleared?

lognaturel:
> I'd suggest removing the itext tags to make it even simpler.

Done, good call.

I've linked #436 to the specific subthreads here to provide more context. Hopefully that will be enough to get us going when we come back to this.

It's ready for another review when you have a chance, @ggalmazor!

PR #438, opened 2019-05-16
Itemset filtering

ggalmazor:
We're trying to find out ways to improve user-experience of big cascading selects.

This PR explores hijacking the normal XPath filtering of an itemset's nodeset when it uses a predicate with an XPath expression like `[field = /some/other/path]`.

The hijack consists in resolving the field name and filtering value defined in the predicate during `EvaluationContext.expandReferenceAccumulator()` and letting the `TreeElementChildrenList` compute a filtered list in memory with that.

This is a super narrow optimization that targets the main use case of cascading selects.

Check out https://github.com/opendatakit/javarosa/commit/b59e3f026a0d6eebc8f177f685eaeb76efbe7ba4

Benchmark results for the "browsing" use-case defined in the `NigeriaWardsBrowsingRunBenchmark` class:

| Settings | Time | Error |
| --- | --- | --- |
| No filtering hijack, no cache	| 24,698.80 µs	| 361.68 µs |
| Filtering hijack, no cache	| 15,410.88 µs	| 147.33 µs |
| Filtering hijack, cache	| 8,141.32 µs	| 100.18 µs |

chart

### DIFF
--- a/src/jmh/java/org/javarosa/benchmarks/NigeriaWardsBrowsingRunBenchmark.java
+++ b/src/jmh/java/org/javarosa/benchmarks/NigeriaWardsBrowsingRunBenchmark.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2019 Nafundi
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.javarosa.benchmarks;
+
+import static org.javarosa.benchmarks.BenchmarkUtils.dryRun;
+import static org.javarosa.benchmarks.BenchmarkUtils.prepareAssets;
+import static org.javarosa.form.api.FormEntryController.EVENT_BEGINNING_OF_FORM;
+import static org.javarosa.form.api.FormEntryController.EVENT_END_OF_FORM;
+import static org.javarosa.form.api.FormEntryController.EVENT_GROUP;
+import static org.javarosa.form.api.FormEntryController.EVENT_PROMPT_NEW_REPEAT;
+import static org.javarosa.form.api.FormEntryController.EVENT_QUESTION;
+import static org.javarosa.form.api.FormEntryController.EVENT_REPEAT;
+import static org.javarosa.form.api.FormEntryController.EVENT_REPEAT_JUNCTURE;
+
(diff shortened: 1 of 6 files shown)


lognaturel:
Very cool stuff! I can't believe how quickly you were able to whip that up. I think it's taken me longer just to wrap my brain around it.

Let me see if I correctly understand how `expandReferenceAccumulator` works to identify choices that match a choice filter before the proposed change:
- all the children of the secondary instance `/root` that have name `item` are fetched. By default `getChildrenWithName(String name)` performs a linear search through all children and matches by name but thanks to @dcbriccetti's optimization in 09034d6, in the external secondary instance used to define an itemset case, the children are returned in constant time because they all have the same name and  multiplicities > 0.
- then there's a loop over the full set of children to get the references
- then there's a loop over the references to filter down just the references that match the predicate
- then there's a loop that recursively descends into each reference that has been filtered, hits the base case

So that looks overall O(N) where N is the number of `item` nodes in the secondary instance, is that right? For some reason I remembered there was something O(N^2) in there. We're crossing back over to territory that @dcbriccetti has explored in the past.

Your optimized filtering cuts out all the loops but the first one. That is, the new `TreeElementChildrenList.get` variant is responsible for looping through the items and filtering them so the three subsequent loops described above are cut out. Assuming we started with O(N), we're still in O(N) land but the constant multiplier has gone from ~3 (the last loop is over a filtered set so it's not ~4) to ~1. 

The cache is built the first time a specific filter is applied (e.g. Aba North is selected as the lga) and allows for constant time retrieval of filtered choices the next time the exact same filter is applied. I think that means it does the same thing we've been describing at https://github.com/opendatakit/javarosa/pull/437#discussion_r284443873 in terms of speeding up viewing a question with select choices that haven't changed. The additional thing it does is speed up changing back to the same choice. That is, if I select Aba North, then select something else, then select Aba North again, getting the filtered list is O(1).

Does that sound right-ish?

If so (😅) the cache concept is really where things get exciting because we go from O(N) to O(1). It's similar to some of what @dcbriccetti proposed in https://github.com/opendatakit/javarosa/pull/235/files but in this case the index needs to be built from node values (e.g. in `<item><lga>foo</lga></item>`, we need to index from `foo` to the item) to nodes rather than from node names to nodes.

Not having to recompute the choice list every time the same question is rendered is definitely something we want. But I'm wondering whether that might best be done with some of the ideas we were exploring in the other thread -- that is, not clearing out the choice list in `ItemsetBinding` unless the value in the primary instance that the predicate refers to has changed.

The thing that's too bad about the cache implementation here is that the first time a filtered list has to be computed (and I imagine most would only be computed once), it's O(N) because the cache is built incrementally. Building the whole cache in one step would be ideal. The tradeoff is some memory usage but that's mostly because `TreeElement`s get copied, right? Is that necessary? Secondary instances are static so it seems you could just refer to the existing ones and basically take memory concerns off the table.

I'm getting a little ahead of myself. I'll pause here and if I have the right general idea we can discuss further! 😅

ggalmazor:
Hi! I'm doing some cleaning up of my queued PRs. I'm closing this one. 

Anyone can feel free to take the lead on this one and contact me if they need help.

Cheers!

PR #445, opened 2019-06-07
File and data instance parsing improvement

JohnTheBeloved:
Closes #
Issue with large time difference between loading forms with only Internal secondary instances(ISIs) and forms with External secondary instances(ESIs), ISI forms are significantly slower [(~5times slower)](https://docs.google.com/spreadsheets/d/1N2QjEoClxJEf-xZX9CvjzK-UYOggq83ZvReIxgsrocM/edit?disco=AAAADHf6lHw)

I have created three different approaches to solve this issue, in order to compare speeds. The approach in this PR is “Approach 2”. “Approach 3" and “Approach 4” are described below

Compare profiling for  [master branch](https://docs.google.com/spreadsheets/d/1N2QjEoClxJEf-xZX9CvjzK-UYOggq83ZvReIxgsrocM/edit#gid=0) and [this branch(`approach-2`)](https://docs.google.com/spreadsheets/d/1N2QjEoClxJEf-xZX9CvjzK-UYOggq83ZvReIxgsrocM/edit#gid=1200603691)

### Major updates 
**Approach 2**
- Serialising xform file path instead of whole TreeElement (due to the fact that ESI forms load faster than ISI forms) so that at each subsequent loading of the Formdef from the cache, the ISIs are parsed from the XForm. This is because re-parsing the XML file(containing the ISI) is faster than caching the TreeElement and deserialising it which may be due to IO performance on android.

- Profiled on Android studio and Tested on Android device
- Unit tests and Benchmarks created for methods

Two other approaches apart from this were tried and results of the manual timing of the form load can be found [here](https://docs.google.com/spreadsheets/d/1N2QjEoClxJEf-xZX9CvjzK-UYOggq83ZvReIxgsrocM/edit#gid=876410270)
 
**Approach 3**
 - Serialising the ISI strings to the FormDef cache i.e. rather than the xform path(approach-2), or the TreeElement(master branch).
In this case, the whole String is serialised as bytes to the FormDef cache

**Approach 4**
- Storing the ISI strings to separate files and then serializing the paths to each file in the FormDef cache so they can be re-parsed on subsequent Form loads for each ISI

#### How does this change affect users? 

1. Forms with many small ISIs are now on average 2x faster on first load and 2X faster on subsequent loads (10 secondary instances, each containing 50 elements)

2. Forms with many larger ISIs are now on average 4x faster on first load and 3X faster on subsequent loads (10 secondary instance, each containing 500 elements)

3. Forms with very large ISIs are now on average 12x faster on first load and 15X faster on subsequent loads (1 secondary instance containing 50,000 elements)

These results are based on profiling tests with three different Android devices.

See more detailed result [here](https://docs.google.com/spreadsheets/d/1N2QjEoClxJEf-xZX9CvjzK-UYOggq83ZvReIxgsrocM/edit#gid=1200603691) 

#### Describe intentional changes to behaviour and behaviour that could have accidentally been affected by code changes. In other words, what are the regression risks?

No user facing behaviour was changed apart from faster form loading,

After application update, Loading a cached form might break Form loading due to a different structure of the FormDef cache in this PR update (Xform Path are serialized instead of the TreeElement).

It might be necessary to come up with a way of deleting older form caches after this update is installed.


lognaturel:
Thank you, @JohnTheBeloved, for taking on this important line of improvements.

I haven't had a chance to reason through this entirely but I have a couple of preliminary thoughts that may give some ideas for another round of edits. 

I think we need to define what we'd need for a merge. Here are some things I've thought of:
- reasonable confidence that this does not slow down common categories of forms very much
- use of the most efficient and appropriate Java constructs possible
- someone else (such as @ggalmazor or me) who understands the details of the code and the design decisions so it is easier to maintain than other parts of the codebase 
- confidence that redundancy has been reduced where possible (to reduce maintenance burden)

The first one is particularly important. I'm also working on more in-depth feedback on #423 (which I think we can incorporate in follow-on PRs) but one thing that I see that needs to be done there is augmenting the suite of benchmark forms. The characteristics of forms are very likely to result in different performance in different parts of the code so it's really important to have a range of forms. There is more information and some suggested forms [here](https://forum.opendatakit.org/t/automated-benchmarks-in-javarosa/18808/6). At first look through, it appears that forms with relatively small secondary instances would likely perform worse. It would be important to verify that.

For the second point, I think it would be worth doing another pass looking for code [like this](https://github.com/opendatakit/javarosa/pull/445/files#diff-4e885cb61e0ef02a70d0a6b6a0012809R122) which could be simplified. There, the while loop condition is a call to a recursive method. There's no need for recursion there and having a method that consumes tokens as a while loop condition is surprising. Nested loops would probably make this easier to understand.

lognaturel:
Another thing I've been thinking about is that this PR proposes several different interventions. Have you quantified the effect of each? That is, if one of them only results in a small performance change, it may be preferable to avoid it to reduce complexity.

The intent of the question "Why is this the best possible solution? Were any other approaches considered?" is to get a sense of how the PR author has thought about the design of their solution and what ideas they've considered and discarded. For example, did you have a strong reason to change the algorithm used to go from XML file to a tree of (KXML) `Element`s or is the most important change in introducing `KXmlElementParser` the subtree skipping, did you consider saving a file pointer to skip directly to the location in a file where the instance is defined (e.g. using a `RandomAccessFile`), etc.

JohnTheBeloved:
Hi @lognaturel 

> reasonable confidence that this does not slow down common categories of forms very much

Benchmark Report or Profiling report?

If a profiling report,I would need some set of forms to verify.

For Benchmarks, would create forms with the different [use cases](https://forum.opendatakit.org/t/automated-benchmarks-in-javarosa/18808/6), benchmarking each  - would it be better to create the forms dynamically in other to properly plot complexity to load time or manually create the forms?

Like -

`
XFormComplexity = new XFormComplexity(noOfGroups, noOfQuestions, noOfInstances....);
XFormFileBuilder xFormFileBuilder = new XFormFileBuilder();
String xFormString = state.xFormFileBuilder.build();
FormDef formDef = XformUtils.fromXFormString(xFormString) // benchmark this
`

> use of the most efficient and appropriate Java constructs possible

Maybe what was in my head while writing the method isn't a possible scenario or i got it wrong, Updated the PR to simplify complexity

> someone else (such as @ggalmazor or me) who understands the details of the code and the design decisions so it is easier to maintain than other parts of the codebase

I think it's would be great to go through together (live review as suggested?)

> confidence that redundancy has been reduced where possible (to reduce maintenance burden)

A lot of overloaded methods for backward compatibility I guess,  I don't know, maybe during the live review we could discuss if some could be removed


> On reimplementing KXMLElementParser, 

In any case,I don't think KXmlParser can't be used as it is, there would definitely be the need for some form of abstraction to skipSubTree and also for using RandomAccessFile. Open to suggestion or more explanation

lognaturel:
> Benchmark Report or Profiling report?

Hopefully there is a direct correlation between the two. The profiling was effective for answering the question of what's worth optimizing and where users actually feel pain. Hopefully it informed what benchmarks were added to the suite and now the benchmarks can be used. Do you feel that's the case?

My sense is that benchmarks for the end-to-end load process from XML file and from cache are the most valuable ones. That would be the ones for `FormDefCache.readCache` and `FormParserHelper.parse`.

> would it be better to create the forms dynamically in other to properly plot complexity to load time or manually create the forms?

I think it'd be ideal to have both. The ones in the forum thread about slow forms would certainly be a good start. I think it's important that the process always start from an XML file on disk so generated forms should be written to disk.

> I think it's would be great to go through together (live review as suggested?)

Agreed. I could do 3pm or later your time on Wednesday.

lognaturel:
@JohnTheBeloved, @ggalmazor and I did a first live review to talk through the approach. We agree the high-level goal is to exclude the subtrees that correspond to internal secondary instances from processes that loop over the whole `Element` tree (`consolidateText` (which may be entirely removed), `parseElement`, `collapseRepeatGroups`, etc) and from the serialization process.

Two concerns we have with this approach are that it requires reading through the entire form definition file for each secondary instance (relatedly, it feels awkward to keep the form definition path in multiple objects) and that it adds a fair amount of complexity to replicate the external case. 

We agreed that we should have an extreme form with a very large primary instance and several smaller secondary instances to verify performance against. @JohnTheBeloved will build one.

@JohnTheBeloved had also considered writing the internal secondary instance XML to separate files instead of using the regular serialization process and then using the same code path as for external secondary instances. He's going to do a quick demo to compare that performance (without worrying about questions like where those files get saved, how name collisions get resolved, etc).

We also discussed other possibilities like keeping the position in the form definition XML file where an instance is defined or using regular expressions to pull out the instance definitions. Those will not be investigated in the short term but are other options to consider depending on the results.

Following our conversation, I thought a little bit more about the option of keeping the position in the form definition. I think the challenge there is that `XFormParser` is built around a generic `Reader` which can't provide that location info. If we do choose to investigate that option at some point, [`FileChannel`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html) and [`Channels`](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/Channels.html) are worth looking deeper into.

But here's an additional question that may spark ideas -- @JohnTheBeloved why does the serialization process takes so long for internal secondary instances? That is, I would imagine that just writing out the tree of `TreeElement` objects and reading it back in would be as fast or faster than parsing source XML. What am I missing?

lognaturel:
Thanks again to everyone involved here. There has been more communication on private channels so I wanted to try to summarize here so we have a single public reference point.

@JohnTheBeloved has benchmarked this code with a form with a hundred questions in the primary instance and 10 secondary instances with 10-50 elements each. This case is about 5x slower than current performance. @JohnTheBeloved, please correct me if I'm getting this wrong.

@JohnTheBeloved's preliminary results looking at caching string blobs for internal secondary instances appear not to be promising. Again, would be good to get confirmation, @JohnTheBeloved. I can think of various ways to approach that including things like gzipping the instance blob before writing it out. But @JohnTheBeloved, if you already think it can't be a good solution, it's better to get that information out and move on.

I want to be clear that there may be many ways to address the concerns I highlighted in my initial review. Here are a couple of others:
- instead of serializing the form path and form instance ids in separate instance objects, adding a layer that keeps track of the form path once and a list of instance IDs
- doing some analysis to figure out what categories of forms benefit from caching (e.g. lots of logic) and which don't (e.g. lots of nodes without logic) and giving clients (e.g. Collect) access to something like `formDef.wouldBenefitFromCaching()` so they can bypass the cache entirely in those cases

We owe it to our users to have at least considered a few different options and considered tradeoffs. That doesn't necessarily mean doing full implementations of each, doing some exploration is sufficient. If various options have already been considered and it's just not visible here, it would be really helpful to get a summary of approaches considered under the "Why is this the best possible solution? Were any other approaches considered?" question in the PR template. That's an important way to give confidence to the reviewer that the solution is carefully designed without redoing all of the deep analysis (because sadly, we can't just [mind meld](https://memory-alpha.fandom.com/wiki/Mind_meld) with @JohnTheBeloved!).

CC @adamvert

JohnTheBeloved:
Hi @lognaturel 

I have tried [3 other different approaches](https://docs.google.com/spreadsheets/d/1N2QjEoClxJEf-xZX9CvjzK-UYOggq83ZvReIxgsrocM/edit#gid=114789228) apart from this current branch and of the three, this is the fastest. Please see Updated PR description for the approaches.

It is also important to note that the result I shared with you (_approach-2_  : Parsing the form only once while cummulating the ISI strings and serializing them separately as bytes to cache) was just for the time to parse from _FormXML to the FormDef_ using automated benchmarks. - The time to write to cache wasn't accounted for, You can see _approach-5_ in the PR description for a more detailed report.

In all the approaches we tested, this approach was the fastest. It was also faster than the current implementation (Master Branch)

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -166,7 +176,8 @@ public static IDataReference getAbsRef(IDataReference ref, TreeReference parentR
 
     private HashMap<String, SubmissionProfile> submissionProfiles;
 
-    private HashMap<String, DataInstance> formInstances;
+    private HashMap<String, DataInstance> externalFormInstances = new HashMap<>();
```
lognaturel: Would you consider keeping a single map here? I believe the serialization methods are the only ones that need to know the difference between internal and external instances so a single map seems it would better match that narrow usage. 

In that case, the only method that would need to know about the difference between internal and external form instances is `getExternalInstances`. It would be the one to filter out only the external ones. `addNonMainInstance`, the constructor, etc, could then all stay the same and you wouldn't need `getInternalInstances`.

src/main/java/org/javarosa/xml/TreeElementParser.java
```diff
@@ -55,10 +63,83 @@ public TreeElement parse() throws InvalidStructureException, IOException,
...
+     * Returns the internal secondary instance with the instance ID created
+     * Used for deserializing
+     * @return The TreeElement node of the internal instance
+     * @throws IOException
+     * @throws InvalidReferenceException
+     * @throws InvalidStructureException
+     * @throws XmlPullParserException
+     * @throws UnfullfilledRequirementsException
+     */
+    public TreeElement parseInternalSecondaryInstance() throws IOException, InvalidReferenceException, InvalidStructureException, XmlPullParserException, UnfullfilledRequirementsException {
```
lognaturel: This is no longer used, right?

src/main/java/org/javarosa/xml/TreeElementParser.java
```diff
@@ -55,10 +63,83 @@ public TreeElement parse() throws InvalidStructureException, IOException,
...
+     *
+     * @throws IOException There was an error parsing the xml file
+     * @throws InvalidStructureException
+     * @throws XmlPullParserException
+     * @throws UnfullfilledRequirementsException
+     */
+    public List<TreeElement> parseInternalSecondaryInstances() throws InvalidStructureException, XmlPullParserException, UnfullfilledRequirementsException, IOException {
+        /**
+         * This could be implemented using the XPath reference to target specifically
+         * the path of the instance node, but currently, this targets any path where
```
lognaturel: Could you please explain this concept in greater detail? Evaluating an XPath expression with existing JavaRosa components requires a tree of `TreeElement`s which is what this is building. So it doesn't seem possible to evaluate an XPath expression in the process of building it.

src/main/java/org/javarosa/xml/TreeElementParser.java
```diff
@@ -55,10 +63,83 @@ public TreeElement parse() throws InvalidStructureException, IOException,
...
+            }
+            return internalInstances;
+        } else {
+            throw new InvalidStructureException("Invalid XML File, no detected xml node - Depth is 0");
+        }
+    }
+
+
+    public boolean isInstanceNode() throws IOException, XmlPullParserException {
+        int ret = nextNonWhitespace();
```
lognaturel: Would you consider moving this call to the parsing method and passing the token into the boolean method? This method's name implies it's just doing a check but in fact it's consuming a token.

src/main/java/org/javarosa/xml/InternalDataInstanceParser.java
```diff
@@ -0,0 +1,64 @@
...
+     * @return List of all the #InternalDataInstance in the XForm
+     * @throws IOException                       if FileInputStream can’t find the file, or ElementParser can’t read the stream
+     * @throws InvalidReferenceException         if the ReferenceManager in getPath(String srcLocation) can’t derive a reference
+     * @throws UnfullfilledRequirementsException thrown by {@link TreeElementParser#parse()}
+     * @throws XmlPullParserException            thrown by {@link TreeElementParser#parse()}
+     * @throws InvalidStructureException         thrown by {@link TreeElementParser#parse()}
+     */
+    public static HashMap<String, DataInstance> buildInstances(String xFormSrc)
+        {
+//throws IOException, UnfullfilledRequirementsException, XmlPullParserException, InvalidStructureException, InvalidReferenceException
```
lognaturel: The fact that these exceptions are caught below means we could possibly end up in a state where JR thinks it has read in all secondary instances when deserializing a form but in fact some or all have failed, right? I believe that the user would then get a crash on first form evaluation, does that sound correct? I think that's ok but want to make sure I understand the implications. If that's all sounding right then this comment can be removed.

JohnTheBeloved: Hi @lognaturel 

Updated to reflect the changes in the comments

lognaturel:
Closes https://github.com/opendatakit/javarosa/issues/456

I want to do one last pass tomorrow but this is looking ready to me.

lognaturel:
Thanks again, @JohnTheBeloved! I know a lot of other ideas came out of your analysis. Could you please file issues for others that you think might be worth pursuing in addition to this major one?

https://github.com/opendatakit/javarosa/issues/447 would be an example of that.

PR #448, opened 2019-06-26
Re-enable support for external secondary CSV instances 

### LINKED ISSUE #417: Re-enable support for external secondary CSV instances without breaking pulldata()/search()
#397 added support for parsing a CSV as a secondary instance. It was then disabled in #416 because it broke search/pulldata in Collect.

We will need to figure out a way to add this support back in without interfering with the existing implementations. These may eventually be deprecated but since they solve a real problem and are [in broad use](https://forum.opendatakit.org/t/external-data-current-state-of-affairs/14564/8?u=ln), there needs to be an acceptable, suitably performant alternative in place first. Even if/when they are officially marked for deprecation, I imagine they will stay in for quite a while.

@ggalmazor rightfully pointed out [here](https://github.com/opendatakit/javarosa/pull/397#issuecomment-477977034) that the Collect implementation of search/pulldata doesn't use the instance declared in the form. `pyxform` adds that instance declaration when `pulldata` is used (https://github.com/XLSForm/pyxform/issues/43). This is because clients that do things in a spec-compliant way (e.g. Enketo), need to know what instance to load. Collect just reads any CSV attached with the form into a database ("magic"). 

Since the `pulldata` function can be thought of as an abstraction on top of XPath queries (kind of like `indexed-repeat`), it's easy for clients that do load external secondary CSV instances to implement, hence the instance declaration added by pyxform so they can support it. For that reason, I think it will likely become an official part of the spec. `search` is strange all around, starting from it being an appearance.

Collect search/pulldata are very performant. It's important that the extra overhead of loading the CSV instance doesn't occur if the user intends to use those implementations. That is, if the form has the following instance declaration: `<instance id="foo" src="jr://file-csv/foo.csv" />`, the CSV should only actually be loaded if the form definition includes `instance("foo")`. I'm not immediately seeing a way to do that since secondary instances are [parsed before the main instance](https://github.com/opendatakit/javarosa/blob/c47608da999f593e9fefec618736e6c51ffa1a9d/src/org/javarosa/xform/parse/XFormParser.java#L464). Maybe secondary instances could be "lazy loaded" the first time a reference to them is seen in the main instance instead of being parsed upfront?

There's an additional issue uncovered by #397: `jr://file` is a substring of `jr://file-csv` and the connector matching is done by looking for a prefix. I think this could be fixed by looking for the connector followed by a `/` or something like that.

musamusa:
Closes #417

Test was added to determine when an external secondary instance `id` is referenced in a `pulldata` function that the instance is parsed as Internal form instance (`FormInstance` not `ExternalDataInstance`) without loading the external file into memory.

Enabling external secondary CSV instance  without an update to [FormLoaderTask](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java#L144) will result in forms ESI without `pulldata()` reference throwing `FileNotFoundException`

No, the code was tested using [Sample-Preloading](https://github.com/opendatakit/javarosa/blob/master/src/test/resources/Sample-Preloading.xml) form.

No

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -91,6 +91,7 @@ import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.core.util.externalizable.PrototypeFactoryDeprecated;
 import org.javarosa.model.xform.XPathReference;
 import org.javarosa.xform.util.InterningKXmlParser;
+import org.javarosa.xform.util.SecondaryInstanceAnalyzer;
 import org.javarosa.xform.util.XFormAnswerDataParser;
 import org.javarosa.xform.util.XFormSerializer;
 import org.javarosa.xform.util.XFormUtils;
@@ -172,6 +173,7 @@ public class XFormParser implements IXFormParserFunctions {
 
     private final List<WarningCallback> warningCallbacks = new ArrayList<>();
     private final List<ErrorCallback> errorCallbacks = new ArrayList<>();
+    private SecondaryInstanceAnalyzer secondaryInstanceAnalyzer = new SecondaryInstanceAnalyzer();
 
     //incremented to provide unique question ID for each question
     private int serialQuestionID = 1;
@@ -320,6 +322,7 @@ public class XFormParser implements IXFormParserFunctions {
         mainInstanceNode = null;
         instanceNodes = new ArrayList<>();
         instanceNodeIdStrs = new ArrayList<>();
+        secondaryInstanceAnalyzer = new SecondaryInstanceAnalyzer();
 
         itextKnownForms = new ArrayList<>(4);
         itextKnownForms.add("long");
@@ -471,9 +474,8 @@ public class XFormParser implements IXFormParserFunctions {
                 final String instanceId = instanceNodeIdStrs.get(instanceIndex);
                 final String instanceSrc = parseInstanceSrc(instance, lastSavedSrc);
 
-                // Disable jr://file-csv/ support by explicitly only supporting jr://file/
-                // until https://github.com/opendatakit/javarosa/issues/417 is addressed
-                if (instanceSrc != null && instanceSrc.toLowerCase().startsWith("jr://file/")) {
+                // only build when ESI Id is not referenced in pulldata
+                if (instanceSrc != null && instanceSrc.toLowerCase().startsWith("jr://file") && secondaryInstanceAnalyzer.shouldSecondaryInstanceBeParsed(instanceId)) {
                     final ExternalDataInstance externalDataInstance;
                     try {
                         externalDataInstance = ExternalDataInstance.build(instanceSrc, instanceId);
@@ -493,6 +495,7 @@ public class XFormParser implements IXFormParserFunctions {
                 }
             }
         }
+
         //now parse the main instance
         if (mainInstanceNode != null) {
             FormInstance fi = instanceParser.parseInstance(mainInstanceNode, true,
@@ -636,6 +639,9 @@ public class XFormParser implements IXFormParserFunctions {
             int type = e.getType(i);
             Element child = (type == Node.ELEMENT ? e.getElement(i) : null);
             String childName = (child != null ? child.getName() : null);
+
+            if (child != null)
+                secondaryInstanceAnalyzer.analyzeElement(child);
 
             if ("itext".equals(childName)) {
                 parseIText(child);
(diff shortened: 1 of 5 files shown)


src/main/java/org/javarosa/xform/util/SecondaryInstanceAnalyzer.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.regex.Pattern;
+
+import static org.javarosa.xform.util.XFormSerializer.elementToString;
+
+
+public class SecondaryInstanceAnalyzer {
+    // a list of SIs that need to be built in memory
+    private final List<String> inMemorySecondaryInstances;
+
+    private final Pattern INSTANCE_FUNCTION_PATTERN = Pattern.compile("instance\\s*\\(\\s*'([^\\s]{1,64})'\\s*");
```
lognaturel: This regular expression should have some corresponding tests to document its intended behavior.  A comment would help too (white space is allowed after the function name, before the argument, etc). In particular, there should be an explanation for the 64 character limit.

Also note that double quotes are allowed for strings as well as single quotes.

src/test/java/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -130,6 +132,19 @@ public void parsesPreloadForm() throws IOException {
         assertEquals("Sample Form - Preloading", formDef.getTitle());
     }
 
+    @Test
+    public void parsesExternalSecondaryInstanceAsFormInstance() throws IOException {
+        Path form = r("Sample-Preloading.xml");
```
lognaturel: It would be ideal to have a test that just isolates the case that is trying to be handled here. For example, a very narrow one could define an external CSV instance that is not referred to in an `instance()` function. A slightly broader test form would have one that is and one that isn't so both of those cases can be verified.

src/test/java/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -130,6 +132,19 @@ public void parsesPreloadForm() throws IOException {
...
+        setUpSimpleReferenceManager("file-csv", form.getParent());
+        FormDef formDef = parse(form);
+        Enumeration<DataInstance> elements = formDef.getNonMainInstances();
+
+        while (elements.hasMoreElements()) {
+            DataInstance instance = elements.nextElement();
+            assertTrue(instance instanceof FormInstance);
```
lognaturel: This looks like the opposite of what is desired but perhaps some comments would help clarify. In that form, an instance is declared with `<instance id="hhplotdetails" src="jr://file-csv/hhplotdetails.csv">`. However, it's never referred to as `instance('hhplotdetails')` so the file should not be loaded as an in-memory instance.

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -471,9 +474,8 @@ private void parseDoc(Map<String, String> namespacePrefixesByUri, String lastSav
                 final String instanceId = instanceNodeIdStrs.get(instanceIndex);
                 final String instanceSrc = parseInstanceSrc(instance, lastSavedSrc);
 
-                // Disable jr://file-csv/ support by explicitly only supporting jr://file/
-                // until https://github.com/opendatakit/javarosa/issues/417 is addressed
-                if (instanceSrc != null && instanceSrc.toLowerCase().startsWith("jr://file/")) {
+                // only build when ESI Id is not referenced in pulldata
+                if (instanceSrc != null && instanceSrc.toLowerCase().startsWith("jr://file") && secondaryInstanceAnalyzer.shouldSecondaryInstanceBeParsed(instanceId)) {
```
lognaturel: This will always be false because the primary instance is loaded before the secondary instances are, right? Doesn't `analyzeElement` always happen after this has already been done?

lognaturel:
Here's a quick and dirty demo of what I had in mind -- https://github.com/opendatakit/javarosa/compare/master...lognaturel:issue-417. This uses an existing pass over the `Element` tree, hooks into the parser code so no regular expressions are needed, and only looks for instance function calls when we're known to be in a function call. I don't love that `XPathFuncExpr` makes a static `XFormParser` call but given that it also calls other static methods, I think it's ok. Let me know what you think of that approach and if it's looking ok I'll add tests and get it done.

musamusa:
> Let me know what you think of that approach and if it's looking ok I'll add tests and get it done.

@lognaturel Yes, your implementation worked exactly like mine but your approach is cleaner and without the regular expressions will perform better. I'll go ahead and close mine.

PR #450, opened 2019-06-27
Consider instance name in TreeReference comparison

### LINKED ISSUE #449: Referring to node in secondary instance with path that also exists in primary instance can result in a dependency cycle
#### Software versions 
All JR versions

#### Problem description
A useful pattern with the `last-instance` feature is to accumulate values across instances. See [dont-show-options-selected-in-other-instances.xml.txt](https://github.com/opendatakit/javarosa/files/3336913/dont-show-options-selected-in-other-instances.xml.txt) for an example. That means referring to the same reference in the current instance and in the previous instance. JavaRosa incorrectly reports a cycle because `TreeReference`s don't take instance into consideration when compared for equality.

Try to parse the form above (e.g. by loading it in Collect). See the crash at https://github.com/opendatakit/javarosa/blob/59611cc0755dfbaf6631866ff15d6a9a516df704/src/main/java/org/javarosa/core/model/IDag.java#L473. The cycle reported involves the `previous_schools` node.

#### Expected behavior
No dependency cycle reported.

lognaturel:
Closes #449 

<details>
<summary> Initial description</summary>
I see two testing options: 
* create `TreeReference`s and verify the result of testing for equality between them
* add a simple form that exhibits this problem

I don't like the first because we don't have tests of that kind. I could target just the new behavior which would lead to trivial tests. Or I could try to fully test the equals method which is likely to be a bit of a mess with all the multiplicity and predicate options. Admittedly, maybe the latter is a reason to write those tests but I ran into this by accident and don't want to rathole here.

I don't love the second because I don't think that form would really augment the test suite in a meaningful way.

That said, I can be convinced that I have a bad attitude and should Do The Right Thing™️. (CC @seadowg) 

Tried form from issue in Collect.

This is the simplest possible way to compare two objects that can be null.

</details>

Tried form from issue in Collect and added a test (thank you @ggalmazor and @seadowg).

I would have liked to use Java's `Objects.equals` but that's not available in Android API levels < 19. See https://github.com/opendatakit/collect/issues/3192.

The intent is for this to have no user impact other than enabling previously impossible behavior. If I messed this up and equality of `TreeReference`s has changed in an unexpected way, it could lead to problems with any kind of form, though.

[dont-show-options-selected-in-other-instances.xml.txt](https://github.com/opendatakit/javarosa/files/3336961/dont-show-options-selected-in-other-instances.xml.txt)

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/instance/TreeReference.java
+++ b/src/main/java/org/javarosa/core/model/instance/TreeReference.java
@@ -16,14 +16,8 @@
 
 package org.javarosa.core.model.instance;
 
-import java.io.DataInputStream;
-import java.io.DataOutputStream;
-import java.io.IOException;
-import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.List;
-
 import org.javarosa.core.util.DataUtil;
+import org.javarosa.core.util.Objects;
 import org.javarosa.core.util.externalizable.DeserializationException;
 import org.javarosa.core.util.externalizable.ExtUtil;
 import org.javarosa.core.util.externalizable.ExtWrapNullable;
@@ -31,6 +25,13 @@ import org.javarosa.core.util.externalizable.Externalizable;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.xpath.XPathException;
 import org.javarosa.xpath.expr.XPathExpression;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
 
 public class TreeReference implements Externalizable, Serializable {
     /** @deprecated replaced by {@link #DEFAULT_MULTIPLICITY} */
@@ -402,6 +403,10 @@ public class TreeReference implements Externalizable, Serializable {
             return true;
         } else if (o instanceof TreeReference) {
             TreeReference ref = (TreeReference)o;
+
+            if (!Objects.equals(getInstanceName(), ref.getInstanceName())) {
+                return false;
+            }
 
             if (this.refLevel == ref.refLevel && this.size() == ref.size()) {
                 for (int i = 0; i < this.size(); i++) {
--- a/src/main/java/org/javarosa/core/util/Objects.java
+++ b/src/main/java/org/javarosa/core/util/Objects.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2013 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.javarosa.core.util;
+
+public class Objects {
+    /**
+     * Null-safe equivalent of {@code a.equals(b)}.
+     */
+    public static boolean equals(Object a, Object b) {
+        return (a == null) ? (b == null) : a.equals(b);
+    }
+}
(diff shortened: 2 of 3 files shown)


seadowg:
Yeah I think I agree with @ggalmazor that going with option 1 seems like the way to go.

lognaturel:
🙏 Thank you. Please let me know if you see a better approach or additional tests I should add.

ggalmazor:
The test is a great example of a good place to use the junit `Parameterized` runner class:
```java
@RunWith(Parameterized.class)
public class TreeReferenceEqualsTest {
    @Parameterized.Parameter(value = 0)
    public String testCase;

    @Parameterized.Parameter(value = 1)
    public String pathA;

    @Parameterized.Parameter(value = 2)
    public String pathB;

    @Parameterized.Parameter(value = 3)
    public boolean expectToBeEqual;

    @Parameterized.Parameters(name = "{0}")
    public static Collection<Object[]> data() {
        return Arrays.asList(new Object[][]{
            {"same path in primary instance are equal", "/foo/bar", "/foo/bar", true},
            {"same path in secondary instance are equal", "instance('foo')/bar/baz", "instance('foo')/bar/baz", true},
            {"same path in primary and secondary instance are not equal", "instance('foo')/foo/bar", "/foo/bar", false},
            {"same path in different secondary instances are not equal", "instance('foo')/foo/bar", "instance('bar')/foo/bar", false},
            {"different paths in primary instance are not equal", "/foo/bar", "/foo/bar/quux", false},
            {"different paths in secondary instance are not equal", "instance('foo')/foo/bar", "instance('foo')/foo/bar/quux", false}
        });
    }

    // region Equality across same or different instances

    // Prior to JR 1.15, the instance was not taken into account when comparing two TreeReferences. That meant two
    // references representing the same path in two different instances would be considered the same reference.
    @Test
    public void treeReferencesWithSamePathInPrimaryInstance_AreEqual() {
        if (expectToBeEqual)
            assertThat(buildRef(pathA), is(buildRef(pathB)));
        else
            assertThat(buildRef(pathA), is(not(buildRef(pathB))));
    }

    //endregion
}
```

seadowg:
@ggalmazor for some reason Github is not letting my react to that post so here is my reaction: 👍

lognaturel:
Good point, @ggalmazor, thank you. I used the parameterized approach with a couple of changes. I put the comment with the data that define the cases for this change rather than on the generic test method. I also changed the assertion in the test method to `assertThat(buildRef(pathA).equals(buildRef(pathB)), is(expectToBeEqual))` so the call on `equals` is explicit rather than implicit. Please don't hesitate to send me back for another round if you see any other changes worth making. You can also consider doing a follow-on PR to suggest small changes.

ggalmazor:
Hi!

>  I also changed the assertion in the test method to assertThat(buildRef(pathA).equals(buildRef(pathB)), is(expectToBeEqual)) so the call on equals is explicit rather than implicit

I don't think this is a good idea... Hamcrest matchers work better when all the assertion is in your matchers. By doing an `equals` in the left hand-side, you're taking that out from the assertion.

The consequence is that now, when the test fails, it gives much less information than before. Now a boolean is not true/false, instead of a object not being equal to another.

If you want to make it explicit that we're "equalling" two objects, you could use the `Matchers.equalTo` instead of `Matchers.is` (`is` is more ideomatic on Hamcrest, and it does the same thing as `equalTo`, but it's fine to use it anyway IMO).

You could do this:
```java
@Test
public void treeReferenceEqualsTest() {
    Matcher<TreeReference> matcher = equalTo(buildRef(pathB));
    if (!expectToBeEqual)
        matcher = not(matcher);
    assertThat(buildRef(pathA), matcher);
}
```
(or, you know, use a ternary expression ;) )

lognaturel:
> The consequence is that now, when the test fails, it gives much less information than before. Now a boolean is not true/false, instead of a object not being equal to another.

Yes, but what we want to test is the result of the `equal`s method. Incidentally we can use the `is` matcher to implicitly call that but I realized it's not very consistent with how we'd verify the return value of a method with any other return type.

A failed test using `is` would show something like 
```
Expected: is <instance(foo)/foo/bar>
     but: was </foo/bar>
```

which doesn't particularly correspond to what we're trying to test. On the other hand,

```
Expected: is <false>
     but: was <true>
```

makes it clear that we expected the result of an equality check to be false but it was true for some reason.

Anyway, hopefully these tests are more of a formality than anything else so if you prefer using `is` even with this explanation, that's ok by me! It is what I went for initially for the idiomatic reasons you described but then I realized what it was testing wasn't quite right.

PR #457, opened 2019-07-12
Standardize spacing and reduce API changes (follow-on to #445)

lognaturel:
Follow-on to #445

Ran tests. There should be no functional changes. The API changes are to functionality that has not yet been released.

The first commit has subjective stylistic changes. Hopefully they are non-controversial but please let me know if I forgot any or if I introduced a bad change. The second commit reverts changes that are not currently used. Perhaps those changes will be brought back eventually but they should come in along with their usage.

There should be no change in behavior between this and #445.

@JohnTheBeloved please review and see if you agree or have alternate suggestions.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -149,10 +149,8 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
     private String title;
 
     /**
-     * The file path to the formXML file
-     * this is being used for deserialization of the internal instances
-     * during caching
-     * */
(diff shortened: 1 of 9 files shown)


lognaturel:
I identified a couple of issues: external instances were being read in by the new code path meant just for internal instances and the primary instance was defined as the one without an ID rather than the first one in the form. I added tests to document and verify those fixes.

lognaturel:
Let's try to get this in the next snapshot release so that quality assurance can start on the Collect side. 

@JohnTheBeloved, I'd really like to get your review. Even if @ggalmazor has already had a chance to review and merge, we can file issues for any additional questions you may come up with. Thanks for really focusing #445 down to one category of change -- it made identifying problems a lot easier!

PR #461, opened 2019-07-17
Rename setlocation to setgeopoint

### LINKED ISSUE #459: Rename everything about setlocation to setgeopoint
See discussion at https://forum.opendatakit.org/t/spec-addition-proposal-location-preload/10768/20

lognaturel:
Closes #459 

Ran tests.

I used IntelliJ's refactor tool where possible. Then I did a search for "setlocation" and renamed test forms and their contents by hand. This seemed like the most reliable process.

If someone had a form with setlocation, it no longer works. Since this hasn't been released and was very recently merged, it's very unlikely.

See test forms.

Yes, but will be done with the Collect/pyxform changes.


src/test/resources/org/javarosa/core/model/actions/setgeopoint-action-bad-namespace.xml
```diff
@@ -2,15 +2,15 @@
...
         <model>
             <instance>
-                <data id="setlocation-action-bad-namespace">
+                <data id="setgeopoint-action-bad-namespace">
                     <location/>
                 </data>
             </instance>
             <bind nodeset="/data/location" type="string" />
-            <setlocation ref="/data/location" event="odk-instance-first-load" />
+            <setgeopoint  ref="/data/location" event="odk-instance-first-load" />
```
yanokwa: Remove this space.

lognaturel: 🙏

PR #464, opened 2019-07-18
Added UTF-8 encoding options in compileJava and compileTestJava for I…

neal0892:
Explanation:
When not setting the compiler encoding option explicitly, the problem you can run into is, that the Java compiler uses a different file encoding (in Gradle's case the system's default encoding) than the source code files are encoding in.
It happened to me that I had my Java code files encoded in UTF-8, Intellij IDEA compiling with UTF-8 during development (simply because it's faster), but Gradle compiling with the platform default encoding (e.g. Cp1252) and this results in build-failures(due to presence of greek letters in GeoUtils.java).

Closes #

Executed the builds and verified that the build is successful

NA

This does not changes anything .This change is made so that IntejiIdea does not pick system default encoding while building the project instead of UTF -8.

No

No

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,8 @@ sourceSets {
     test.java.srcDirs = ['src/test/java']
     test.resources.srcDirs = ['src/test/resources']
 }
+compileJava.options.encoding = "UTF-8"
+compileTestJava.options.encoding = "UTF-8"
 
 compileJava {
     options.compilerArgs << "-Xlint:deprecation"


lognaturel:
Thank you, @neal0892! 🎉🎉🎉 Great that other Windows users will be saved the trouble of figuring this out.

neal0892:
@lognaturel ,Thanks for the approval...Looking forward to more meaningful contributions.

PR #467, opened 2019-07-19
Add normalize-space function and tests.

### LINKED ISSUE #460: Add trim() / strip() custom XPath function
#### Software versions 
JavaRosa v1.x.x, Java v1.x.x, operating system: all current

#### Problem description
I would like to be able to easily strip leading and trailing whitespace in ODK. Because of how keyboards on Android tablets or phones work, spaces are often added at the end text entry. This can be a problem when trying `concat(${name}, '-', ${age})` and expecting there not to be whitespace. Another issue would be if pressing `<Enter>` and a newline is at the end of `${name}`.

The `trim` (if going with Java convention) or `strip` function just removes leading and trailing whitespace and calls `String.trim()`

#### Other information 
I have this in my fork of `javarosa` and this is the only thing preventing me from using the community version right now. It is fairly easy to add the function to `javarosa` at https://github.com/opendatakit/javarosa/blob/master/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java, and some tests at https://github.com/opendatakit/javarosa/blob/master/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java, and to update the docs at https://github.com/opendatakit/docs/blob/master/odk1-src/form-operators-functions.rst#strings. I can make pull requests if those are the right places.

jkpr:
Closes #460 

I added tests in the source code to verify functionality.

The Javarosa ecosystem has demand for the `normalize-space` feature. The functionality is described at https://www.w3.org/TR/1999/REC-xpath-19991116/#function-string-length. This solution implements the one-argument form only, similar to `string-length`.

Allows users to use a new XPath function when authoring XForms.

Test form attached:
[normalize-space-test.xlsx](https://github.com/opendatakit/javarosa/files/3408870/normalize-space-test.xlsx)

Documentation links:
https://github.com/opendatakit/docs/issues/1070
https://github.com/opendatakit/xforms-spec/issues/242

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -391,6 +391,9 @@ public class XPathFuncExpr extends XPathExpression {
         } else if (name.equals("string-length")) {
             assertArgsCount(name, args, 1);
             return stringLength(argVals[0]);
+        } else if (name.equals("normalize-space")) {
+            assertArgsCount(name, args, 1);
+            return normalizeSpace(argVals[0]);
         } else if (name.equals("checklist") && args.length >= 2) { //non-standard
             if (args.length == 3 && argVals[2] instanceof XPathNodeset) {
                 return checklist(argVals[0], argVals[1], ((XPathNodeset) argVals[2]).toArgList());
@@ -607,6 +610,12 @@ public class XPathFuncExpr extends XPathExpression {
             return 0.0;
         }
         return (double) s.length();
+    }
+
+    public static String normalizeSpace (Object o) {
+        String s = toString(o);
+        String normalized = s.trim().replaceAll("\\s+", " ");
+        return normalized;
     }
 
     /**
--- a/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
+++ b/src/test/java/org/javarosa/xpath/test/XPathEvalTest.java
@@ -281,6 +281,19 @@ public class XPathEvalTest extends TestCase {
         testEval("ends-with('abc', 'c')",   true);
         testEval("ends-with('', '')",       true);
 
+        logTestCategory("other string functions");
+        testEval("normalize-space('')", "");
+        testEval("normalize-space('  ')", "");
+        testEval("normalize-space(' \t\n ')", "");
+        testEval("normalize-space(' a')", "a");
+        testEval("normalize-space(' a   ')", "a");
+        testEval("normalize-space(' ab ')", "ab");
+        testEval("normalize-space(' a    b ')", "a b");
+        testEval("normalize-space(' a\nb\n')", "a b");
+        testEval("normalize-space('\na\nb\n')", "a b");
+        testEval("normalize-space('\nab')", "ab");
+        testEval("normalize-space(' \ta\n\t  b \n\t c   \n')", "a b c");
+
         logTestCategory("date functions");
         testEval("date('2000-01-01')", DateUtils.getDate(2000, 1, 1));
         testEval("date('1945-04-26')", DateUtils.getDate(1945, 4, 26));


jkpr:
The zero-argument form could be implemented following the pattern at https://github.com/opendatakit/javarosa/blob/master/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L284 for `position` (it takes zero or one argument). However, my exposure to XForms is through ODK and things like `string-length`, so I am happy with the one-argument form only.

lognaturel:
That sounds right. I think you could do something like pass `evalContext.resolveReference(evalContext.getContextRef()).getValue().getDisplayText()` to your `normalizeSpace` method. I haven't tried this so it might not be quite right but I think it's the idea.

Up to you whether you want to give it a try here or not. I'm happy to start with this and then later add the zero-arg `string-length` and `normalize-space` variants for completeness (though I doubt they'd be very popular).

src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -609,6 +612,12 @@ public static Double stringLength (Object o) {
         return (double) s.length();
     }
 
+    public static String normalizeSpace (Object o) {
+        String s = toString(o);
+        String normalized = s.trim().replaceAll("\\s+", " ");
```
lognaturel: Java's `\s` includes form feed and vertical tab which are not part of [XML's whitespace definition](https://www.w3.org/TR/REC-xml/#NT-S) but those aren't use anyway so I don't think it matters.

jkpr:
@lognaturel Awesome! We will be happy to see this get into ODK. I will work on the documentation now.

As for the zero-arg version, I put in some dev time, but I think it might be a little more complicated than the one-arg version 😬.

How would `normalize-space()` (zero-arg) even be used? I think it wouldn't make sense in calculates, so maybe just constraints or constraint messages? Thinking of XLSForms, it would be great if for a `text` question, we could automatically apply `normalize-space` to the text that is entered and save it to that node. However, that doesn't seem possible. I think that would have to be a separate `calculate` to apply `normalize-space` to that `text`, but that is already a different node and would use the one-arg version.

I started a test idea with the the `Scenario` class. I think that class has a lot of potential.
https://github.com/opendatakit/javarosa/blob/master/src/test/java/org/javarosa/core/test/Scenario.java

I am trying to use `normalize-space` in a constraint and test it that way. Currently I am getting a NullPointerException when trying to get the value. Interestingly, this works when I change it to `normalize-space(.)`. Perhaps you can take some of this when it comes time to implement the zero-arg version.

See my code here:
https://github.com/jkpr/javarosa/compare/jkp-normalize-space...jkp-zero-arg-normalize-space?expand=1

lognaturel:
https://github.com/opendatakit/xforms-spec/pull/243#pullrequestreview-264441361 has more discussion about the zero-arg variant usage.

Thank you, @jkpr, for those notes!

tiritea:
Its probably worth noting somewhere [but whether we care enough or not is a different matter...], but the java definition of 'whitespace' - aka '\s' isnt quite identical to the XML definition. AFAIK [XML whitespace](https://www.oracle.com/technetwork/articles/wang-whitespace-092897.html) is only:

space, \n (newline), \r (carriage return), \t (tab)

whereas I think java ['\s' whitespace](https://docs.oracle.com/javase/tutorial/essential/regex/pre_char_classes.html) also includes:

..., \f (form feed), \x0B (vertical tab, whatever that is...)

jkpr:
If we wanted to use XML whitespace only: we can change the replace all to `.replaceAll("[ \\n\\r\\t]+", " ")`

src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
lognaturel: @tiritea I think between this note and yours, that's plenty of info about the decision made here.

tiritea:
Opps sorry. I didn’t see your earlier comment covering exact same.

lognaturel:
No worries, @tiritea! I realized after sending that my message sounded annoyed but I really just wanted to confirm with you that this felt sufficient. I don't think it needs to be formally documented as a deviation since... ODK Typewriter is not in common usage.

PR #469, opened 2019-07-22
Include multiplicities only for repeatable groups in TreeReference.toString() method

grzesiek2010:
Closes https://github.com/opendatakit/collect/issues/3126

I tested the attached form with changes on the Collect side.

I decided to the safer option. I don't change setting multiplicities (what is used in multiple classes) I just include them in TreeReference.toString() if a level represents a repeatable group.

As I said above I decided to use the safer solution so I don't think it's risky. Generally, I changed that toString method only so testing it would be fine.

I used this one:
[test.xml.txt](https://github.com/opendatakit/javarosa/files/3417581/test.xml.txt)
but any form with regular groups and repeatable groups is fine.

No, I thought that we have some samples in the [doc](https://docs.opendatakit.org/form-audit-log/) that we would need to fix but looks as if I was wrong.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -368,6 +368,7 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
             IFormElement temp = elements.get(i);
             if (temp instanceof GroupDef && ((GroupDef) temp).getRepeat()) {
                 TreeReference repRef = FormInstance.unpackReference(temp.getBind());
+                ref.markAsRepeatable(repRef.size() - 1);
                 if (repRef.isParentOf(ref, false)) {
                     int repMult = multiplicities.get(i);
                     ref.setMultiplicity(repRef.size() - 1, repMult);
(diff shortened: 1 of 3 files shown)


lognaturel:
Thanks for looking into this tricky issue, @grzesiek2010! I knew it was subtle but there was more here than I had anticipated.

At first glance, I agreed that this was a safe, straightforward solution. But thinking more through it, I think it has a number of issues. The biggest is that this implicitly ties `TreeReference`s to a specific form's primary instance. That is, `TreeReference`s are currently just convenient representations of XPath paths that are portable -- nothing ties them to a specific form. But by introducing `isRepeatable`, now they only make sense in the context of a specific form definition.

Related to that, `TreeReference`s are not only used to represent paths in the primary instance (which defines the form structure) but also to represent paths in secondary data instances which define select itemsets or just documents to look values up in. Those secondary instances generally do have sequences (e.g. multiple `item` blocks) but they're not backed by a repeat. This would mean those sequences' paths would be represented without position predicates.

Finally, even though it's unlikely, a client could be relying on the existing structure of the `toString`'s output.

An additional note for the future -- if we did want to go this route of adding a field to `TreeReferenceLevel`, we'd have to make sure that field was included in `readExternal` and `writeExternal` so that it would be serialized. 

I've proposed an alternative that I hope you will like at https://github.com/opendatakit/collect/pull/3231 that doesn't require new state or coupling.

PR #474, opened 2019-08-04
Split select multi answer to verify if each still in choices

### LINKED ISSUE #473: Select multiple answer is cleared if itemset is used and more than one choice is selected
#### Software versions 
JavaRosa v2.15.0

#### Problem description
When a select multiple is dynamically generated, it is impossible to select multiple choices because the answer gets cleared.

See https://forum.opendatakit.org/t/problem-with-select-multiple-with-odk-collect-v1-23-1-and-v1-23-0/21242 for sample form and reproduction steps.

lognaturel:
Closes #473

I completely forgot that multiple selects use the same code paths as single selects and that dynamic multiple selects are possible/valuable when doing #437. 

Added automated tests. Tried the form from the forum post in Collect with patched JavaRosa.

The other solution I seriously considered was reverting #437 and coming back to the issue with a broader fix. I think that if reviewers are not feeling great about this approach, that would still be a reasonable alternative.

I still think that there's value to removing values that are no longer applicable from the answer.

Some decision points that would be good to get a check on:
* How to differentiate between single and multiple selections ([here](https://github.com/opendatakit/javarosa/compare/master...lognaturel:issue-473?expand=1#diff-b60f702426880ccc5f1b079019104bd4R1040) and [here](https://github.com/opendatakit/javarosa/compare/master...lognaturel:issue-473?expand=1#diff-b60f702426880ccc5f1b079019104bd4R1090)). I used `instanceof` which is always a code smell. I think it's possible to use the reference to look up the question type in the formdef but that seemed to me like it would be harder to understand. Other ideas welcome.
* `getFilteredSelections` is an odd method that certainly feels out of place. I think there's a good change that all of this will be changed again soon so I didn't want to change the public interface of `MultipleItemsData`.
* I tried to make sure this would be as fast as possible even with many selections which is why I used a map to represent the original answer and whether its components are still in the choices. Looking things up in the keyset at each iteration is O(1). This is another place where constructing an index of underlying value to choice as (one of the things that) @ggalmazor has proposed in #438 and @JohnTheBeloved is continuing to explore could be beneficial.

It should only fix the bug described in #473 and expand the fix from #437 to select multiples. This code affects all selects with choice filters *as well as ranking* so those are at regression risk.

See issue and test.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -1031,15 +1031,21 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
             throw new XPathException("Could not find references depended on by" + itemset.nodesetRef.getInstanceName());
         }
 
-        // Get the answer to the current question so that it can be compared with the new choice list. If the answer
-        // to the question is no longer in the choice list, it will be cleared.
-        String currentQuestionAnswer = null;
+        // Get the answer to the current question to remove selection(s) that are no longer in the choice list.
+        Map<String, Boolean> currentAnswersInNewChoices = null;
         IAnswerData rawValue = getMainInstance().resolveReference(curQRef).getValue();
         if (rawValue != null) {
-            currentQuestionAnswer = rawValue.getDisplayText();
+            currentAnswersInNewChoices = new HashMap<>();
+
+            if (rawValue instanceof MultipleItemsData) {
+                for (Selection selection : (List<Selection>) rawValue.getValue()) {
+                    currentAnswersInNewChoices.put(selection.choice != null ? selection.choice.getValue() : selection.xmlValue, false);
+                }
+            } else {
+                currentAnswersInNewChoices.put(rawValue.getDisplayText(), false);
+            }
         }
 
-        boolean currentAnswerIsInNewChoices = false;
         for (int i = 0; i < matches.size(); i++) {
             TreeReference item = matches.get(i);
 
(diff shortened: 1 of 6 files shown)


lognaturel:
@ggalmazor, @grzesiek2010, it would be really ideal to get this to QA during your day on Monday if one or both of you have time to look at it. Please feel free to push to my branch if you see changes you'd like to make. Thanks!

lognaturel:
@grzesiek2010, @ggalmazor and I discussed in Slack ([here](https://opendatakit.slack.com/archives/C35ENHA64/p1565016258085900)). They've both taken a look and feel this is safe. We will release now so users can resume data collection and QA will take a look first thing their morning to see if we missed anything.

lognaturel:
I'm doing some additional sanity checks and wonder whether https://forum.opendatakit.org/t/problem-with-constraint-in-odk-v-1-23-1/21257 might be related.

lognaturel:
I have confirmed that the issue in https://forum.opendatakit.org/t/problem-with-constraint-in-odk-v-1-23-1/21257 is also caused by #437. I haven't dug any deeper to understand what's going on but it's clear that #437 is much riskier than I had anticipated. Let's revert it for now and re-evaluate how to proceed after that.

lognaturel:
I had reverted to the commit before this one so actually all I'd confirmed was that the bad commit came at or after this one. It didn't make much sense how the two would be related and indeed they aren't.

I think we learn more by merging this and being ready to revert both this and #437 if we discover any related problems. I think we are likely ok, now.

PR #478, opened 2019-08-12
Interpret times using the current date rather than January 1, 1970

grzesiek2010:
Closes #https://github.com/opendatakit/collect/issues/3305

I tested the case described in https://github.com/opendatakit/collect/issues/3305
also, I confirmed that it doesn't cause regression here https://github.com/opendatakit/collect/issues/170

Another solution I was thinking about was using always 01/01/1970 (in COllect as well) it would fix the issue but it's a bit worse in my opinion because changing timezones it would cause different time changes than we have now (because they were different in 1970).
In most cases, the current date is ok so my solution LFTM.

Time is a tricky case because of timezones and daylight saving. We can't be sure if it's the last problem in this area. The change is pretty small so I think testing the described case and related already fixed problems like https://github.com/opendatakit/collect/issues/170 would be enough.

Any form with time question like the one attached to the issue.

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/data/test/TimeDataLimitationsTest.java
+++ b/src/test/java/org/javarosa/core/model/data/test/TimeDataLimitationsTest.java
@@ -1,0 +1,81 @@
+package org.javarosa.core.model.data.test;
+
+import org.javarosa.core.model.data.TimeData;
+import org.javarosa.core.model.utils.DateUtils;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.TimeZone;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ This test is intended to show the limitation of {@link TimeData}.
+
+ Using this data type in countries which change their time (summer/winter - DST) will cause that forms saved during
+ wintertime and then edited during summertime (and vice versa) will be treated as saved in a neighbor's timezone.
+ It's because we have just time and time offset like: 10:00:00.000+02:00 but we don't know when the form has been
+ saved so we parse it using the current date.
+
+ Example:
+ If we saved 10:00:00.000+02:00 during summertime (in Poland) and we are editing the form during winter time our
+ timezone is +01:00 not +02:00. As mentioned above javarosa doesn't know that the form has been saved in the same location
+ but different timezone because of DST so it treats the value like saved in the neighbor timezone
+ (in Kiev or London for example).
+
+ Related issues:
+ https://github.com/opendatakit/javarosa/pull/478
+ https://github.com/opendatakit/collect/issues/170
+ */
+public class TimeDataLimitationsTest {
+
+    private TimeZone backupTimeZone;
+
+    @Before
+    public void setUp() {
+        backupTimeZone = TimeZone.getDefault();
+    }
+
+    @After
+    public void tearDown() {
+        TimeZone.setDefault(backupTimeZone);
+    }
+
+    @Test
+    public void editingFormsSavedInDifferentTimezoneTest() {
+        // A user is in Warsaw (GMT+2) saved a form with the time question
+        TimeZone.setDefault(TimeZone.getTimeZone("Europe/Warsaw"));
+        String savedTime = "10:00:00.000+02:00";
+
+        // A user opens saved form in Warsaw as well - the hour should be the same
+        TimeData timeData = new TimeData(DateUtils.parseTime(savedTime));
+        assertEquals("10:00", timeData.getDisplayText());
+
+        // A user travels to Kiev (GMT+3) and opens the saved form again - the hour should be edited +1h
+        TimeZone.setDefault(TimeZone.getTimeZone("Europe/Kiev"));
+        timeData = new TimeData(DateUtils.parseTime(savedTime));
+        assertEquals("11:00", timeData.getDisplayText());
+    }
+
+    @Test
+    public void editingFormsSavedInTheSameLocationButAfterDSTChangeTest() {
+        DateUtils.DateFields dateFields = new DateUtils.DateFields();
+        dateFields.year = 2019;
+        dateFields.month = 8;
+        dateFields.day = 1;
+
+        // A user is in Warsaw (during summer time - GMT+2) and saved a form with the time question
+        TimeZone.setDefault(TimeZone.getTimeZone("Europe/Warsaw"));
+        String savedTime = "10:00:00.000+02:00";
+
+        // A user opens saved form in Warsaw and during summertime as well - the hour should be the same
+        TimeData timeData = new TimeData(DateUtils.parseTimeWithFixedDate(savedTime, dateFields));
+        assertEquals("10:00", timeData.getDisplayText());
+
+        // A user opens saved form in Warsaw as well but during wintertime - the hour is edited -1h (the mentioned limitation)
+        dateFields.month = 12;
+        timeData = new TimeData(DateUtils.parseTimeWithFixedDate(savedTime, dateFields));
+        assertEquals("09:00", timeData.getDisplayText());
+    }
+}
--- a/src/test/java/org/javarosa/core/model/utils/test/DateUtilsTests.java
+++ b/src/test/java/org/javarosa/core/model/utils/test/DateUtilsTests.java
@@ -129,56 +129,64 @@ public class DateUtilsTests {
 
         TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
 
-        testTime("10:00", 1000 * 60 * 60 * 10 - getOffset());
-        testTime("10:00Z", 1000 * 60 * 60 * 10);
+        Calendar midnight = Calendar.getInstance();
+        midnight.set(Calendar.HOUR_OF_DAY, 0);
+        midnight.set(Calendar.MINUTE, 0);
+        midnight.set(Calendar.SECOND, 0);
+        midnight.set(Calendar.MILLISECOND, 0);
 
-        testTime("10:00+02", 1000 * 60 * 60 * 8);
-        testTime("10:00-02", 1000 * 60 * 60 * 12);
+        long midnightDate = midnight.getTime().getTime();
 
-        testTime("10:00+02:30", 1000 * 60 * (60 * 10 - 150));
-        testTime("10:00-02:30", 1000 * 60 * (60 * 10 + 150));
+        testTime("10:00", midnightDate + 1000 * 60 * 60 * 10 - getOffset());
+        testTime("10:00Z", midnightDate + 1000 * 60 * 60 * 10);
+
+        testTime("10:00+02", midnightDate + 1000 * 60 * 60 * 8);
+        testTime("10:00-02", midnightDate + 1000 * 60 * 60 * 12);
+
+        testTime("10:00+02:30", midnightDate + 1000 * 60 * (60 * 10 - 150));
+        testTime("10:00-02:30", midnightDate + 1000 * 60 * (60 * 10 + 150));
 
         TimeZone offsetTwoHours = TimeZone.getTimeZone("GMT+02");
 
         TimeZone.setDefault(offsetTwoHours);
 
-        testTime("10:00", 1000 * 60 * 60 * 10 - getOffset());
-        testTime("10:00Z", 1000 * 60 * 60 * 10);
+        testTime("10:00", midnightDate + 1000 * 60 * 60 * 10 - getOffset());
+        testTime("10:00Z", midnightDate + 1000 * 60 * 60 * 10);
 
-        testTime("10:00+02", 1000 * 60 * 60 * 8);
-        testTime("10:00-02", 1000 * 60 * 60 * 12);
+        testTime("10:00+02", midnightDate + 1000 * 60 * 60 * 8);
+        testTime("10:00-02", midnightDate + 1000 * 60 * 60 * 12);
 
-        testTime("10:00+02:30", 1000 * 60 * (60 * 10 - 150));
-        testTime("10:00-02:30", 1000 * 60 * (60 * 10 + 150));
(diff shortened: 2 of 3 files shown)


lognaturel:
I've only taken a quick look and I don't understand the fix at first glance but @ggalmazor has spent a lot of time with timezones so I'm hoping he will. I think the description of current behavior at https://github.com/opendatakit/collect/issues/3305#issuecomment-520533726 is key.

ggalmazor:
Hi, @grzesiek2010!

Thanks for taking a stab at this. The date & time parsing code in JR needs a lot of attention and it's a shame that we can't use the new Java8 dates API yet...

Thanks for documenting the issue with such detail in collect and here. It helps a lot to understand the problem.

I think that using the current date helps avoiding the issue in general, but I think it's still vulnerable when the current date coincides with a DST change day. For example, in Spain, that's March 31st and Octobre 27th.You describe these steps to reproduce the issue:

1. Open the attached form
2. Add any answer
3. Finalize the form
4. Open the saved instance (don't click on the question - important)
5. Go to the end of the form and save it
6. Open the saved form again - time will be changed +1h

I think I could reproduce the issue on your fixed code if I do (in Spain):

- I do step 2 October 27th. I set my time answer to 02:00, which produces `2019-10-27T02:00:00.000+02:00`
- I do step 3 the same day and `02:00+02:00` is saved
- I do step 4 **the next day** and follow through the rest of steps

  We should see that the submission gets a wrong `02:00+01:00` answer.

Do you agree with this?

ggalmazor:
Here's a test that reproduces the scenario I mention in my previous comment:

https://gist.github.com/ggalmazor/f519d16ed50540fc563f9ff7fa6bdc99

ggalmazor:
Honestly, the more I think about it, the less sure I am about what would be a wrong answer in this context. 

For a `02:00+02:00` answer, internally having `2019-10-28T02:00:00.000+02:00` is wrong because that value doesn't exist in Spain. The correct internal value would be `2019-10-28T02:00:00.000+01:00` which doesn't match the input `02:00+02:00`.

Unless someone can explain it better, I don't even understand how I'm supposed to think about how JR parses time.

grzesiek2010:
Of course, it would be much easier if we could store just plaint time like `08:00` without parsing it but I doubt it's possible. Users might need it.

When it comes to the test you described in your first comment... you chose a tricky case because the time is set from 3AM to 2AM so 2AM appears two times that day once for GMT+02 and once for GMT+01 right? I did the same and I got `02:00:00.000+01:00`
could you confirm what was the result time after reproducing my steps in your first comment? Because maybe I missed something.

ggalmazor:
> Of course, it would be much easier if we could store just plaint time like 08:00 without parsing it but I doubt it's possible. Users might need it.

My point is (and it's not a question just for you, but for us to discuss) is that what does `08:00+02:00` even mean without the context of a date. Such an answer can't be understood without giving more context and I'm wondering whether forms would be much better off by asking for a date&time, or by having a date field, a location field, and a time field separately.

> I did the same and I got 02:00:00.000+01:00

That's the expected result if we use the current date to parse time values. Did you check the test on the gist I linked? It reproduces this issue in a deterministic way. https://gist.github.com/ggalmazor/f519d16ed50540fc563f9ff7fa6bdc99

If you think about it, it's not such a tricky case. With so many DST zones it will eventually happen in a real form.

lognaturel:
I now remember this issue. I think a good title for this PR would be "Interpret times using the current date rather than January 1, 1970" right?

As I understand it, the idea behind using current date to interpret a time is that because timezones change only twice a year in places with DST, it's likely that the timezone hasn't changed between when a form was first filled and then reopened/viewed on Google Sheets. In other words, by using the current date, it's more likely that the times will be interpreted the same way. On the other hand, using Jan 1, 1970 guarantees an inconsistent interpretation of time on form re-open half of the year for locations with daylight savings time. So this is likely to be an improvement in general except, as @ggalmazor notes, in the case where data is collected around a DST boundary.

This does not make any changes to underlying data, only to what time is displayed.

We've talked about it before and what we should absolutely make sure we do is document the outcome of this discussion so that we can point to it next time there's a problem around timezone issues with times. https://github.com/opendatakit/javarosa/pull/51 has a previous discussion. Here's https://docs.opendatakit.org/form-question-types/#time-widget with some user-facing documentation but I think we should also have a developer-facing summary of decisions, probably in the code and linking to https://github.com/opendatakit/collect/issues/170, https://github.com/opendatakit/collect/issues/120 and other related conversations.

As @ggalmazor points out, there's no perfect solution here. Fundamentally, a time with timezone information and no date information can't be consistently displayed always.

And just for fun, here is @grzesiek2010 saying that this solution wouldn't be a good one: https://github.com/opendatakit/collect/issues/170#issuecomment-319374291 😅

grzesiek2010:
ahhh.... ok so there is no perfect solution. A perfect solution would be using date + time like in dateTime widget (as @ggalmazor noticed) or just simple time like 08:00 without parsing it. @lognaturel do you think such a  big change is possible to implement?

lognaturel:
Here's where @grzesiek2010 previously considered this change but didn't make it: https://github.com/opendatakit/javarosa/pull/189#issuecomment-329746860

> just simple time like 08:00 without parsing it. 

The challenge with something like this is that some users would end up with data in both formats for a single data collection campaign. I don't think that's acceptable. I think what we'd need to do is to introduce a new data type.

ggalmazor:
:+1: to introduce a new data type :)

I agree that using the current date to parse times gets us closer to an optimal solution and that we can document the expected outliers on DST boundaries. 

In addition to the change in the docs, I think that it would be best to have a test like the one I've linked but I really would want to avoid adding mockito/powermock to the testing stack. We could avoid that  by making a small change to the DateUtil class so that it uses Calendar instances. This would improve the overall design and let us set arbitrary dates from tests without mocking stuff.

lognaturel:
I think we're all on the same page. 😊 We all agree that time with an offset but no date is impossible to interpret consistently. We may eventually introduce a new datatype for time without offset but not today. So in the mean time, let's make sure we make the best of this existing datatype and document the expected behavior and past conversations as well as we can.

> I agree that using the current date to parse times gets us closer to an optimal solution

It does look that way. @grzesiek2010, can you think of why we didn't do it that way last time we were in these parts of the code? In particular, I'm wondering about https://github.com/opendatakit/javarosa/pull/189#issuecomment-329746860.

Can you also please explain why you had to add `midnightDate` to each test's expected value?

> I think that it would be best to have a test like the one I've linked but I really would want to avoid adding mockito/powermock to the testing stack. We could avoid that by making a small change to the DateUtil class so that it uses Calendar instances.

This makes sense. The idea would be to add a new variant that a `Calendar` can be passed into for testing and not change the existing API.

grzesiek2010:
> can you think of why we didn't do it that way last time we were in these parts of the code? In particular, I'm wondering about #189 (comment).

I can't recall, but probably I just thought it's a perfect solution and didn't know about this problem. 

> Can you also please explain why you had to add midnightDate to each test's expected value?

As you know my change is that now the time is parsed using current dates not 1970-01-01. So for example:
`testTime("10:00Z", 1000 * 60 * 60 * 10);` means 1970-01-01 10:00Z = 1000 * 60 * 60 * 10
but if I want to parse it using current dates I need to add the date in miliseconds for midnight and then add miliseconds for hours.
`testTime("10:00Z", midnightDate + 1000 * 60 * 60 * 10);`

lognaturel:
Thanks for the explanation. To summarize:

> testTime("10:00Z", 1000 * 60 * 60 * 10);

This gets us number of milliseconds from midnight in winter (non-daylight-savings) time. Because 1970/1/1 is epoch, the "day" component of the time is essentially cancelled out.

> testTime("10:00Z", midnightDate + 1000 * 60 * 60 * 10);

The date component representing the current date is now in the parsed time. In real usage in `XFormAnswerDataParser`, the result of `parseTime` is immediately made into a `TimeData` which loses the date component. It feels weird that `parseTime` would include a date component but I'm not really seeing a better way to make this change. What do you think @ggalmazor?

I think that in addition to the tests @ggalmazor has mentioned, the two tests that relate to time interpretation with DST (`DateUtilsSCTOTests.testParseTime_with_DST` and `DateUtilsTests.testParseTime_with_DST`) should be re-established with updated comments. Probably we should have given more thought to why those tests existed when they were ignored.

grzesiek2010:
I was thinking about this problem more, trying to understand it better... I hate all those time-related issues.
I wanted to add more tests as you asked but I think it doesn't make sense to test this change particularly with DST because the test I fixed https://github.com/opendatakit/javarosa/pull/478/files#diff-b1b9ae1026c082689972b37b877ee5bd does the job.
Let's don't think about any particular place in the world whether it's Spain, Poland or whatever.
If we have saved time eg. `10:00+02` and time change takes place (summer->winter or winter->summer) it changes our timezone so it is like changing our location.

DST will work in the same way like just traveling to another timezone physically.
Of course, a user will know: `hey I'm still in the same place where I have saved that form`, but javaraosa just doesn't know about it and when DST happens it will treat saved forms like they were saved in another location.

so since that data type doesn't contain any info about **date** or **location** both cases are just the same and that one test is enough.

@lognaturel @ggalmazor If I don't understand something let me know if you have any tests you can also push them here I won't be angry :smile:  but honestly I don't think we need more here.

lognaturel:
> DST will work in the same way like just traveling to another timezone physically.

Right, that's a good way to think about it. I think you're right that this means the test @ggalmazor proposed isn't what we want.

Is this correct:
- Prior to the change, a date saved during winter time would always be correctly written back out after an edit 
- After the proposed change, a date saved during winter time would only be correctly written out after an edit if the edit also happened in winter time
- Prior to the change, a date saved during summer time would never be written back out correctly (because 1/1/1970 was winter time)
- After the proposed change, a date saved during summer time would only be correctly written out after an edit if the edit also happened in summer time

This change assumes that the user is likely to make edits in close time proximity to when the form was originally saved.

What would be great are tests that show this change in behavior. That is, a test that fails on master, passes on this branch and actually shows what the user would experience rather than what `testTime` does which is looking at internals.

ggalmazor:
> What would be great are tests that show this change in behavior. That is, a test that fails on master, passes on this branch and actually shows what the user would experience rather than what testTime does which is looking at internals.

This sounds great :) @grzesiek2010, please, let me know if I can help you with that test. Maybe something from the test of the gist I linked could be reused. I remember we discussed making some improvements to the date&time parsing internals to make it easier to test i.e. overloading some methods to inject a Calendar instance that could be changed from tests to avoid having to using Powermock and co.

ggalmazor:
Do you want me to stand down reviewing the code of this PR or should I jump in now? I see that the commits are from Aug 12, and I guess we could expect more commits comming next, right?

src/test/java/org/javarosa/core/model/utils/test/DateUtilsTests.java
```diff
@@ -129,56 +129,64 @@ public void testTimeParses() {
 
         TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
 
-        testTime("10:00", 1000 * 60 * 60 * 10 - getOffset());
-        testTime("10:00Z", 1000 * 60 * 60 * 10);
+        Calendar midnight = Calendar.getInstance();
```
ggalmazor: This would commonly be called `startOfDay` rather than `midnight` (see [Java8's `LocalDate` API](https://docs.oracle.com/javase/8/docs/api/index.html))

grzesiek2010:
Generally, your cases are correct.

The good news is that countries, where ODK Collect, is most popular don't have DST:
https://en.wikipedia.org/wiki/Daylight_saving_time_by_country#/media/File:DST_Countries_Map.png

> What would be great are tests that show this change in behavior. That is, a test that fails on master, passes on this branch and actually shows what the user would experience rather than what testTime does which is looking at internals.

ok, such a short test which shows the limitation should be good.

lognaturel:
> The good news is that countries, where ODK Collect, is most popular don't have DST

OH! That explains why we don't get a lot of complaints. VERY interesting.

We don't need to discuss this further now but I'm leaning towards proposing we drop the timezone component for time for a JavaRosa 3 / Collect 2 release.

ggalmazor:
> JavaRosa 3 / Collect 2

:rainbow: :star2: :grinning: :star2: :rainbow:

I'm very interested in this. Has anything been discussed yet?

lognaturel:
> I'm very interested in this. Has anything been discussed yet?

No. 😂😭

ggalmazor:
I'll leave merging the PR to @lognaturel or @grzesiek2010 ;)

grzesiek2010:
I'm on it it should be ready tomorrow in the worst case.

grzesiek2010:
Ok, I added a test that shows and explains the limitation so I think it's ready.

src/test/java/org/javarosa/core/model/data/test/TimeDataLimitationsTest.java
```diff
@@ -0,0 +1,67 @@
...
+ timezone is +01:00 not +02:00. As mentioned above javarosa doesn't know that the form has been saved in the same location
+ but different timezone because of DST so it treats the value like saved in the neighbor timezone
+ (in Kiev or London for example).
+
+ Related issues:
+ https://github.com/opendatakit/javarosa/pull/478
+ https://github.com/opendatakit/collect/issues/170
+ */
+public class TimeDataLimitationsTest {
+
```
ggalmazor: We should add a @Before and @After that ensures each test runs in isolation and has no side-effects for other tests.

```java
private TimeZone backupTimeZone;

@Before
public void setUp() {
    backupTimeZone = TimeZone.getDefault();
}

@After
public void tearDown() {
    TimeZone.setDefault(backupTimeZone);
}
```

grzesiek2010: Fixed.

PR #483, opened 2019-09-11
Issue 482 get list of functions without handlers

### LINKED ISSUE #482: Provide the list of custom function names that are missing a handler
Clients using JavaRosa like Briefcase and Validate need to work with forms that are designed to work with third-party clients that might use custom functions that are not defined in the xforms specs.

https://github.com/opendatakit/briefcase/issues/796 is an example of one of those custom functions causing problems in Briefcase.

One solution for this is to provide dummy function handlers, but for that, JavaRosa users need to know the list of custom functions that require one.

We should add a method that returns a `List<String>` of function names that are missing a handler for a particular form.

I think a good spot for this method would be the `FormDef` class, but we might discover better places as we implement this.

A source for a list of functions would be the triggerables DAG located at `FormDef.dagImpl.triggerablesDAG`. We could traverse it in search of `XPathFuncExpr` objects and build a list of function names with those. Then we can filter out those that have a handler in `FormDef.exprEvalContext.functionHandlers`. The resulting list would contain functions without handlers.

ggalmazor:
Closes #482

Added automated test and improved general test coverage

There could be better ways to get the list of functions without handlers, although having an enum with all built-in functions feels like a must-have in any case.

This is an additive change that shouldn't affect users because no one would be using it.

The changes to the XPathFuncExpr class are covered by several tests, therefore, they should be safe.

No.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -16,6 +16,19 @@
 
 package org.javarosa.core.model;
 
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Set;
 import org.javarosa.core.log.WrappedException;
 import org.javarosa.core.model.IDag.EventNotifierAccessor;
 import org.javarosa.core.model.actions.Action;
@@ -67,20 +80,6 @@ import org.javarosa.xpath.XPathException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.DataInputStream;
-import java.io.DataOutputStream;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Set;
-
 /**
  * Definition of a form. This has some meta data about the form definition and a
  * collection of groups together with question branching or skipping rules.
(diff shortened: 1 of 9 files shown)


lognaturel:
This doesn't get a list of functions without handlers, it gets a list of functions that are not the built-in functions, right? That is, if the client does happen to define some handlers, the functions it defines them for will still be in this list.

When I was thinking about this option at https://github.com/opendatakit/briefcase/issues/796#issuecomment-530135378, I was imagining that the built-in functions were in the handler list just like the client-defined ones. But of course that's not the case. That means that unfortunately getting this list requires pre-processing.

Did you consider the second concept I mentioned which is to pass in a flag for ignoring unknown functions? I imagine that would come in to one of the `XFormUtils` methods and that somehow it would have to make its way to `eval` (this is the part that may make this approach unreasonable). Then the last branch could do nothing instead of throwing a `XPathUnhandledException`. The two big advantages are that no pre-processing is needed and that it's a simpler interface for clients to use.

ggalmazor:
> This doesn't get a list of functions without handlers, it gets a list of functions that are not the built-in functions, right? That is, if the client does happen to define some handlers, the functions it defines them for will still be in this list.

Ugh, you're right. I forgot to filter functions using `exprEvalContext.getFunctionHandlers()`

lognaturel:
Not a big deal! Let's first make sure this is the best possible API for clients.

ggalmazor:
> That means that unfortunately getting this list requires pre-processing.

I think that getting a list of functions will always require pre-processing, but this only happens if/when a client calls `FormDef.getFunctionsWithoutHandlers()`. We can even cache it to ensure we ever do that once.

ggalmazor:
> Did you consider the second concept I mentioned which is to pass in a flag for ignoring unknown functions? 

I have to say I didn't :P

I think `FormDef.getFunctionsWithoutHandlers()` would be a better addition to the API than overloading initialization methods to avoid throwing if a function handler is missing. It sounds like we could find other reasons to ignore failures and keep on overloading... 

I too don't like the idea of propagating all the required flags down to the `eval` part that would conditionally throw an exception.

ggalmazor:
Maybe I went too far with the BoyScout rule in this PR. If it feels risky, I could produce a streamlined version that doesn't make so many changes to `XPathFuncExpr`.

I felt encouraged to make so many changes due to the great test coverage we already have but in hindsight, it might feel too much.

lognaturel:
> It sounds like we could find other reasons to ignore failures and keep on overloading...

That's true. Ok, how about yet another concept -- allowing clients to register a default handler for the fallback case. That could allow a client to get the unhandled function names, ignore them, log to its own logging subsystem, whatever. That gives more power to the client and doesn't require pre-processing.

> Maybe I went too far with the BoyScout rule in this PR.

For now I'm more concerned about figuring out the best API in terms of simplicity to use on the client side and performance! I think this level of cleanup is likely ok if we determine it to be the best approach.

lognaturel:
Replaced by #484.

PR #486, opened 2019-09-13
Allow relative references as jr:choice-name arguments

### LINKED ISSUE #435: jr:choice-name() broken when used in repeats
#### Software versions 
JavaRosa v1.x.x, Java v1.x.x, operating system

#### Problem description
jr:choice-name() always returns nothing when used against a select question inside a repeat loop. Probably due to the relative reference produced by pyxform for the quoted select question reference

ODK Forum discussion here: https://forum.opendatakit.org/t/problem-with-odk-calculation-field/19619

same form available from: https://forum.opendatakit.org/t/problem-with-odk-calculation-field/19619
#### Other information 
Things you tried, stack traces, related issues, suggestions on how to fix it...

lognaturel:
Closes #435

- Adds a method to set the current language to `Scenario`
- Anchors XPath path strings passed in to `jr:choice-name` (#435)
- Adds test for `jr:choice-name` and attempts to improve comments

Reviewer: the most important thing is to think of other classes of input to `jr:choice-name` that I may not have considered.

Added automated tests. The cases I could think of are `jr:choice-name` calls at the top level, in a repeat and in a multi-language context. 

I added a test with a dynamic itemset to make it clear that it's not a supported case. I think this could intersect with #436 because doing something like computing the unfiltered list once and then filtering as needed or maintaining the filtered list until one of the values in the filter predicate changes could address this problem. 

Note to self for future exploration for supporting dynamic itemsets:
Conceptually, it seems to me like `jr:choice-name` should be able to use the question XPath path (second argument) to generate the correct filtered choices. In fact, I can see that when the handler calls `populateDynamicChoices`, that question node is the context node. Everything is as I would expect it until `EvaluationContext.expandReferenceAccumulator` evaluates the predicate and does the call on `rescope`. It's the call on `eval` right after that which fails because the context node is `instance(cities)/root/item`. I think it would be helpful to compare this with a call to `populateDynamicChoices` that does work as expected.

The only other option I considered was maintaining the previously-documented limitation that the path provided must be absolute. The change is https://github.com/opendatakit/javarosa/pull/486/commits/437969cd86fc32c7faa52fc09a3385ac3e9e3561. I can't see any reason why relative references shouldn't be allowed. The context node is the node that the `jr:choice-name` expression appears in which seems like exactly what would be expected.

I decided to add more tests and comments to hopefully help whoever is here next. I also wanted to be sure to understand as much of the intended behavior and limitations of `jr:choice-name` as possible to evaluate risk.

The intent is only to allow relative references as `jr:choice-name` arguments. There's a risk that I changed the behavior of `jr:choice-name` for some case I haven't anticipated. 

https://github.com/opendatakit/javarosa/pull/486/files#diff-3a63dc2cd5c82a90b2ee226249303d8e

No, the absolute path limitation looks like it was never documented.


lognaturel:
@ggalmazor The desire is to get this in Collect v1.24 which we hope to freeze by October 9. So no big rush. I'll probably try to get another couple of JR fixes in if you want to review them together.

ggalmazor:
Hi!

I've been with this for some time and the changes themselves are LGTM.

I've been thinking about other use cases that might be outside the scope of the tests you've added and the only thing I could come up with was passing an xpath function expression as the value to `jr:choice-name`. Not really sure how this would translate to real user needs, though (just looking for untested holes).

I wrote a failing test that concatenates a couple of field values to get the value for jr:choice-name - https://github.com/ggalmazor/javarosa/commit/88dae27bc14e953902691fdd8b93668bff70153e

I'm getting a type mismatch error. This is expected because the function handler declares just a variant with a couple `String.class` args, but now I'm confused because, why passing an xpath ref works with that prototype declaration? (See https://github.com/opendatakit/javarosa/blob/master/src/main/java/org/javarosa/core/model/FormDef.java#L952)

lognaturel:
I think 7932e6b might be the test you were trying to write.

I'm afraid `jr:choice-name` is far from intuitive. The first thing to understand about it is that it's looking up the LABEL for a select from that select's value. Confusingly, in forms generated by XLSForm, the value is defined by a node called `name`. That's what `<value ref="name"/>` does in the `itemset` declaration. (You know this but just being explicit)

The first parameter represents the underlying select value. In a typical form definition, this will be specified as an XPath path because what the form designer wants to get the label for is some selection that the user made. For example, if there's a question at path `/data/city`, the form designer wants to pull the translated, user-visible name of the city that the user chose. But the first argument could also be a literal string and that's what the first test I wrote does. It could also be the result of some expression that evaluates to a string like what you added.

The second parameter is a string that represents the XPath path of the node that defines the select. This is needed because in a case where the value that's being looked up is a literal string or the result of some expression like in your test, the function has no idea where to look up the label. I think that even in the case where a path to a select question is passed in as the first parameter, it's resolved to a string before it gets to the handler so there's no way to know where to look the value up either. So we end up with a very weird function signature indeed.

I think your test is useful for various reasons. I added it and some more comments to the form.

PR #488, opened 2019-09-16
Switch to AdoptOpenJDK for license compliance

### LINKED ISSUE #487: Build JavaRosa with OpenJDK and update the readme

yanokwa:
Closes #487. We should issue a point release with this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Deviations from OSSRH's documentation are that maintainers use `gpg2` v2.1 and g
 </settings>
 ```
 
-Official releases are typically done by @yanokwa who uses Oracle Java v1.8.x running on macOS Sierra. 
+Official releases are typically done by @yanokwa who uses AdoptOpenJDK v1.8.x running on macOS Sierra.
 
 To generate official signed releases, you'll need the GPG folder, GPG passwords, a configured `secrets.xml` file.
 


ggalmazor:
Just curious why AdoptOpenJDK and not Java.net's OpenJDK (or any other vendor)? Can we be less specific and just say OpenJDK 8.0.221?

Also, I think we should specify the full version needed (we know that a lower patch version of Java8 won't work well with cryptography).

-- 

So, just to have a full picture of what we're doing in other tools:

- The releases of Aggregate and JavaRosa that I've released so far, have been built with Oracle's Java8u201.
- Briefcase is being released with Java.net's OpenJDK 11.0.2 release.

yanokwa:
AdoptOpenJDK because https://openjdk.java.net only seems to have Java 12 binaries. The rest are either reference implementations for some, but not all OSes. https://adoptopenjdk.net has JDK 8, 11, and 12 and for all OSes. Seems way more straightforward for users.

Totally OK specifying the full version. Could we say "We recommend OpenJDK 11, but OpenJDK 8.0.221 or greater should work".

I think to be totally safe as far as licensing, we should build all future releases with OpenJDK. Does that seem possible/reasonable?

yanokwa:
@lognaturel and I discussed this with @ggalmazor and the agreement here is that we will build with Adopt OpenJDK, and we should suggest users run Adopt OpenJDK, but developers can use whatever OpenJDK that they want. 

Adopt OpenJDK has patches that are in Oracle Java but not in Java.net's OpenJDK. But because Adopt OpenJDK is easier to well, adopt, we are using it. It might be slightly risky because of aforementioned patches, but it's a widely-used JDK.

lognaturel:
> The releases of Aggregate and JavaRosa that I've released so far, have been built with Oracle's Java8u201.

And to be explicit, this is totally fine because Java8u201 came before Oracle's licensing requirements.

PR #493, opened 2019-09-26
Add support for odk-new-repeat event and actions nested in repeats

### LINKED ISSUE #465: Add support for the odk-new-repeat event, deprecate jr-insert
See https://github.com/opendatakit/xforms-spec/pull/241 for specification.

lognaturel:
Closes #465

Automated test with form that captures all cases mentioned in [the specification](https://opendatakit.github.io/xforms-spec/#the-odk-new-repeat-event).

#### Why is this the best possible solution? Were any other approaches considered?I followed the event/action patterns that are already established so I didn't feel like there were many decisions to make. 

The one thing I went back and forth on is whether or not to block actions that respond to `odk-new-repeat` at the model level. See the last test for an example. The specification currently says "Actions triggered by odk-new-repeat must be nested in the repeat form control." I initially was going to leave incidental support but I think it's confusing when there's unofficially supported behavior so I made that enforced on initial parse.

This should be a purely additive change making it very low risk. Some code has moved around in `XFormParser` but as far as I know, the affected nodes are for configuring a type of repeat that is undocumented and unused by Collect.

I added an action controller to `GroupDef` and I could imagine some mistakes there having negative consequences.

See test.

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/actions/OdkNewRepeatEventTest.java
+++ b/src/test/java/org/javarosa/core/model/actions/OdkNewRepeatEventTest.java
@@ -1,5 +1,6 @@
 package org.javarosa.core.model.actions;
 
+import org.javarosa.core.model.data.IntegerData;
 import org.javarosa.core.test.Scenario;
 import org.javarosa.xform.parse.XFormParseException;
 import org.junit.Test;
@@ -23,7 +24,7 @@ public class OdkNewRepeatEventTest {
         assertThat(scenario.repeatInstancesOf("/data/my-repeat").size(), is(0));
         scenario.createMissingRepeats("/data/my-repeat[0]");
         assertThat(scenario.repeatInstancesOf("/data/my-repeat").size(), is(1));
-        assertThat(scenario.answerOf("/data/my-repeat[0]/defaults-to-now"), is(notNullValue()));
+        assertThat(scenario.answerOf("/data/my-repeat[0]/defaults-to-position").getDisplayText(), is("1"));
     }
 
     @Test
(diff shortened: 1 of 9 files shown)


src/test/java/org/javarosa/core/model/actions/OdkNewRepeatEventTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Specification: https://opendatakit.github.io/xforms-spec/#the-odk-new-repeat-event.
+ */
+public class OdkNewRepeatEventTest {
+    @Test
```
ggalmazor: I think I'm coming back from the idea of statically declaring a form state in tests. I think now I'd write all the tests in a progressive way, imitating what a user would do while using Collect. I think it is closer to how JR gets used by clients and doesn't require so much hacking.

Having this into consideration, this test could be written like so:
```
@Test
public void setValueOnRepeatInsertInBody_setsValueInRepeat() {
    Scenario scenario = Scenario.init(r("event-odk-new-repeat.xml"));
    scenario.next(); // /data/my-toplevel-value
    scenario.next(); // Prompt new repeat
    scenario.createNewRepeat();

    assertThat(scenario.answerOf("/data/my-repeat[0]/defaults-to-now"), is(notNullValue()));
}
```

(`scenario.createNewRepeat();` just calls to `formEntryController.descendIntoNewRepeat();`)

src/test/java/org/javarosa/core/model/actions/OdkNewRepeatEventTest.java
```diff
@@ -0,0 +1,86 @@
...
+    public void setValueOnRepeatInsertInBody_setsValueInRepeat() {
+        Scenario scenario = Scenario.init(r("event-odk-new-repeat.xml"));
+
+        assertThat(scenario.repeatInstancesOf("/data/my-repeat").size(), is(0));
+        scenario.createMissingRepeats("/data/my-repeat[0]");
+        assertThat(scenario.repeatInstancesOf("/data/my-repeat").size(), is(1));
+        assertThat(scenario.answerOf("/data/my-repeat[0]/defaults-to-now"), is(notNullValue()));
+    }
+
+    @Test
```
ggalmazor: This one could look like this:
```java
@Test
public void addingRepeat_doesNotChangeValueSetForPreviousRepeat() {
    Scenario scenario = Scenario.init(r("event-odk-new-repeat.xml"));
    scenario.next();
    scenario.next();
    scenario.createNewRepeat();

    String firstRepeatDisplayText = scenario.answerOf("/data/my-repeat[0]/defaults-to-now").getDisplayText();
    // Go through the couple of fields in the group
    scenario.next();
    scenario.next();
    // Exit the group
    scenario.next();

    scenario.createNewRepeat();
    String secondRepeatDisplayText = scenario.answerOf("/data/my-repeat[1]/defaults-to-now").getDisplayText();

    assertThat(scenario.answerOf("/data/my-repeat[0]/defaults-to-now").getDisplayText(), is(firstRepeatDisplayText));
    assertThat(secondRepeatDisplayText, is(not(firstRepeatDisplayText)));
}
```

ggalmazor: Having narrow-focused forms, the number of calls to `scenario.next()` would decrease.

src/test/java/org/javarosa/core/model/actions/OdkNewRepeatEventTest.java
lognaturel: Agreed that's an improvement!

lognaturel:
I initially went back and forth about whether to do a single form or several smaller ones. I eventually decided on a single one because for me it is helpful to see the cases together to make sure I'm not forgetting anything and to get a good feel for the intended behavior. I don't get that as easily from the test names. I think it's also easier to understand why some of these cases are important in the context of the prior ones (e.g. with jr:template and without). My preference would be to keep a single form but please do push back if you have counters to my reasoning. If you're ok with the single form, please let me know whether you still feel like using `next` to navigate is better. What would you think about jumping to the section relevant to the current test and then using `next` as a middle ground?

I cherry picked your commit, thank you.

I also realized that a cryptic error was being shown about the action names being unknown and addressed that at 8e13cd1.

PR #495, opened 2019-09-30
Add support for nested odk-instance-first-load event and space-separated event lists

### LINKED ISSUE #466: Support multiple space-separated events as action triggers
For example: `odk-instance-first-load odk-new-repeat`

lognaturel:
Closes #466

Added test form for nested `odk-instance-first-load` event in isolation and then space-separated event lists.

First I realized that `odk-instance-first-load` only triggered actions defined in the model so 56ebdd6 adds support for top-level events on actions nested in form controls. The best way I could think to do this is to register the form controls that have actions triggered by top-level events when doing the initial parsing. Without adding that state, we'd need to traverse the entire form definition in search of responding nodes every time a top-level event is triggered. This would be ok for `odk-instance-first-load` because it would just extend the load time of the form but I thought it was worth trying to solve in a more general way. I could be convinced to reconsider.

The problematic thing about adding this state is that it needs to be serialized. It doesn't make sense to serialize `IFormElement`s because 1) they're big and 2) the deserialization process doesn't match references so there would be duplicate objects representing the same form element which seems bad. Instead I serialized the corresponding references.

I apologize if someone is offended by the name `elementsWithActionTriggeredByToplevelEvent` and am open to suggestions!

41b745c adds support for space-separated event lists. I did that by registering lists of events rather than single events. I didn't really think of any alternatives for this one.

Once #493 is merged I will add a test for the combined functionality.

The intent is for this not to change any existing behavior. It should just add support for `odk-instance-first-load` nested in form controls and space-separated event lists. Because I add state to `FormDef`, there's risk around form caching. I tried to add test coverage for that but wonder whether there are more cases than top-level question and question nested in group with relative ref that I should add.

See test.

Already documented at https://opendatakit.github.io/xforms-spec/#actions.

### DIFF
--- a/src/main/java/org/javarosa/core/model/CoreModelModule.java
+++ b/src/main/java/org/javarosa/core/model/CoreModelModule.java
@@ -29,6 +29,7 @@ public class CoreModelModule implements IModule {
             "org.javarosa.core.model.QuestionDef",
             "org.javarosa.core.model.RangeQuestion",
             "org.javarosa.core.model.GroupDef",
+            "org.javarosa.core.model.instance.TreeReference",
             "org.javarosa.core.model.instance.FormInstance",
             "org.javarosa.core.model.instance.ExternalDataInstance",
             "org.javarosa.core.model.data.BooleanData",
--- a/src/test/java/org/javarosa/core/test/Scenario.java
+++ b/src/test/java/org/javarosa/core/test/Scenario.java
@@ -16,6 +16,10 @@
 
 package org.javarosa.core.test;
 
+import static java.nio.file.Files.createTempFile;
+import static java.nio.file.Files.delete;
+import static java.nio.file.Files.newInputStream;
+import static java.nio.file.Files.newOutputStream;
 import static org.javarosa.core.model.instance.TreeReference.CONTEXT_ABSOLUTE;
 import static org.javarosa.core.model.instance.TreeReference.INDEX_TEMPLATE;
 import static org.javarosa.core.model.instance.TreeReference.REF_ABSOLUTE;
@@ -29,6 +33,9 @@ import static org.javarosa.form.api.FormEntryController.EVENT_REPEAT;
 import static org.javarosa.form.api.FormEntryController.EVENT_REPEAT_JUNCTURE;
 import static org.javarosa.test.utils.ResourcePathHelper.r;
 
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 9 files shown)


src/main/java/org/javarosa/core/model/CoreModelModule.java
```diff
@@ -29,6 +29,7 @@
             "org.javarosa.core.model.QuestionDef",
             "org.javarosa.core.model.RangeQuestion",
             "org.javarosa.core.model.GroupDef",
+            "org.javarosa.core.model.instance.TreeReference",
```
lognaturel: `TreeReferences` are already serialized in e.g. `XPathReference` or `Triggerable` but the type is passed inline. I think this is necessary to have a collection of `TreeReference`s that is serialized.

src/test/java/org/javarosa/core/test/Scenario.java
```diff
@@ -502,4 +515,25 @@ public static AnswerResult from(int jrCode) {
...
+
+        // Create an empty FormDef and deserialize the form into it
+        FormDef deserializedFormDef = new FormDef();
+        deserializedFormDef.readExternal(
+            new DataInputStream(newInputStream(tempFile)),
+            PrototypeManager.getDefault()
+        );
+
+        delete(tempFile);
+        return new Scenario(deserializedFormDef, new FormEntryController(new FormEntryModel(deserializedFormDef)));
```
lognaturel: Not really sure why the constructor takes in a FormEntryController since it's derived from a FormDef but it was already like that and I didn't change it.

ggalmazor: It's just because normally I wouldn't have any kind of computation in constructor methods and, since the controller is a final member, it has to be provided.

If creating an object from any set of arguments requires computation, I normally have a factory methods that deals with the computation and provides all the arguments for the single class contructor.

ggalmazor:
> First I realized that odk-instance-first-load only triggered actions defined in the model so 56ebdd6 adds support for top-level events on actions nested in form controls. The best way I could think to do this is to register the form controls that have actions triggered by top-level events when doing the initial parsing. Without adding that state, we'd need to traverse the entire form definition in search of responding nodes every time a top-level event is triggered. This would be ok for odk-instance-first-load because it would just extend the load time of the form but I thought it was worth trying to solve in a more general way. I could be convinced to reconsider.

I agree with your assessment and I can't provide any better alternative, so I say let's go with it.

> I apologize if someone is offended by the name elementsWithActionTriggeredByToplevelEvent and am open to suggestions!

:man_shrugging: eh, seen worse.

src/test/java/org/javarosa/core/test/Scenario.java
lognaturel: > normally I wouldn't have any kind of computation in constructor methods

Ahh, interesting. That makes sense.

I think it's ok to leave like this for now.

lognaturel:
Thanks! I cherry picked your two commits and added a test form with bad event names to document the exception and ensure it is maintained. If you're happy with this, please merge!

PR #505, opened 2019-10-22
Check source code against the minimum Android API Level Collect supports

ggalmazor:
Just trying stuff out. Will complete this template later

Closes #


lognaturel:
Neat! It seems heavy to have to build an Android project but I can't think of any other strategy. The error messages are very helpful and familiar to Android devs so that's great. I say let's go with it and just make sure there's some documentation on it to e.g. remind us to change the minSdk as needed.

ggalmazor:
Awesome! Now that this looks like a good idea, I'd throw a little improvement in to make the `minSdk` an argument of the androidLint task, so that we can configure it at CI conf level.

ggalmazor:
OK, I think this PR is ready now. I had to work around some issues. I'll document them here just in case:
- In order to set the `minSdkVersion` value in the Android project, using a `filter` method of the Gradle `Copy` task type was the cleanest option but it turned out it was corrupting some binary files in the project.

  I'm sure there's a safe way to do it, but I was spending too much time and I finally reverted to do a search&replace on the build file's contents instead.
- I took from Aggregate the system to get values from build properties that might or might not be defined in `gradle.properties`. I added a doc block explaining the situation.

  Everything is rigged to work with defaults (defaulting to SDK level 16) so that cloning the repo is enough to build and run the `androidLint` task.
- Had to add `--max-workers=1` to the `androidLint` task because there`s a known issue when launching Android builds from Java projects (see [thread](https://issuetracker.google.com/issues/80464216))

lognaturel:
Having this in CI will bring some good peace of mind!

PR #507, opened 2019-10-30
WIP: 422 position predicate shortcut

### LINKED ISSUE #422: Add support for position predicate shortcut
A lot of XPath examples show expressions like `/path/to/some/node[4]` which is a shortcut for `/path/to/some/node[position() = 4]`. See section 2.5 Abbreviated Syntax at https://www.w3.org/TR/1999/REC-xpath-19991116/.

JavaRosa does not support that shortcut. In practice, form designers typically want to access data based on some filtering value rather than position so it's not a big user-facing limitation but people who know XPath run into this issue when testing.

dcbriccetti:
Closes #422

For comment. Does this simple text substitution approach do the job? Or do we need to get into the parser? Is it correct to assume that any sequence of digits inside square brackets in a nodeset string is a position predicate shortcut?

### DIFF
--- a/src/main/java/org/javarosa/model/xform/XPathReference.java
+++ b/src/main/java/org/javarosa/model/xform/XPathReference.java
@@ -14,9 +14,6 @@
  * the License.
  */
 
-/**
- *
- */
 package org.javarosa.model.xform;
 
 import java.io.DataInputStream;
@@ -37,9 +34,6 @@ import org.javarosa.xpath.parser.XPathSyntaxException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/**
- *
- */
 public class XPathReference implements IDataReference {
     private static final Logger logger = LoggerFactory.getLogger(XPathReference.class);
 
@@ -47,7 +41,6 @@ public class XPathReference implements IDataReference {
     private String nodeset;
 
     public XPathReference () {
-
     }
 
     public XPathReference (String nodeset) {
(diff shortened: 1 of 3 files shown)


lognaturel:
I like this concept but I think it's too early. That is, I think the substitution needs to happen in the predicate evaluation:

> A PredicateExpr is evaluated by evaluating the Expr and converting the result to a boolean. If the result is a number, the result will be converted to true if the number is equal to the context position and will be converted to false otherwise; if the result is not a number, then the result will be converted as if by a call to the boolean function. Thus a location path para[3] is equivalent to para[position()=3].

https://www.w3.org/TR/1999/REC-xpath-19991116/#predicates

For a test form, you could have an integer question and then use the path to that question as the predicate.

ggalmazor:
Hi, @dcbriccetti! I wanted to draw your attention to https://github.com/opendatakit/javarosa/blob/master/src/test/java/org/javarosa/core/test/Scenario.java#L160 in relation to this PR in case that there's anything that could be reused in this PR.

dcbriccetti:
> Hi, @dcbriccetti! I wanted to draw your attention to https://github.com/opendatakit/javarosa/blob/master/src/test/java/org/javarosa/core/test/Scenario.java#L160 in relation to this PR in case that there's anything that could be reused in this PR.

Are you pointing me specifically to this method?
`    public static TreeReference getRef(String xpath) `

ggalmazor:
> Are you pointing me specifically to this method? `public static TreeReference getRef(String xpath)`

Indeed.

lognaturel:
Closing for now to be re-opened as needed!

PR #511, opened 2019-11-15
Fixed countNonEmpty() function

### LINKED ISSUE #509: count-non-empty function doesn't work
#### Software versions 
JavaRosa v2.17.0 and older versions

#### Problem description
[count-non-empty](https://docs.opendatakit.org/form-operators-functions/?highlight=count%20selected#count-non-empty) function doesn't work

Try to open this form: 
[test.xml.txt](https://github.com/opendatakit/javarosa/files/3803410/test.xml.txt)

grzesiek2010:
Closes #509 

I tested implemented changes manually and added an automated test.

The func should return double like other calc functions.

It's not risky, we can just test the attached form to confirm that the problem is fixed. Nothing else should be affected, since it's a small fix.

The form attached to the issue.

No.

### DIFF
--- a/src/test/java/org/javarosa/xform/parse/XFormParserTest.java
+++ b/src/test/java/org/javarosa/xform/parse/XFormParserTest.java
@@ -1,5 +1,27 @@
 package org.javarosa.xform.parse;
 
+import static java.nio.file.Files.copy;
+import static java.nio.file.Files.readAllBytes;
+import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsInstanceOf.instanceOf;
+import static org.javarosa.core.model.Constants.CONTROL_RANGE;
+import static org.javarosa.core.model.Constants.CONTROL_RANK;
+import static org.javarosa.core.test.AnswerDataMatchers.intAnswer;
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.javarosa.xform.parse.FormParserHelper.deserializeAndCleanUpSerializedForm;
+import static org.javarosa.xform.parse.FormParserHelper.getSerializedFormPath;
+import static org.javarosa.xform.parse.FormParserHelper.parse;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
 import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.GroupDef;
 import org.javarosa.core.model.IDataReference;
@@ -17,6 +39,7 @@ import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.core.reference.ReferenceManagerTestUtils;
 import org.javarosa.core.services.transport.payload.ByteArrayPayload;
+import org.javarosa.core.test.Scenario;
 import org.javarosa.core.util.externalizable.DeserializationException;
 import org.javarosa.model.xform.XFormSerializingVisitor;
 import org.javarosa.model.xform.XPathReference;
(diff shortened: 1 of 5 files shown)


src/test/resources/countNonEmptyForm.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ggalmazor: This form feels overcomplicated to test the behavior of `count-non-empty` and `count` expressions. Would you consider simplifying it so that future readers can get a quicker sense of how it relates to the assertions in the test?

src/test/java/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -190,6 +190,12 @@ public void throwsParseExceptionOnBadRangeForm() throws IOException {
         parse(r("bad-range-form.xml"));
     }
 
+    @Test
+    public void formWithCountNonEmptyFunc_ShouldNotThrowException() throws IOException {
```
ggalmazor: I'm not sure what's the goal of this test. Could you elaborate? 

I think I was expecting some kind of assertion here. Maybe assert that after creating some repeats those nodes with the `count-non-empty` and `count` calculate expressions have some expected values.

src/test/resources/countNonEmptyForm.xml
grzesiek2010: Ok, I simplified it.

src/test/java/org/javarosa/xform/parse/XFormParserTest.java
grzesiek2010: We don't need any assertion here because without the fix the form just can't be parsed and an exception is thrown:
`java.lang.RuntimeException: Error evaluating field 'weight_mean2': unrecognized data type in 'calculate' expression: java.lang.Integer`
then fails. With the fix it doesn't fail so I think it's enough.

lognaturel: What I've done in similar cases is asserted that the form title matches expectations. That way it's really a test to make sure that the form parse worked.

ggalmazor: I'm sorry to be nitpicky, but that's not how tests work... A test should help us document some behavior and/or prevent regressions. Neither things are provided by this test.

I think it's a missed opportunity to explain what the `count` and `count-non-empty` functions do in the context of a form. Adding some should be easy with the tooling we have in place.

If I can't seduce you into doing this, I'd rather remove the test.

lognaturel: @ggalmazor I think it’s essential that we have a test that parses a form with the function call. It’s the lack of such a test that led to the broken behavior. This test as written would crash on master. I agree that not having any assertion is sub-optimal but I don’t see what’s so bad about doing an assertion that confirms that the form parse was successful. 

The behavior of the function is tested in XPathEvalTest. I don’t see the benefit of testing it again here. It would require filling in values and I think it would obscure the true goal of the test which is really ensuring that the function is properly parsed. That said, I think we’ve collectively spent too much time on this already and if you feel strongly about it’s fine to add.

ggalmazor: It's just that the current test can fail or pass for things unrelated to the functions we want to verify because it's non-specific.

I also think this really is a low hanging fruit. Check https://github.com/opendatakit/javarosa/commit/5eaee0782746b4fe502a89b5eac7e8b601621b43#diff-448a15bf3c816beba3aac62fe1c2fa24R195 to see what I mean (super straightforward behavior verification).

ggalmazor: In case this is just me being nitpicky about tests, and since @lognaturel is OK with it, I'll approve the PR

lognaturel: Ok, please cherry pick that commit and merge. Your test would crash in the same way if there were a regression which is what's really important.

grzesiek2010:
Ok I added your commit @ggalmazor and merged the pr. @lognaturel if we want this pr to be verified by the QA team we can do that once we update javarosa in collect.

ggalmazor:
Thank you for your patience, @grzesiek2010!

PR #513, opened 2019-12-06
Operators should be case insensitive

### LINKED ISSUE #512: Operators in calculations are case sensitive
#### Software versions 
JavaRosa v2.17.0 and older versions

#### Problem description
In calculations you can use operators like: and, or, div, mod but you can't use AND, OR etc.
it breaks form parsing. I think it's not intuitive and have seen some questions on the forum where it was a problem why a user was't able to use a form.

grzesiek2010:
Closes #512 

I added tests and tested this pr with collect.

Operators should be case insensitive, as I mentioned in the issue I have seen some problem on the forum related to this issue.

It's a safe change. We can just confirm that forms with operators are loaded and that's all.

Any form with [operators](https://docs.opendatakit.org/form-operators-functions/) like the one I used for tests.

No.

### DIFF
--- a/src/main/java/org/javarosa/xpath/parser/Lexer.java
+++ b/src/main/java/org/javarosa/xpath/parser/Lexer.java
@@ -101,16 +101,16 @@ public class Lexer {
             } else if (c == ':' && d == ':') {
                 token = new Token(Token.DBL_COLON);
                 skip = 2;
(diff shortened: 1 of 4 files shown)


lognaturel:
I know this looks like a trivial change but we don't want to encourage form design practices that wouldn't work across tools. It's important to make sure that the other tool maintainers are on board and update the form spec before making these kinds of changes. Closing for now and we can reopen if it's approved at the spec level.

PR #518, opened 2019-12-13
Fixed getSubmissionProfile() method

grzesiek2010:
Closes #https://github.com/opendatakit/collect/issues/3533

I added automated tests and tested the changes manually with collect.

We keep a list of submissionProfiles, I don't know why since we use just one but I decided not to change it. Instead, I fixed `getSubmissionProfile()`
The problem with that method was that we add submissionProfiles using DEFAULT_SUBMISSION_PROFILE (which is "1") if a the submission doesn't have an id and using id otherwise but `getSubmissionProfile()` returned values only using DEFAULT_SUBMISSION_PROFILE key, so if we had a submission with its id the method returned null.

I don't now why we allow to keep a list of submissionProfiles since only one submission is allowed... am I right?

We should test the forms with submissions (with ids and without), that's all nothing else should be affected.

Forms added to automated tests:
`parseFormWithSubmissionElementWithId()` and `parseFormWithSubmissionElement()`

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -1764,12 +1764,9 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
     }
 
     public SubmissionProfile getSubmissionProfile() {
-        // At some point these profiles will be set by the <submit> control in the
-        // form.
-        // In the mean time, though, we can only promise that the default one will
-        // be used.
-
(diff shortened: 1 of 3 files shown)


lognaturel:
Unfortunately that form was leveraging [this XForms feature](https://en.wikibooks.org/wiki/XForms/Submit#The_Submit_Element) which is not supported by Collect. The implementation here is ok. If we do want to make a change, we should consider removing anything to do with an `id` attribute on a `submission` since it's not part of the [ODK XForms spec](https://opendatakit.github.io/xforms-spec/#submission).

PR #520, opened 2019-12-17
Remove redundant dependency cycle while parsing forms

ggalmazor:
This PR removes a duplicated(redundant) cycles check that was done when finalizing the triggerable DAG during form parsing.

Added automated regression tests to ensure no cycles are allowed.

We could always leave the other cycles check and remove the one at `finalizeTriggerables` but it feels like this should be coupled to the finalization method to avoid accidentally missing to call the cycles check method.

The only visible changes should be that now the exception class and error message has changed.

Nope.

Nope.


ggalmazor:
I'm creating this PR as a draft because we will sequence changes related to the DAG in a specific way, and this one comes after other stuff gets merged. This will probably require some dregee of rebasing, but it feels important to be able to explore this change in isolation to advance faster by addressing things in parallel.

ggalmazor:
Close in favor of #526

PR #521, opened 2019-12-17
Polymorphic triggerable

ggalmazor:
Closes #

### DIFF
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -16,21 +16,31 @@
 
 package org.javarosa.core.model;
 
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.javarosa.core.model.condition.Condition;
 import org.javarosa.core.model.condition.EvaluationContext;
+import org.javarosa.core.model.condition.Recalculate;
 import org.javarosa.core.model.condition.Triggerable;
 import org.javarosa.core.model.instance.AbstractTreeElement;
 import org.javarosa.core.model.instance.FormInstance;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
+import org.javarosa.core.util.externalizable.DeserializationException;
+import org.javarosa.core.util.externalizable.ExtUtil;
+import org.javarosa.core.util.externalizable.ExtWrapList;
+import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.debug.EvaluationResult;
 import org.javarosa.debug.Event;
 import org.javarosa.debug.EventNotifier;
@@ -598,34 +608,6 @@ public class TriggerableDag {
         accessor.getEventNotifier().publishEvent(new Event(lead + ": " + (ref != null ? ref.toShortString() + ": " : "") + quickTriggerables.size() + " triggerables were fired."));
     }
 
-    /**
-     * API for retrieving the list of conditions, for use when
-     * serializing the form definition (e.g., into .cache file).
-     */
-    public List<Triggerable> getConditions() {
-        List<Triggerable> conditions = new ArrayList<>();
-        for (QuickTriggerable qt : unorderedTriggerables) {
-            if (qt.isCondition()) {
-                conditions.add(qt.getTriggerable());
-            }
-        }
-        return conditions;
-    }
-
-    /**
-     * API for retrieving thelist of recalculates, for use when
-     * serializing the form definition (e.g., into .cache file).
-     */
-    public final List<Triggerable> getRecalculates() {
-        List<Triggerable> recalculates = new ArrayList<>();
-        for (QuickTriggerable qt : unorderedTriggerables) {
-            if (qt.isRecalculate()) {
-                recalculates.add(qt.getTriggerable());
-            }
-        }
-        return recalculates;
-    }
-
     public void reportDependencyCycles() {
         Set<TreeReference> vertices = new HashSet<>();
         List<TreeReference[]> edges = new ArrayList<>();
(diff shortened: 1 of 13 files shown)


src/main/java/org/javarosa/core/model/condition/Condition.java
```diff
@@ -58,12 +58,8 @@ public Condition() {
 
     }
 
-    public Condition(IConditionExpr expr, int trueAction, int falseAction, TreeReference contextRef) {
-        this(expr, trueAction, falseAction, contextRef, new ArrayList<>(0));
-    }
-
-    public Condition(IConditionExpr expr, int trueAction, int falseAction, TreeReference contextRef, ArrayList<TreeReference> targets) {
-        super(expr, contextRef, targets);
+    protected Condition(IConditionExpr expr, TreeReference contextRef, TreeReference originalContextRef, List<TreeReference> targets, Set<QuickTriggerable> immediateCascades, int trueAction, int falseAction) {
```
lognaturel: Not clear on implications of adding `originalContextRef and `immediateCascades` here. Seems `immediateCascades` may only have been needed for fast 2014?

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -641,11 +641,11 @@ public final IConditionExpr getConditionExpressionForTrueAction(FormInstance mai
      * API for retrieving the list of conditions, for use when
      * serializing the form definition (e.g., into .cache file).
      */
-    public List<Condition> getConditions() {
-        List<Condition> conditions = new ArrayList<>();
+    public List<Triggerable> getConditions() {
```
lognaturel: This doesn't seem right. In this case, we really do want only conditions and I think the type should indicate that.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -655,11 +655,11 @@ public final IConditionExpr getConditionExpressionForTrueAction(FormInstance mai
      * API for retrieving thelist of recalculates, for use when
      * serializing the form definition (e.g., into .cache file).
      */
-    public final ArrayList<Recalculate> getRecalculates() {
-        ArrayList<Recalculate> recalculates = new ArrayList<>();
+    public final List<Triggerable> getRecalculates() {
```
lognaturel: Same comment as before. This type should be explicit.

src/main/java/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
```diff
@@ -124,25 +123,25 @@ DataBinding createBinding(IXFormParserFunctions parserFunctions, FormDef formDef
...
 
         switch (type) {
             case "relevant":
-                prettyType  = "display";
-                trueAction  = Condition.ACTION_SHOW;
+                prettyType = "display";
+                trueAction = Condition.ACTION_SHOW;
                 falseAction = Condition.ACTION_HIDE;
                 break;
             case "required":
```
lognaturel: When looking at cycle-checking behavior, we've considered relevants and calculates but actually requireds and readonlys could also lead to cycles, right? I guess there's no particular need to test for those since they all behave the same.

src/main/java/org/javarosa/core/model/condition/Recalculate.java
```diff
@@ -101,16 +110,6 @@ public boolean equals(Object o) {
         }
     }
 
-    //droos 1/29/10: we need to come up with a consistent rule for whether the resulting data is determined
```
lognaturel: Can we keep this in `wrapData` for now? I don't think those rules are documented anywhere else.

src/main/java/org/javarosa/core/model/condition/Condition.java
```diff
@@ -103,65 +94,26 @@ public boolean isCascadingToChildren() {
...
 
     @Override
-    @SuppressWarnings("unchecked")
     public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOException, DeserializationException {
-        expr = (IConditionExpr) ExtUtil.read(in, new ExtWrapTagged(), pf);
-        contextRef = (TreeReference) ExtUtil.read(in, TreeReference.class, pf);
-        originalContextRef = (TreeReference) ExtUtil.read(in, TreeReference.class, pf);
-        List<TreeReference> tlist = (List<TreeReference>) ExtUtil.read(in, new ExtWrapList(TreeReference.class), pf);
-        targets = new ArrayList<>(tlist);
+        readExternal(this, in, pf);
```
lognaturel: Is there a particular reason to have the superclass have a variant that takes an object as the first param over calling `super.readExternal` and having the same signature?

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -343,7 +340,7 @@ public void finalizeTriggerables(FormInstance mainInstance, EvaluationContext ev
             }
 
             // save for aggressive 2014 behavior
```
lognaturel: Maybe rework this comment while here since there's no more aggressive 2014.

lognaturel:
What would you think about moving 6b4d2e7 and beyond to another branch? The commits before seem like basically a no-brainer to merge so we could get those merged and out of the way.

ggalmazor:
OK, I dropped the commits after https://github.com/opendatakit/javarosa/commit/6b4d2e748539fc0818c4e16dd7715b082a343704 in this PR

src/main/java/org/javarosa/core/model/condition/Condition.java
ggalmazor: This should be safe. I re-routed Triggerable creation to a couple of factories inside it. this only makes explicit all the parts we need to build one.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: I think it's fine. This method is part of the external serialization code, which is kind of an internal detail that we have tight control over. We could even inline that method and we wouldn't have to worry about it.

Anyway, after studying the code, I see there are some confusing concepts related to `Condition` that we might want to come back to and review. The best example of confusing code is when we ask for a condition when evaluating the relevance of an element and we implicitly assume that we're asking for the _relevance_ condition when the DAG could contain any number of different conditions too. Check `FormDef.isRepeatRelevant` and `TriggerableDag.getTriggerableForRepeatGroup`.

At this moment, here's my rationale:
- Our domain talks about "Relevant", "Required", "Constraint", "Read Only", and "Calculate", not about "Condition".

  Previously, these domain concepts were gathered by variable and member names, which is fine for now, I think. I don't think we're losing domain detail by using `Triggerable` here.

  Looking forward, I think we could want to create classes for all the domain concepts for type checking only. Maybe for object creation too, since we already have established that the target reference set for a relevance condition, when applied to a group, should include all descendant refs (kinda different behavior we could want to move into the type).

- Regarding the high-level (abstract) behavior of `Condition`, and `Recalculate`, I don't see any difference: they are triggered, they are eval'ed, and side-effects of their evaluation are applied to some targets.

- We control the creation of triggerables in the internals of the DAG, which is less risky than having this as part of the public API, where having static type checking would be super useful.

src/main/java/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
ggalmazor: I'll throw extra test cases for those too. I prefer to be comprehensive. Otherwise, the lack of tests might be misguiding or confusing for future readers.

src/main/java/org/javarosa/core/model/condition/Recalculate.java
ggalmazor: Sure!

src/main/java/org/javarosa/core/model/condition/Condition.java
ggalmazor: It's just to justify the `readExternal` override in `Recalculate`, but I guess you're right. I want the subtypes to be completely symmetrical respecting to their supertype.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: So, in the end, I refactored the serialization code to make those method private and now I'm confortable returning `Condition` and `Recalculate`. Let me know what you think ;)

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: I think it's actually easy to follow these values in `immediateCascades` around, so I went ahead and removed the comment.

ggalmazor: In the next PR (review of collection types), I add a doc block with the details about what side effect each method has. There, I explain how the immediateCascades set gets while building the DAG.

src/main/java/org/javarosa/core/model/TriggerableDag.java
lognaturel: Looks great to me, thanks!

lognaturel:
Would be great to have a little context in the PR description for posterity to summarize the scope and reasoning of the changes. I'll merge now since it will be easier to keep building on top of and that info can be added later.

src/main/java/org/javarosa/core/model/condition/Condition.java
lognaturel: I've done much more objectionable things to try to avoid inheritance so I get it. I think in this case keeping it as it was is the right call. 👍

PR #523, opened 2019-12-18
Improve collection types

### LINKED ISSUE #458: Don't include self-references in dependency graph
All expressions that include references to form nodes are included in a dependency graph. This makes it possible to only recompute expressions when their dependencies are updated which is an important performance measure for forms with a lot of dependent logic. See [the W3C XForms documentation](https://www.w3.org/TR/2006/REC-xforms-20060314/sliceD.html#rpm-processing-recalc-mddg), `IDag` and `Triggerable`.

Currently, when a calculation in a particular node refers to that same node, that relationship is captured by the dependency graph. This blocks usage of self-referencing expressions because the dependency graph then has a cycle in it.

The W3C XForms spec says the following:

> The depList for a vertex v is assigned to be the vertices other than v whose computational expressions reference v (described below). Vertex v is excluded from its own depList to allow self-references to occur without causing a circular reference exception.

See #440 and #441 for examples of useful self-referencing expressions.

I'm not exactly sure whether all expressions (`calculate`, `relevant`, etc) on a particular node are already recomputed when that node changes so that will be something to explore.

Related to #440, I believe that when expressions are in repeats or depend on values in repeats, the repeat path without any predicates is used in the dependency graph. So a change to any repeat instance will trigger recomputation. This is why computations in repeats are slow.

ggalmazor:
Closes #458

This PR is built on top of #526 to bring a review of the collection types involved in the DAG. The first commit is `Migrate Triggerable.targets to Set` 5d20c1a97c0d887ddf8d97ecc3598d43d26731bb

The main goal would be to use the most appropriate collection type for the task in hand e.g. using sets when we don't want to have duplicates, and linked list supported collections when the order of elements is important.

Important topics:
- Makes part of the DAG building algorithm immutable and side-effect free to make it easier to understand.

  It's important that we understand what are the memory use implications and decide whether we want to undo some of these side-effects as a tradeoff for consuming less memory.
  
  We could also study ways to keep everything easy to understand (immutable & side-effect free) while programming in a way that the garbage collector helps keep memory use in line.

- Refactors the DAG building process to take advantage of specific collection types

  The key point here is that we're able to keep the insertion order and avoid having to sort the DAG, which would bring performance improvements.

Added automated tests

I think this question doesn't apply. Maybe reviewers can make specific questions.

This PR should have no behavior changes, although due to the changes to the DAG building code, some change propagation events might appear in a different order. This shouldn't be a problem because triggerable chain is always guaranteed to happen in the right order according to whether they have dependencies or not.

No.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/QuickTriggerable.java
+++ b/src/main/java/org/javarosa/core/model/QuickTriggerable.java
@@ -58,10 +58,8 @@ public final class QuickTriggerable {
         return this.triggerable.equals(triggerable);
     }
 
-    public Triggerable changeContextRefToIntersectWithTriggerable(Triggerable other) {
-        // TODO It's fishy to mutate the Triggerable here. We might prefer to return a copy of the original Triggerable
-        triggerable.changeContextRefToIntersectWithTriggerable(other);
-        return triggerable;
+    public void intersectContextWith(Triggerable other) {
+        triggerable.intersectContextWith(other);
     }
 
     public void setImmediateCascades(Set<QuickTriggerable> deps) {
--- a/src/main/java/org/javarosa/core/model/QuickTriggerable.java
+++ b/src/main/java/org/javarosa/core/model/QuickTriggerable.java
@@ -50,7 +50,7 @@ public final class QuickTriggerable {
         return triggerable.apply(mainInstance, ec, qualified);
     }
 
-    public List<TreeReference> getTargets() {
+    public Set<TreeReference> getTargets() {
         return triggerable.getTargets();
     }
 
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -258,26 +258,39 @@ public class TriggerableDag {
         return null;
     }
 
-    public final Triggerable addTriggerable(Triggerable t) {
-        QuickTriggerable qt = findTriggerable(t);
-        if (qt != null) {
-            return qt.changeContextRefToIntersectWithTriggerable(t);
-        } else {
-            qt = QuickTriggerable.of(t);
-            allTriggerables.add(qt);
-
-            Set<TreeReference> triggers = t.getTriggers();
-            for (TreeReference trigger : triggers) {
-                Set<QuickTriggerable> triggered = triggerablesPerTrigger.get(trigger);
-                if (triggered == null) {
-                    triggered = new HashSet<>();
-                    triggerablesPerTrigger.put(trigger.clone(), triggered);
-                }
-                triggered.add(qt);
-            }
-
-            return t;
+    /**
+     * Adds the provided triggerable to the DAG.
+     * <p>
+     * This method has side-effects:
+     * <ul>
+     *     <li>If a similar triggerable has been already added, its context gets intersected with the provided triggerable to cover both using only one entry</li>
+     *     <li>This method builds the index of triggerables per trigger ref</li>
+     * </ul>
+     */
+    Triggerable addTriggerable(Triggerable triggerable) {
+        QuickTriggerable existingQuickTriggerable = findTriggerable(triggerable);
+        if (existingQuickTriggerable != null) {
+            // We found a quick triggerable wrapping a triggerable matching the provided one.
+            // Since Triggerable.equals() doesn't have into account contexts or targets, the
+            // safest thing to do would be to create a new quick triggerable for the incoming
+            // triggerable, but we could cover both triggerables if we compute the intersection
+            // between their respective context and reuse the one we've already stored.
+            existingQuickTriggerable.intersectContextWith(triggerable);
+            return existingQuickTriggerable.getTriggerable();
         }
+
+        QuickTriggerable newQuickTriggerable = QuickTriggerable.of(triggerable);
+        allTriggerables.add(newQuickTriggerable);
+
+        // Build the triggerable per trigger index
+        Set<TreeReference> triggers = triggerable.getTriggers();
+        for (TreeReference trigger : triggers) {
+            if (!triggerablesPerTrigger.containsKey(trigger))
+                triggerablesPerTrigger.put(trigger, new HashSet<>());
+            triggerablesPerTrigger.get(trigger).add(newQuickTriggerable);
+        }
+
+        return triggerable;
     }
 
     /**
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -327,10 +327,7 @@ public class TriggerableDag {
      *                               triggers can't be laid out appropriately
      */
     public void finalizeTriggerables(FormInstance mainInstance, EvaluationContext evalContext) throws IllegalStateException {
-        Set<QuickTriggerable> newTriggerablesDAG = buildDag(mainInstance, evalContext);
-
-        triggerablesDAG.clear();
-        triggerablesDAG = newTriggerablesDAG;
+        triggerablesDAG = buildDag(mainInstance, evalContext);
 
         //
         // build the condition index for repeatable nodes
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -64,7 +65,7 @@ public class TriggerableDag {
      * the list, where tA comes before tB, evaluating tA cannot depend on any
      * result from evaluating tB.
      */
-    protected final List<QuickTriggerable> triggerablesDAG = new ArrayList<>();
+    protected final Set<QuickTriggerable> triggerablesDAG = new LinkedHashSet<>();
 
     /**
      * NOT VALID UNTIL finalizeTriggerables() is called!!
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -373,7 +373,10 @@ public class TriggerableDag {
     private List<QuickTriggerable[]> getDagEdges() {
         List<QuickTriggerable[]> edges = new ArrayList<>();
         for (QuickTriggerable source : allTriggerables) {
-            // Compute the set of direct children triggerables of this source vertex
+            // Compute the set of edge targets from the source vertex in this
+            // loop using the triggerable's target tree reference set.
+            // We will create an edge for all the source's target references
+            // that, in turn, trigger another triggerable.
             Set<QuickTriggerable> targets = new HashSet<>();
             for (TreeReference targetRef : source.getTargets())
                 targets.addAll(triggerablesPerTrigger.containsKey(targetRef)
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -59,13 +59,12 @@ public class TriggerableDag {
 
     protected final EventNotifierAccessor accessor;
 
+    protected final Set<QuickTriggerable> allTriggerables = new HashSet<>();
     protected Set<QuickTriggerable> triggerablesDAG = emptySet();
 
     protected Map<TreeReference, QuickTriggerable> repeatConditionsPerTargets = new HashMap<>();
-
     protected final Map<TreeReference, Set<QuickTriggerable>> triggerablesPerTrigger = new HashMap<>();
 
-    protected final Set<QuickTriggerable> allTriggerables = new HashSet<>();
 
     protected TriggerableDag(EventNotifierAccessor accessor) {
         this.accessor = accessor;
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -327,6 +327,27 @@ public class TriggerableDag {
      *                               triggers can't be laid out appropriately
      */
     public void finalizeTriggerables(FormInstance mainInstance, EvaluationContext evalContext) throws IllegalStateException {
+        Set<QuickTriggerable> newTriggerablesDAG = buildDag(mainInstance, evalContext);
+
+        triggerablesDAG.clear();
+        triggerablesDAG = newTriggerablesDAG;
+
+        //
+        // build the condition index for repeatable nodes
+        //
+        conditionRepeatTargetIndex.clear();
+        for (QuickTriggerable qt : triggerablesDAG) {
+            if (qt.isCondition()) {
+                for (TreeReference target : qt.getTargets()) {
+                    if (mainInstance.getTemplate(target) != null) {
+                        conditionRepeatTargetIndex.put(target, qt);
+                    }
+                }
+            }
+        }
+    }
+
+    public Set<QuickTriggerable> buildDag(FormInstance mainInstance, EvaluationContext evalContext) {
         Set<QuickTriggerable> newTriggerablesDAG = new LinkedHashSet<>();
         // TODO Study how this algorithm ensures that we follow the ancestor >>> descendant direction and if the sorting step is strictly required
 
@@ -381,23 +402,7 @@ public class TriggerableDag {
                     edges.remove(i);
             }
         }
-
-        triggerablesDAG.clear();
-        triggerablesDAG = newTriggerablesDAG;
-
-        //
-        // build the condition index for repeatable nodes
-        //
-        conditionRepeatTargetIndex.clear();
-        for (QuickTriggerable qt : triggerablesDAG) {
-            if (qt.isCondition()) {
-                for (TreeReference target : qt.getTargets()) {
-                    if (mainInstance.getTemplate(target) != null) {
-                        conditionRepeatTargetIndex.put(target, qt);
-                    }
-                }
-            }
-        }
+        return newTriggerablesDAG;
     }
 
     public void throwCyclesInDagException(Collection<QuickTriggerable> vertices) {
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -365,45 +365,37 @@ public class TriggerableDag {
         return edges;
     }
 
-    public static Set<QuickTriggerable> buildDag(Set<QuickTriggerable> unorderedTriggerables, List<QuickTriggerable[]> edges) {
+    private static Set<QuickTriggerable> buildDag(Set<QuickTriggerable> vertices, List<QuickTriggerable[]> edges) {
         Set<QuickTriggerable> dag = new LinkedHashSet<>();
         // TODO Study how this algorithm ensures that we follow the ancestor >>> descendant direction and if the sorting step is strictly required
 
-        // DAGify the triggerables based on dependencies and sort them so that
-        // triggerables come only after the triggerables they depend on
-        List<QuickTriggerable> vertices = new ArrayList<>(unorderedTriggerables);
+        Set<QuickTriggerable> remainingVertices = new HashSet<>(vertices);
+        Set<QuickTriggerable[]> remainingEdges = new HashSet<>(edges);
+        while (remainingVertices.size() > 0) {
 
-        List<QuickTriggerable> orderedRoots = new ArrayList<>();
-        Set<QuickTriggerable> roots = new HashSet<>(vertices.size());
-        while (vertices.size() > 0) {
-            // determine root nodes
-            roots.clear();
-            roots.addAll(vertices);
-            for (QuickTriggerable[] edge : edges) {
+            Set<QuickTriggerable> roots = new HashSet<>(remainingVertices);
+            for (QuickTriggerable[] edge : remainingEdges)
                 roots.remove(edge[1]);
-            }
 
-            // if no root nodes while graph still has nodes, graph has cycles
             if (roots.size() == 0)
                 throwCyclesInDagException(vertices);
 
-            // order the root nodes - so the order is fixed
-            orderedRoots.clear();
-            orderedRoots.addAll(roots);
+            List<QuickTriggerable> orderedRoots = new ArrayList<>(roots);
             // TODO Study if insertion order & using LinkedSets would suffice to ensure proper triggerable chaining
             Collections.sort(orderedRoots, QuickTriggerableComparator.INSTANCE);
+            dag.addAll(orderedRoots);
 
-            // remove root nodes and edges originating from them
-            // add them to the triggerablesDAG.
-            for (QuickTriggerable root : orderedRoots) {
-                dag.add(root);
-                vertices.remove(root);
-            }
-            for (int i = edges.size() - 1; i >= 0; i--) {
-                QuickTriggerable[] edge = edges.get(i);
-                if (roots.contains(edge[0]))
-                    edges.remove(i);
-            }
+            Set<QuickTriggerable> newRemainingVertices = new HashSet<>();
+            for (QuickTriggerable vertex : vertices)
+                if (!dag.contains(vertex))
+                    newRemainingVertices.add(vertex);
+            remainingVertices = newRemainingVertices;
+
+            Set<QuickTriggerable[]> newRemainingEdges = new HashSet<>();
+            for (QuickTriggerable[] edge : remainingEdges)
+                if (!roots.contains(edge[0]))
+                    newRemainingEdges.add(edge);
+            remainingEdges = newRemainingEdges;
         }
         return dag;
     }
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -84,7 +84,7 @@ public class TriggerableDag {
     /**
      * List of all the triggerables in the form. Unordered.
      */
-    protected final List<QuickTriggerable> unorderedTriggerables = new ArrayList<>();
+    protected final Set<QuickTriggerable> unorderedTriggerables = new HashSet<>();
 
     protected TriggerableDag(EventNotifierAccessor accessor) {
         this.accessor = accessor;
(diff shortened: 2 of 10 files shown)


ggalmazor:
Christmas is here and I know it's going to be hard to push this one forward, especially since it requires some degree of agreement about the correctness and completeness of the DAG test cases.

Since I'd like to continue working on the codebase on top of this PR, It would be great to at least have a sanity check to the code changes themselves, focusing on how memory usage has changed compared to the current implementation that reuses objects by passing them as input args to methods and mutating them (side-effects) inside them.

If we determine that now JR would use more memory than before, it would be great to find a way to maintain the immutable & side-effect free implementation on this PR, and, only if this is not possible, change it back to a side-effect based, object reusing style.

@dcbriccetti, I was hoping that you could help us with this ;)

ggalmazor:
I've rebased the branch of this PR on top of 526 to make it more concise. I've also taken the chance to divide some commits into smaller chunks to make them easier to review and give more context.

ggalmazor:
I've also checked that no object cloning is done during the DAG build process, which would support our current understanding that the immutable, side-effect free code in this PR shouldn't increase noticeabily memory usage.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -365,45 +365,37 @@ public void finalizeTriggerables(FormInstance mainInstance, EvaluationContext ev
...
-            for (int i = edges.size() - 1; i >= 0; i--) {
-                QuickTriggerable[] edge = edges.get(i);
-                if (roots.contains(edge[0]))
-                    edges.remove(i);
-            }
+            dag.addAll(orderedRoots);
+
+            Set<QuickTriggerable> newRemainingVertices = new HashSet<>();
+            for (QuickTriggerable vertex : vertices)
+                if (!dag.contains(vertex))
```
lognaturel: This feels hard to follow to me. How about `remainingVertices.removeAll(orderedRoots)`? I know the goal was to remove mutation but in this case the mutating approach seems easier to understand. `remainingVertices.removeAll(orderedRoots)` feels nicely symmetric with `dag.addAll(orderedRoots)` so it feels like the roots are moving from one collection to the other.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -365,45 +365,37 @@ public void finalizeTriggerables(FormInstance mainInstance, EvaluationContext ev
...
+                    newRemainingVertices.add(vertex);
+            remainingVertices = newRemainingVertices;
+
+            Set<QuickTriggerable[]> newRemainingEdges = new HashSet<>();
```
lognaturel: Here I don't think there's a particularly clean way to do the mutation so this seems fine to me.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -381,8 +379,6 @@ public void finalizeTriggerables(FormInstance mainInstance, EvaluationContext ev
                 throwCyclesInDagException(vertices);
 
             List<QuickTriggerable> orderedRoots = new ArrayList<>(roots);
-            // TODO Study if insertion order & using LinkedSets would suffice to ensure proper triggerable chaining
-            Collections.sort(orderedRoots, QuickTriggerableComparator.INSTANCE);
```
lognaturel: I don't think you need `orderedRoots` at all, then, and can just `addAll(roots)`.

src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
```diff
@@ -392,25 +392,23 @@ private void applyInstanceProperties(FormInstance instance) {
             EvaluationContext ec = new EvaluationContext(instance);
             final List<TreeReference> nodes = ec.expandReference(ref, true);
 
-            if (nodes.size() > 0) {
```
lognaturel: It's not immediately clear why this is redundant. Is it just that if there was no node then there couldn't be a bind?

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -59,32 +59,12 @@
 
     protected final EventNotifierAccessor accessor;
 
-    /**
-     * NOT VALID UNTIL finalizeTriggerables() is called!!
-     * <p>
-     * Topologically ordered list, meaning that for any tA and tB in
```
lognaturel: Although I do like the new naming and find it helpful, I still find these docs (or a more concise version) helpful.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -392,41 +392,16 @@ private static void throwCyclesInDagException(Collection<QuickTriggerable> trigg
...
-            // so we'll be more inclusive than needed and see if any
-            // of our triggers are keyed on the predicate-less path
-            // of this ref
-            Set<QuickTriggerable> dependantTriggerables = triggerIndex.get(target.hasPredicates() ? target.removePredicates() : target);
-            if (dependantTriggerables != null)
-                allDependantTriggerables.addAll(dependantTriggerables);
+    private static Set<QuickTriggerable> getDependantTriggerables(QuickTriggerable triggerable, Map<TreeReference, Set<QuickTriggerable>> triggerIndex) {
+        Set<QuickTriggerable> result = new HashSet<>();
+        for (TreeReference targetRef : triggerable.getTargets()) {
+            Set<QuickTriggerable> children = triggerIndex.getOrDefault(targetRef, emptySet());
```
lognaturel: I need to come back to 86e2745 and think through it.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -495,42 +495,41 @@ public void throwCyclesInDagException(Collection<QuickTriggerable> vertices) {
...
-                TreeElement child = repeatTemplate.getChildAt(i);
-                descendantRefs.add(child.getRef().genericize());
-                descendantRefs.addAll(getChildrenRefsOfElement(mainInstance, child));
-            }
-        } else {
-            List<TreeReference> refSet = evalContext.expandReference(original);
-            for (TreeReference ref : refSet) {
-                descendantRefs.addAll(getChildrenRefsOfElement(mainInstance, evalContext.resolveReference(ref)));
-            }
+    private Set<TreeReference> getChildrenOfReference(FormInstance mainInstance, EvaluationContext ec, TreeReference ref) {
```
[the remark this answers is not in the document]
ggalmazor: I'll revisit this and see how we can help readers navigate this.

src/main/java/org/javarosa/core/model/QuickTriggerable.java
```diff
@@ -58,10 +58,8 @@ public boolean contains(Triggerable triggerable) {
         return this.triggerable.equals(triggerable);
     }
 
-    public Triggerable changeContextRefToIntersectWithTriggerable(Triggerable other) {
-        // TODO It's fishy to mutate the Triggerable here. We might prefer to return a copy of the original Triggerable
```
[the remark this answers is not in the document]
ggalmazor: I think this is just me talking to myself.

I studied the possibility of making the operation immutable, but I found it to be a can of worms, so flaging the method as `void` felt the best compromise to surface the side-effect.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -327,10 +327,7 @@ public final Triggerable addTriggerable(Triggerable t) {
      *                               triggers can't be laid out appropriately
      */
     public void finalizeTriggerables(FormInstance mainInstance, EvaluationContext evalContext) throws IllegalStateException {
-        Set<QuickTriggerable> newTriggerablesDAG = buildDag(mainInstance, evalContext);
-
-        triggerablesDAG.clear();
-        triggerablesDAG = newTriggerablesDAG;
+        triggerablesDAG = buildDag(mainInstance, evalContext);
```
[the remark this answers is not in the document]
ggalmazor: I was trying to make the least number of changes per commit. I'm sorry it misled you a bit here :P

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: > it feels like the roots are moving from one collection to the other.

This is a very nice way of thinking about what's happening here. It didn't occur to me that it could be explained that way, and I think it would make the algorithm easier to understand which makes it worth the mutation of that collection.

src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
ggalmazor: Yes.

I'll revisit the commit in case I can leave more clues about this.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: This is related to the issue that the lifecycle of the DAG is still not explicit on its API. 

We currently have to call certain methods in a certain sequence in order to have a functional DAG object. That's why this member is not `final` and we needed the comment in the first place.

I'd rather have a builder object of sorts to manage the DAG's creation, implement its lifecycle in an explicit way and have a completely immutable DAG object in return.

For now, I thought that the use of `DAG` in the member's name could tell everything that has to be known about the contents and ordering of the list, and the member being not `final` and assigning an `emptySet()` could tell about it not actually being usable before something else happens.

ggalmazor: > I'd rather have a builder object of sorts to manage the DAG's creation, implement its lifecycle in an explicit way and have a completely immutable DAG object in return.

This would also help us separate the contexts where a `Triggerable` should be used as opposed to a `QuickTriggerable`. The former should be used during DAG building, the latter only when the DAG is stable.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: I've added a commit to deal with this.

Also, the commits belonging to 86e2745 could be split into another PR if you want.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -64,7 +65,7 @@
      * the list, where tA comes before tB, evaluating tA cannot depend on any
      * result from evaluating tB.
      */
-    protected final List<QuickTriggerable> triggerablesDAG = new ArrayList<>();
+    protected final Set<QuickTriggerable> triggerablesDAG = new LinkedHashSet<>();
```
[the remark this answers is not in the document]
ggalmazor: Let's hear him out :)

My rationale is that the use of `DAG` in the wording is enough and that it's really an implementation detail that doesn't really help us avoid bugs i.e. using a LinkedList doesn't prevent us from inserting elements in the wrong order.

src/main/java/org/javarosa/core/model/TriggerableDag.java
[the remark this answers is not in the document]
ggalmazor: Well, it turns out that after 86e2745e756bc341b0e06f06667368754e4c69f2, these methods are unused, so I've removed them.

ggalmazor:
I reworded the commit messages to fit the subject under 72 chars and I didn't realize that the links to commits you were hoping to stufy better would change as a result. Sorry, @lognaturel!

73e63649d759e210056fc7735271bb4b5946ddc0 is still there, but the one about avoiding to hit the main instance is in 2bd11f6a78d4b39e4e384a7b6f26d9f408fd6535 now

src/main/java/org/javarosa/core/model/TriggerableDag.java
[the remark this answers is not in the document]
lognaturel: Those are fair arguments.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -344,26 +343,31 @@ public void finalizeTriggerables(FormInstance mainInstance) throws IllegalStateE
...
+            // Compute the set of roots (nodes that don't show up
+            // as edge targets) with the remaining vertices
             Set<QuickTriggerable> roots = new HashSet<>(remainingVertices);
             for (QuickTriggerable[] edge : remainingEdges)
                 roots.remove(edge[1]);
 
             if (roots.size() == 0)
                 throwCyclesInDagException(vertices);
 
+            // "Move" the roots detected during this iteration
```
lognaturel: With your new comments, this whole method feels pretty comfortable to understand to me, now. 👍

src/main/java/org/javarosa/core/model/TriggerableDag.java
lognaturel: All that sounds good but I wasn't so much talking about the lifecycle warning as the brief descriptions. The naming helps but I like to read some prose about what fields represent to get a feel for what a class does. That certainly is a personal preference.

ggalmazor: Understood. I'll add some comments then :)

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -270,144 +291,148 @@ public void copyItemsetAnswer(FormInstance mainInstance, EvaluationContext evalC
...
+            for (TreeReference targetRef : source.getTargets())
+                targets.addAll(triggerIndex.containsKey(targetRef)
+                    ? triggerIndex.get(targetRef)
+                    : emptySet());
+
+            // Account for cycles by self-reference
+            if (targets.contains(source))
+                throwCyclesInDagException(targets);
+
+            if (source.canCascade())
```
lognaturel: This check isn't necessary because targets will always be empty if a source can't cascade.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -270,144 +291,148 @@ public void copyItemsetAnswer(FormInstance mainInstance, EvaluationContext evalC
...
+     * This method has side-effects:
+     * <ul>
+     *     <li>Throws IllegalStateException when cycles are detected, either due
+     *     to a self-reference or more complex cycle chains</li>
+     *     <li>Builds a cache of immediate cascades of each vertex, meaning that
+     *     we will remember the dependant vertices without having to traverse
+     *     the DAG again</li>
+     * </ul>
+     */
+    private static List<QuickTriggerable[]> getDagEdges(Set<QuickTriggerable> vertices, Map<TreeReference, Set<QuickTriggerable>> triggerIndex) {
```
lognaturel: Would be helpful for `triggerIndex` to have the same name as the field that's passed in for it (`triggerablesPerTrigger`).

ggalmazor:
We need a test that explores deeply nested computations and repeat groups before 2bd11f6a78d4b39e4e384a7b6f26d9f408fd6535

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -270,144 +291,148 @@ public void copyItemsetAnswer(FormInstance mainInstance, EvaluationContext evalC
...
+        List<QuickTriggerable[]> edges = new ArrayList<>();
+        for (QuickTriggerable source : vertices) {
+            // Compute the set of direct children triggerables of this source vertex
```
lognaturel: Should this say "direct dependent" instead of "direct children"? In the case of relevance on a group, the references of all the nodes inside the group (and the groups inside the group, etc) will be included, right? All of these descendants are direct dependents of the containing group because their relevance is directly affected by the parent group's relevance.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -270,144 +291,148 @@ public void copyItemsetAnswer(FormInstance mainInstance, EvaluationContext evalC
...
-            if (deps.contains(qt))
-                throwCyclesInDagException(deps);
-
-            if (qt.canCascade())
-                for (QuickTriggerable qu : deps) {
-                    QuickTriggerable[] edge = {qt, qu};
-                    partialOrdering.add(edge);
-                }
+    public void finalizeTriggerables(FormInstance mainInstance) throws IllegalStateException {
+        List<QuickTriggerable[]> edges = getDagEdges(allTriggerables, triggerablesPerTrigger);
```
lognaturel: Currently this is only ever used when all the contents are put in a set so there's no reason for this to be a list. 

It occurs to me that there are no tests that fail if the dag is unsorted. That is, I can go from `LinkedHashSet` to `HashSet` in the initialization of `triggerablesDAG` and all tests still pass. The `buildDag` algorithm is now easy to understand (I wish I could quickly generate an animation because it's very cool how the ordering is guaranteed) so that gives confidence but test coverage would would help me sleep even better at night. Does this seem worthwhile, @ggalmazor?

I'm not entirely sure what such a test would look like. I'd want to do something like force a hash code for an element at the end of a triggerable chain to sort at the beginning of `triggerablesDAG`.

src/main/java/org/javarosa/core/model/TriggerableDag.java
lognaturel: Thanks, I find the comments very helpful. 😊

src/main/java/org/javarosa/core/model/TriggerableDag.java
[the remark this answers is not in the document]
lognaturel: See https://github.com/opendatakit/javarosa/pull/523#discussion_r365445589. I think the absolute ideal would be to have tests that fail if the nodes are in the wrong order. I discussed some ideas for adding an end-to-end test with a form definition in the comment but I'm wondering whether introducing a unit test on `buildDag` could be sufficient.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: > Currently this is only ever used when all the contents are put in a set so there's no reason for this to be a list. 

Let's make it a `Set`. It's an inexpensive change and it makes the context richer.

> It occurs to me that there are no tests that fail if the dag is unsorted [...] I'm not entirely sure what such a test would look like

I have two ideas:
- We could write a couple of simple low-level tests around `buildDag` that explicitly show how the output DAG is sorted while describing its behavior.
- We could write an end-to-end test with 3 fields `a`, `b`, `c`. 
  - `b` depends on `a` as in `a * 2`.
  - `c` depends on `a` and `b`, as in `a * b`.
  - We give initial values to all fields in the form.
  - We change `a` to trigger all computations and verify that `c` was computed using the updated value of `b`, not the initial value in the form.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: Ugh, language is hard. I think the comment is misleading by using the word "children". We should  stick with DAG terminology. I'll review this and try to improve it with your suggestions.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: I've pushed a commit including the end-to-end test, which I think would be enough. Let me know if you'd want the low-level test too.

ggalmazor:
I think I've gone through your latest comments in the PR and now I'm trying to focus exclusively on the changes regarding the mutually exclusive recursive methods to get all descendant refs from 73e63649d759e210056fc7735271bb4b5946ddc0 and 2bd11f6a78d4b39e4e384a7b6f26d9f408fd6535

In order to start from a safe place, I'm reverting the changes from 73e63649d759e210056fc7735271bb4b5946ddc0. 

Then, after evaluating the move of those methods from `TriggerablesDag` to `FormInstanceParser`, I've assessed that it's safe because we established that they will only get called when the triggerable can cascade which, in turn, it's only true for groups with a relevance condition. This should make it safe to replace the descendant target computation from `TriggerablesDag.getDependantTriggerables` to `FormInstanceParser.applyInstanceProperties`. If you're not confident about this, we should make it a target for new tests.

With this setup, my plan is to dissect `FormInstance.getTemplatePath` and write tests that exercise both forks of the `if` block at `FormInstanceParser.getDescendantRefs`.

ggalmazor:
I've been playing with this test to understand how the check in FormInstanceParse.getDescendantRefs works:

```java
@Test
public void name2() {
    TreeElement root = new TreeElement("data");
    TreeElement group = new TreeElement("group");
    TreeElement field = new TreeElement("field");
    root.addChild(group);
    group.addChild(field);
    FormInstance formInstance = new FormInstance(root);

    TreeReference ref = TreeReference.rootRef()
        .extendRef("data", 0)
        .extendRef("group", 0);
    TreeElement templatePath = formInstance.getTemplatePath(ref);

    Set<TreeReference> descendantRefs = FormInstanceParser.getDescendantRefs(formInstance, new EvaluationContext(formInstance), ref);
    System.out.println(descendantRefs);
}
```

The breakthrough comes when declaring the `group` element to have a multiplicity to `1`, which forces the `getDescendantRefs` method to follow the `else` path in the check, which uses the `EvaluationContext` to expand the tree reference and get the corresponding elements.

Unfortunately, there's no way the parser would reproduce this scenario when parsing a form's XML document. We can't describe in XML a group at multiplicity `1` without declaring another group at multiplicity `0`. This is key because the check at `getDescendantRefs` will revert to the zeroth group when resolving the `mainInstance.getTemplatePath(original);` if no group with `jr:template=""` is declared.

These findings confirm our theory that the `else` branch is not really needed because, even though we can artificially exercise it from our tests by manipulating the contents of the instance, no form can actually do it.

src/main/java/org/javarosa/core/model/TriggerableDag.java
lognaturel: > and it doesn't guarantee the order of its elements

Well, it guarantees that they'll be in the insertion order. But no matter, we agree on the change to `Set` which is great.

> I've pushed a commit including the end-to-end test, which I think would be enough.

Ah yes, perfect!

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -270,144 +291,160 @@ public void copyItemsetAnswer(FormInstance mainInstance, EvaluationContext evalC
...
 
-            return t;
+        QuickTriggerable newQuickTriggerable = QuickTriggerable.of(triggerable);
+        allTriggerables.add(newQuickTriggerable);
+
+        // Build the triggerable per trigger index
+        Set<TreeReference> triggers = triggerable.getTriggers();
+        for (TreeReference trigger : triggers) {
+            if (!triggerablesPerTrigger.containsKey(trigger))
+                triggerablesPerTrigger.put(trigger, new HashSet<>());
```
lognaturel: The original code used to use a clone of the reference as an index. This uses a shared reference. I can't immediately tell whether that's a safe change or not and want to make sure it's deliberate.

ggalmazor: I changed that because we're not mutating those keys and each time we call `Triggerable.getTriggers()` we get new objects anyway.

If you think this could be problematic, we can clone them again.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: > Well, it guarantees that they'll be in the insertion order. But no matter, we agree on the change to Set which is great.

Ouch, you're right! I guess I still have the parallel streams mindset, where you can't trust the order will continue being the same as the original insertion order after some stream manipulations.

src/main/java/org/javarosa/core/model/TriggerableDag.java
lognaturel: But do we know for sure that those references can't be mutated by some other process in JavaRosa after the DAG has been built? I find the fact that those references can be mutated a bit scary in general. The index is accessed when triggerables are triggered so if the keys are mutated that could mean the wrong fields being triggered.

src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
```diff
@@ -378,68 +386,105 @@ private void verifyItemsetSrcDstCompatibility (FormInstance instance) {
...
             }
         }
 
         applyControlProperties(instance);
     }
 
-    private static void attachBindGeneral (DataBinding bind) {
-        TreeReference ref = FormInstance.unpackReference(bind.getReference());
+    private Set<TreeReference> getDescendantRefs(FormInstance mainInstance, EvaluationContext evalContext, TreeReference original) {
+        // original has already been added to the 'toAdd' list.
```
lognaturel: This comment doesn't make sense anymore because there isn't a `toAdd` list. Does the `original` now get added because it is in its own descendant list?

src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
```diff
@@ -378,68 +386,105 @@ private void verifyItemsetSrcDstCompatibility (FormInstance instance) {
...
-            bind.calculate.addTarget(ref);
+        Set<TreeReference> descendantRefs = new HashSet<>();
+        TreeElement repeatTemplate = mainInstance.getTemplatePath(original);
+        if (repeatTemplate != null) {
+            for (int i = 0; i < repeatTemplate.getNumChildren(); ++i) {
+                TreeElement child = repeatTemplate.getChildAt(i);
+                descendantRefs.add(child.getRef().genericize());
+                descendantRefs.addAll(getDescendantRefs(mainInstance, child));
+            }
+        } else {
```
lognaturel: Do you want to add a todo about this branch being unreachable or do you think that since you're looking at that now you'll be doing something about it soon enough?

lognaturel:
I have a couple of new questions above. Other than that, I have completed my review. It looks like a lot of things can now be private in `TriggerableDag`. Worth doing here?

src/main/java/org/javarosa/core/model/TriggerableDag.java
lognaturel: @ggalmazor traced through this with me and showed me that indeed `getTriggers` takes us into `XPathConditional.getTriggers` which builds new `TreeReference`s from the actual expressions. So this is safe.

src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
lognaturel: `bind.relevancyCondition.addTarget(ref)` right before the call adds the original reference.

lognaturel:
Thanks, @ggalmazor! Have you had a chance to compare benchmarks from before and after this change? It seems like there could be some positive performance implications that come with some of the simplifications and O(1) collection changes.

After more interactive review, I feel more confident that it's unlikely that this will have significant negative memory implications. So @dcbriccetti, I think we're ok without additional review on this one. Thanks!

PR #525, opened 2019-12-20
Issue 522 triggerables triggered twice on new repeat

### LINKED ISSUE #522: Triggerables triggered twice on repeat creation
A call on `triggerTriggerables` was introduced in `createNewRepeat` at https://github.com/opendatakit/javarosa/commit/9b88318435487f8fca91d60e67e0457b2a181d97#diff-96c7e556ff24192dbc4851b9b4410bbfR508. However, `dagImpl.createRepeatGroup` already makes the same call. The call on `triggerTriggerables` should be removed.

ggalmazor:
Closes #522

This PR is built on top of #523 and starts in commit commit 3f97d3bef1b3b95bf1615a1cf2e55eff6be8b60d

Run the tests.

This is a straightforward change. No alternatives have been considered.

There should be no behavior change

Nope.

No.


dcbriccetti:
I’m having a look

PR #528, opened 2020-01-07
Improve DAG test suite

ggalmazor:
This PR tries to define and document as much of the DAG's behavior as possible through new test methods and forms.

Run automated tests

Doesn't apply: this PR doesn't have behavior changes. Only test code has been changed.

Doesn't apply

No.

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
+++ b/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
@@ -3,6 +3,7 @@ package org.javarosa.core.model;
 import static java.lang.String.format;
 import static java.lang.String.join;
 import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toList;
 import static java.util.stream.IntStream.range;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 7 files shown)


src/test/java/org/javarosa/core/model/TriggerableDagTest.java
```diff
@@ -528,6 +528,40 @@ public void parsing_forms_with_cycles_should_fail() throws IOException {
...
+        exceptionRule.expect(RuntimeException.class);
+        exceptionRule.expectMessage("Dependency cycles amongst the xpath expressions in relevant/calculate");
+
+        Scenario.init("Some form", html(head(
+            title("Some form"),
+            model(
+                mainInstance(t("data id=\"some-form\"",
+                    t("count", "1")
+                )),
+                bind("/data/count").type("int").relevant(". + 1")
```
lognaturel: Relevants should be boolean expressions. It's interesting that `. + 1` is allowed and I'm not really sure what the behavior is. For clarity I think it'd be good to add e.g. `> 0`.

src/test/java/org/javarosa/core/model/TriggerableDagTest.java
```diff
@@ -531,16 +531,237 @@ public void parsing_forms_with_cycles_by_self_reference_in_relevance_should_fail
...
+            head(
+                title("Some form"),
+                model(
+                    mainInstance(t("data id=\"some-form\"",
+                        t("group",
+                            t("a", "1"),
+                            t("b", "1")
+                        )
+                    )),
+                    bind("/data/group/a").type("int").calculate("/data/group[position() = position(current())]/b + 1"),
```
lognaturel: `/data/group[position() = position(current())]/b` could also be expressed as `../b`, right? I suppose those should have exactly the same behavior so it doesn't matter which is used?

lognaturel:
Closing because it's tightly related to #526 and will be easier to just make changes there.

PR #530, opened 2020-01-08
Improve getting dependant triggerables

ggalmazor:
This PR is built on top of #523 and brings a performance improvement during the DAG build process.
 It starts on 88874a6e053ffb7ecbafa1cb728e1932c216e74c

Run automated tests

Before this PR, we used to hit the main instance to get descendant references of each triggerable, which is potentially a slow process.

The main reason behind this PR is that we should be able to build a descendant tree of triggerables with the information that they already have (their target refs) and the triggerable index the DAG stores.

We have established that doing so breaks some trigger chains when a group has relevance conditions. This has been solved by adding all the group's descedant refs as targets of the relevance condition.

There should be no behavior change.

Performance-wise, the changes should be safe, if not faster than the original code, although no benchmarking has been done.

No.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -687,7 +687,7 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
         // DAGify the triggerables based on dependencies and sort them so that
         // triggerables come only after the triggerables they depend on
         //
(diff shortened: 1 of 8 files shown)


ggalmazor:
@lognaturel, I'd like to add that there's no noticeable performance pitfall like the one we detected before in the tests that used the eIMCI form.

This PR runs smoothly, as it should :)

ggalmazor:
Closing in favor of #523

PR #535, opened 2020-01-30
Issue 534 fix date utils test

### LINKED ISSUE #534: Wrong parsing of time during certain hours of the day
#### Problem description
`DateUtilTest` fails during certain times of the day depending on the system's time zone and the amount of hours set in the `offsetTwoHours` variable (currently `+2`). In my computer (Spain) it fails from 22:00 to 3:00, and we've seen the build fail in CI at 23:40

- You can try changing your system's time to different times until you get a test failure
- You can tweak the value of `offsetTwoHours` until you get a test failure

#### Expected behavior
Tests should be deterministic. It's not clear whether the failure comes from an issue on the tests or the implementation, though.

#### Other information 
N/A

ggalmazor:
Closes #534

This PR has become more than simply fixing the DateUtilsTest class into a review of the whole test suite to ensure we always deal with timezones and locales in a way that's consistent, safe and easier to understand.

- Run automated tests
- Use [external](https://www.epochconverter.com/timezones?q=1580378400000&tz=Pacific%2FHonolulu) tools to double-check that the `java.time` values used to replace long timestamps actually represent the same instant.

- Using `java.time` as much as possible is the safest way to move forward even if the implementation codebase will use Jodatime while we support Android API levels that don't include it.

  This is the standard way to deal with dates in Java starting from Java8 and it's very well documented with lots of examples.

  Also, using a `java.time` serves as a safe and solid baseline that provides an extra sanity check over our code and tests.

- Using human readable date and time definitions in our tests makes them easier to understand. When something goes wrong, it will be easier to detect problems if we can read the values involved, as opposed to using timestamps.

- Although Using the `@Before` and `@After` hooks to prevent side-effects due to setting default locales and time zones in test methods is super safe, using the methods provided by `SystemHelper` (new in this PR) make more explicit where this is actually required.

This PR doesn't change the implementation codebase, so no behavior change expected whatsoever.

No

No

### DIFF
--- a/src/test/java/org/javarosa/core/model/utils/test/DateUtilsFormatSanityCheckTests.java
+++ b/src/test/java/org/javarosa/core/model/utils/test/DateUtilsFormatSanityCheckTests.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2009 JavaRosa
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package org.javarosa.core.model.utils.test;
+
+import static java.util.TimeZone.getTimeZone;
+import static org.hamcrest.Matchers.is;
+import static org.javarosa.core.model.utils.DateUtils.FORMAT_ISO8601;
+import static org.javarosa.core.model.utils.DateUtils.formatDateTime;
+import static org.javarosa.core.model.utils.DateUtils.parseDateTime;
+import static org.junit.Assert.assertThat;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Date;
+import java.util.TimeZone;
+import java.util.stream.Stream;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Parameterized.class)
+public class DateUtilsFormatSanityCheckTests {
+    @Parameterized.Parameter(value = 0)
+    public long inputTimestamp;
+
+    @Parameterized.Parameters(name = "Input timestamp: {0}")
+    public static Collection<Object[]> data() {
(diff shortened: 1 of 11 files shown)


src/test/java/org/javarosa/core/model/utils/test/DateUtilsTests.java
```diff
@@ -66,25 +69,25 @@ public void testGetXMLStringValueFormat() {
...
 
-            private String language;
+            private Locale locale;
             private String january;
             private String sunday;
         }
 
         LangJanSun langJanSuns[] = new LangJanSun[]{
-            new LangJanSun("en", "Jan", "Sun"),
-            new LangJanSun("es", "ene", "dom"),
```
ggalmazor: This test failed in my machine because my JVM produces `ene.` instead of `ene`

src/test/java/org/javarosa/core/model/utils/test/DateUtilsParseTimeTests.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @After
+    public void tearDown() {
+        TimeZone.setDefault(backupZone);
+        Locale.setDefault(backupLocale);
+    }
+
+    @Test
+    public void testTimeParses() {
+        // The tricky part of DateUtils.parseTime is that we allow for input time
```
lognaturel: Thank you for this insightful comment.

src/test/java/org/javarosa/core/model/utils/test/DateUtilsSCTOTests.java
```diff
@@ -77,27 +60,25 @@ public void testParseTime() {
     // - We're effectively binding all times to the EPOCH date
```
lognaturel: Is this comment still accurate? Aren't all times now bound to whatever the current time of the device is?

src/test/java/org/javarosa/core/model/utils/test/DateUtilsFormatSanityCheckTests.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Collection;
+import java.util.Date;
+import java.util.TimeZone;
+import java.util.stream.Stream;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Parameterized.class)
+public class DateUtilsFormatSanityCheckTests {
```
lognaturel: This could probably be renamed. `DateUtilsFormatTests` would be better, I think.

ggalmazor: This one was tricky to name. "Format" tells half of the story. These tests format a date and then parse it to verify we get the same value back.

PR #536, opened 2020-01-31
Make SelectChoice Serializable

grzesiek2010:
Closes https://github.com/opendatakit/collect/issues/3615

I tested this change with Collect.

We just need `SelectChoice` to be serializable in Collect.

It doesn't require testing.

No.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/SelectChoice.java
+++ b/src/main/java/org/javarosa/core/model/SelectChoice.java
@@ -3,6 +3,8 @@ package org.javarosa.core.model;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.io.Serializable;
+
 import org.javarosa.core.model.data.helper.Selection;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.services.locale.Localizable;
@@ -13,7 +15,7 @@ import org.javarosa.core.util.externalizable.Externalizable;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.xform.parse.XFormParseException;
 
-public class SelectChoice implements Externalizable, Localizable {
+public class SelectChoice implements Externalizable, Localizable, Serializable {
 
     private String labelInnerText;
     private String textID;


grzesiek2010:
I promised to take another look to make sure we need this change or maybe it is possible to handle it somehow in ODK Collect.
I added this change because I wanted to use SelectChoices in DialogFragment ([RankingWidgetDialog](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/fragments/dialogs/RankingWidgetDialog.java#L55) to use it in [RankingListAdapter](https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/adapters/RankingListAdapter.java#L42)) to fix this issue: https://github.com/opendatakit/collect/issues/3486
prior to the fix I passed just strings https://github.com/opendatakit/collect/pull/3487/files#diff-5dcb4ce9c0336f02f73bf9e41c9554baL54
The thing is that in RankingListAdapter I need to calculate every item name https://github.com/opendatakit/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/adapters/RankingListAdapter.java#L59
to do that I need SelectChoices. Prior to the fix I used https://github.com/opendatakit/collect/pull/3487/files#diff-ec912db8da5def803b4dde7a3c59aafbL65
but if I use items form external files it's not possible to do that using form controller (because items from external files are not imported in the same way like items from the same xml file as I remember?).

So this is the easiest way. Maybe I could try to call the list of SelectChoices (from RankingListAdapter) that I already keep in RankingWidget (without passing it to RankingListAdapter) but it would be a bit strange and made the code complicated so I would be in favor of this small change.

lognaturel:
Thanks for looking into this and for your patience, as always.

I understand that this looks like a tiny change but Java serialization is [so problematic](https://www.journaldev.com/2452/serialization-in-java) that it is on the way to being removed (https://news.ycombinator.com/item?id=17165800, https://openjdk.java.net/jeps/154). From Joshua Bloch's Effective Java: "There is no reason to use Java serialization in any new system you write." [Android documentation](https://developer.android.com/reference/java/io/Serializable#recommended-alternatives) on `Serializable` has sections on implementing it judiciously and recommended alternatives.

`SelectChoice` has a `TreeElement` field so `TreeElement` would also need to be serializable.

Additionally, it doesn't feel right to make a change to JavaRosa for an Android-specific reason. I think a better approach would be to separate data and presentation by using something like a viewmodel to keep the necessary data separate from display concerns.

PR #537, opened 2020-02-03
Simplify evaluation of triggerables

ggalmazor:
Run the automated tests

We have established that the `firedAnchors` map tries to cache generic refs to their corresponding set of expanded refs. After studying the code, this will actually provide more work, instead of less work, to the evaluation engine, since we would still be expanding the cached qualified references in `firedAnchors`. This is documented in https://github.com/ggalmazor/javarosa/blob/dag_documentation/dag.md

The changes in this PR try to take advantage of the insights we got while studying the DAG to simplify it by making the same with less code or by making it more explicit about its actual behavior.

Users should see no different behavior.

No.

No.


lognaturel:
This PR can now also remove https://github.com/opendatakit/javarosa/blob/master/dag.md#analysis-of-the-firedanchors-map.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -168,15 +168,8 @@ public TriggerableDag(EventNotifierAccessor accessor) {
 
             try {
 
-                // Now identify all of the fully qualified nodes which this
-                // triggerable updates. (Multiple nodes can be updated by the same trigger)
-                List<TreeReference> qualifiedList = evalContext.expandReference(contextRef);
```
lognaturel: Think through what it would mean for `contextRef` to not be fully qualified.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -109,50 +109,50 @@ public TriggerableDag(EventNotifierAccessor accessor) {
...
+                evaluated.add(triggerable);
             }
         }
-        return fired;
+        return evaluated;
     }
 
     /**
      * Step 3 in DAG cascade. evaluate the individual triggerable expressions
      * against the anchor (the value that changed which triggered recomputation)
```
lognaturel: It might be helpful to modify this comment now that there isn't an `anchor` parameter.

ggalmazor:
>  I'd sleep better at night if we added a couple of end-to-end tests on real forms before merging this, though

Sure! I think we'll be adding a couple of full form answering scripts based both on the WHO VA and Tecno SAS6 forms, but I was planning on doing so after finishing the work I'm doing on form analysis tools and metrics.

In case that that doesn't feel enough, do you have any specific scenario that wouldn't be already represented in the current test suite? Since triggerable evaluation is at the core of JR, I was confident that the current tests would be enough to change this code safely.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: If the reference is not fully qualified, then the `EvaluationContext` that we create later won't be able to resolve relative paths, which would be bad. They need to be absolute too.

Fortunately, we know that all trigger refs arriving to `evaluateTriggerable` are absolute and fully qualified.

In any case, I think I want to study this just a little bit more to be on the safe side. We can always conditionally expand the trigger reference if we detect that it's ambiguous (there's a `TreeReference.isAmbiguous()` method). If the trigger ref is relative, we can also contextualize it with the triggerable too.

Both things are what we were doing before and I removed in this PR. The difference would be that now we would be doing it conditionally and giving enough context to future readers. It would feel bad, though, to not be able to explain precisely under which scenarios that could happen.

src/main/java/org/javarosa/core/model/TriggerableDag.java
ggalmazor: All these comments need a review, honestly. I think that the new DAG documentation superseeds them, though.

lognaturel:
This is on hold because the work @ggalmazor did at https://github.com/opendatakit/javarosa/pull/537#discussion_r374563873 led him to discover some cases broken by this.

ggalmazor:
I was searching for a button to transform this PR into a draft, but I didn't find it. Is it possible to do that anyway?

lognaturel:
No! 😭 The draft feature has been great. I hope going back to draft state is something Github adds soon. https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107

ggalmazor:
So, after giving this another spin on top of the latest changes (#546), it looks like the only simplification that's still relevant is the one about removing the `firedAnchors` cache map.

@lognaturel, do you think we should pursue this? We established that this cache object should have a negative impact on performance under scenarios where many triggerables are evaluated on each pass.

ggalmazor:
> With the current firedAnchors map, I've tried to have a key with more than one value and I can't seem to get one. What would it take for that to happen?

I found an instance of this in the child vaccination smoke test. The triggerable in `/data/household_count` with the calculate expression of `if(/data/building_type = 'single', 1, if(/data/flatcount = 0 or indexed-repeat(/data/household/finalflat, /data/household, /data/flatcount) != 'yes', /data/flatcount + 1, /data/flatcount ))` gets 4 refs for its key in `firedAnchors`

lognaturel:
> gets 4 refs for its key in firedAnchors

Ah yes, I see. It would have to be the same generic reference in the same expression. In this case, `/data/flatcount` is the generic key and the 4 references are actually the exact same contextualized value 4 times, right? Not very useful. 

What about an expression like `/data/foo[0] + /data/foo[1] + /data/foo[2]`? Those would be three different contextualized refs associated with the same generic ref? Still not very useful but I just want to make sure I get what's going on.

ggalmazor:
I remove the `QuickTriggerable.findAffectedTriggerables()` and I think this PR is ready now :)

lognaturel:
I've been hesitant about this one because it makes me nervous that I don't understand the intent behind `firedAnchors`. That makes me think that I may not be able to think of the kind of form that would be affected by removing it.

That said, @ggalmazor's analysis of why it looks like a performance enhancement gone wrong is compelling. I've traced several forms through that code path and it really doesn't make a whole lot of sense. As @ggalmazor has also pointed out, the new smoke tests cover a lot of common form design patterns. I think the benefits of code that actually makes some amount of sense outweighs the risk here.

In exploring various different forms, I came across https://github.com/opendatakit/javarosa/pull/560.

PR #538, opened 2020-02-03
Dag documentation

ggalmazor:
This PR adds documentation about the inner workings of the DAG, as well as deviations from specs and possible improvements that we might want to consider.

### DIFF
--- a/dag.md
+++ b/dag.md
@@ -1,0 +1,134 @@
+# DAG
+
+## DAG lifecycle
+
+### 1 - The form is parsed
+
+`XFormsParser.parseDoc` parses the blank form's XML contents and returns a `FormDef`
+
+- `parseElement(_xmldoc.getRootElement(), _f, topLevelHandlers)` triggers a recursive parsing process
+- Eventually, `StandardBindAttributesProcessor` objects parse the `<bind>` elements. Here condition and computation triggerables are parsed and:
+  - They are added to the DAG by calling to `TriggerableDag.addTriggerable(...)`.
+  
+    This also builds an index of triggerables per trigger reference.
+    
+  - They're assigned as members of the output `DataBinding` objects.
+- `FormInstance fi = instanceParser.parseInstance(...)` triggers calling `FormInstanceParser.applyInstanceProperties(...)`, which iterates the parsed bindings and sets a two-way relationship between the `DataBinding` objects and the `TreeElement` object corresponding to their `nodeset` references.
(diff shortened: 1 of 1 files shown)


dag.md
```diff
@@ -0,0 +1,234 @@
...
+
+We need to remember that we need to follow an alternative strategy to evaluate the triggerables belonging to repeat groups because the DAG uses genericized trigger references to index sets of triggerables. In this case, calling `Set<QuickTriggerable> alreadyEvaluated = triggerTriggerables(mainInstance, evalContext, repeatGroup.getRef(), new HashSet<>(0))` once per sibling (starting from the deleted group's position) deals with that.
+
+We also have to evaluate any triggerable triggered by children references of the repeat group, which is done by `evaluateChildrenTriggerables(mainInstance, evalContext, repeatGroup, false, alreadyEvaluated)`. Even though this is done inside the loop, the `if` check ensures this is only done once, coinciding with the first iteration (corresponding to the sibling that takes the place of the deleted one). Presumably, this is to have a nice ordering of published events in the `EventNotifier`, although there's no apparent functional requirement for this, and the call could be done outside the loop in `TriggerableDag.deleteRepeatGroup(...)`.
+
+#### d - A complex itemset value is copied to the main instance
```
lognaturel: Would be good to add a test form for this.

It's not part of the [spec](https://opendatakit.github.io/xforms-spec/). Forms that use the feature have to be built by hand. In the longer run, should we consider deprecating it and maybe removing that code? It seems that reducing complexity for dynamic selects would be beneficial all around.

This would be useful if, for example, select options were "people" nodes with "phone number", "address", etc child nodes and selecting one "people" node would copy over the entire subtree, right?

ggalmazor: > It's not part of the spec. Forms that use the feature have to be built by hand

Ahhh, I was suspecting that. Then yes, I also think that we should get rid of it. 

I don't think we could actually deprecate it though, because clients don't actually interact with this feature (there's no public API method that we could tag). The feature automatically kicks in when a complex structure is detected in the values of a select field's itemset.

> This would be useful if, for example, select options were "people" nodes with "phone number", "address", etc child nodes and selecting one "people" node would copy over the entire subtree, right?

Yes, that seems to be the use for this, but I couldn't find any working example of this. After studying the whole DAG's behavior, the code that deals with this is kinda straightforward.

Depending on what we decide to do with this feature, I could try to build a form and add some tests but if we decide to ultimately remove it, then I don't think it's worth the effort.

PR #539, opened 2020-02-05
Improve test scenario

ggalmazor:
Run automated tests

This PR improves the tools we use to write tests, describe JR's behavior through them, and provide examples of how clients might use it.

`Scenario` was created at a time when I was starting to study JR and the xforms specs. Now I have a deeper understanding of it and I don't trust the initial implementation so much, especially when dealing with transforming xpath strings to `TreeReference` instances when repeat groups are involved.

This PR tries to align the implementation with my current understanding of how things should work in JavaRosa and XForms.

No behavior change for users.

No.

No.


src/test/java/org/javarosa/core/model/TriggerableDagTest.java
```diff
@@ -890,5 +890,183 @@ public void deleteRepeatGroupWithCalculationsTimingTest() throws Exception {
         logger.info("Deletion of {} repeats took {}", numberOfRepeats, duration.toString());
     }
 
+    @Test
```
ggalmazor: I'm not sure how this commit got here. Sorry!

src/test/java/org/javarosa/core/test/Scenario.java
```diff
@@ -119,6 +118,84 @@ private Scenario(FormDef formDef, FormEntryController formEntryController) {
         this.formEntryController = formEntryController;
     }
 
+    // region Miscelaneous
+
+    public static TreeReference getRef(String xpath) {
```
lognaturel: This method basically just adds support for the position predicate shortcut, right? It's confusing to see this just in tests. Why not actually support https://github.com/opendatakit/javarosa/issues/422?

src/test/java/org/javarosa/core/test/Scenario.java
```diff
@@ -119,6 +118,84 @@ private Scenario(FormDef formDef, FormEntryController formEntryController) {
...
+    }
+
+    public EvaluationContext getEvaluationContext() {
+        return formDef.getEvaluationContext();
+    }
+
+    public Scenario populateDynamicChoices() {
+        FormIndex backupIndex = formEntryController.getModel().getFormIndex();
+        while (!atTheEndOfForm()) {
+            silentNext();
```
lognaturel: This populates dynamic questions of the next question? Why? Doesn't getting the choices implicitly populate them if dynamic? This seems like it would better match what an actual client does.

src/main/java/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -483,14 +483,16 @@ public int hashCode() {
     }
 
     public String toString() {
-        return toString(true);
+        return toString(true, true);
     }
 
     public String toString(boolean includePredicates) {
```
lognaturel: I don't think that still taking in parameters but overriding them is a good idea. Could you mark those two variants as deprecated instead?

lognaturel: I don't think that still taking in parameters but overriding them is a good idea. Could you mark those two variants as deprecated instead?

Did you check to make sure that none of the downstream tools we know about use the toString in anything user-facing or in something that may later be machine-parsed?

Since this is a public API change, I would have expected it to be called out in the PR description.

src/main/java/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -676,4 +676,10 @@ public TreeReference removePredicates() {
         }
         return predicateless;
     }
+
+    public TreeReference removePredicates(int i) {
```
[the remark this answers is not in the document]
ggalmazor: I'm mirroring what's done at `removePredicates`, which also returns a mutated clone :mutant_emoji: 

Do you think we should change the naming here? I wouldn't be able to just rename `removePredicates()` because it's public, though (I'd have to deprecate it)

src/test/java/org/javarosa/core/test/Scenario.java
ggalmazor: Huh, I didn't know #422 was in the pipeline. We should totally support that, although it feels a change too big for this PR.

I just wanted to be sure that the refs we work with in tests are exactly the same thing that would be produced by the runtime. As you know, equality rules in tree refs are tricky and I was losing confidence on the previous `xpath -> TreeReference` factories that we were using in tests before this.

Also, after this PR gets merged I could send in another for #422 based on the `getRef` implementation if you feel like we're going in the right direction.

src/test/java/org/javarosa/core/test/Scenario.java
ggalmazor: Nevermind this method. This was something that I tried to simplify the `SelectChoice` tests but then I decided to go in another direction. I committed this method by mistake :/

Tests can call `Scenario.choicesOf` as always to get select choices, which populates them implicitly.

My original idea was to make it explicit at critical points of a test for future readers of tests, but ultimately I thought it wasn't bringing so much value to the table. The comments that are already in place work OK, and, as you've said, it is closed to what clients would do.

src/main/java/org/javarosa/core/model/instance/TreeReference.java
ggalmazor: I think you caught this one in the process of being changed. Ultimately I removed all the overloads.

> Did you check to make sure that none of the downstream tools we know about use the toString in anything user-facing or in something that may later be machine-parsed?

Ugh, I didn't expect anyone to be coupled to the `toString()` method, and I can see some uses in Collect. It feels wrong to use `toString()` to implement business logic!

The implementation of `TreeReference.toString()` has been bugging me from the beginning. The default implementation should match the internal state of the object (with predicates and zero-based indexes), not the way around. 

In tests, we're writing and reading refs all the time, and it's annoying to have to double-check them in case you forgot to include predicates, or the multiplicity values are not what you expect.

Here's what I'll do:
- I'll revert the change in this PR
- I'll study all the places in Collect, Aggregate, Briefcase, and Validate where we use any overload of `TreeReference.toString()` and open issues to provide new API methods in JR.

ggalmazor:
Thanks for your review, @lognaturel!

> This took me a while to review because it wasn't a PR I was expecting! I think the changes overall look positive and I look forward to seeing answers to my questions.

Don't worry about it, I wasn't blocked by this. It's just that while working on the regressions we detected some days ago, I needed to review the `Scenario` class, and I thought we could expedite things by creating a PR for it, so that when the PR with the regression fixes come, we can focus only on the test and the fix.

src/main/java/org/javarosa/core/model/instance/TreeReference.java
ggalmazor: Just to illustrate how annoying the current `toString()` implementations are, take a look at some test results that I'm working on right now:

```
java.lang.AssertionError: 
Expected: is "/foo"
     but: was </foo>
Expected :is "/foo"
Actual   :</foo>
```
The difference is in a predicate, but no overload of `toString()` shows them, and I can't make my tests show them without implementing custom Hamcrest matchers, or providing a fourth new overload.

src/main/java/org/javarosa/core/model/instance/TreeReference.java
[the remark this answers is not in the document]
lognaturel: Ahh, I didn't see it was an overload of an existing method. And this is only needed for `Scenario.getRef`. Let's keep it for now and try to remove it when addressing #422.

src/test/java/org/javarosa/core/test/Scenario.java
lognaturel: > I could send in another for #422 

Let's plan on that.

lognaturel:
> I wasn't blocked by this

Oh good, I'm glad. 

Reverting the changes to `toString` seems like the right thing to do. I agree that the current `toString` behavior and usage is really frustrating. Off the top of my head, I couldn't recall how bad the usage in Collect is, but now I see why my spidey-sense was tingling.

How about introducing `toXpathString` which does what we'd like the `toString` to do and use that in the `reason` string for assertions for now? We can then work towards breaking the Collect dependance on the `toString` and eventually make the switch.

> It feels wrong to use toString() to implement business logic!

Completely agree. But even if there hadn't been that, the changes to the public `toString` methods would still have been user-facing changes that would have been good to flag in the PR description. I have to admit I missed this in my first review!

ggalmazor:
> Off the top of my head, I couldn't recall how bad the usage in Collect is, but now I see why my spidey-sense was tingling.

I've started tracking down all those usages at https://github.com/opendatakit/collect/issues/3631

ggalmazor:
> How about introducing toXpathString which does what we'd like the toString to do and use that in the reason string for assertions for now? We can then work towards breaking the Collect dependance on the toString and eventually make the switch.

Knowing that we will probably be addressing this, I think I can live with things like they're now for some more time ;)

ggalmazor:
> the changes to the public toString methods would still have been user-facing changes that would have been good to flag in the PR description 

Yes, you're absolutely right. Sorry!

PR #542, opened 2020-02-10
Issue 541 fix regression relevance propagation

### LINKED ISSUE #541: Regression: wrong relevance updates of fields in groups with relevance conditions
#### Software versions 
This problem affects to the unreleased version of JavaRosa currently in the `master` branch.

#### Problem description
73e63649d759e210056fc7735271bb4b5946ddc0, and 86e2745e756bc341b0e06f06667368754e4c69f2 change the way we resolve triggerable targets for relevance conditions defined in groups. Before these commits, we resolved these refs during triggerable evaluation, which felt like something that could be improved by pre-computing them while parsing the form instead.

As a result of those changes, given a group and a field inside that group with relevance conditions in both, their related relevance condition triggerables became equipotent, which could break unexpectedly their order of evaluation. 

The bottom line is that the last relevance condition evaluated is the one that decides whether a field is relevant or not. If the group's relevance condition is met, and it's the last one to be evaluated, a field will be relevant regardless of its particular relevance condition, which can lead to unexpected wrong states.

##### Deviation from specs in relevance propagation

This has also to do with a deviation from the specs that we've been discussing for some time too. The XForm specs don't talk about relevance changes propagating to descendants of a group. Relevance is a UI concern and making a group irrelevant should only mean that the group would be skipped while traversing the form. No changes should happen relevancy-wise in descendant fields.

Nevertheless, JavaRosa is currently propagating the group's relevance to its descendants to achieve a performance optimization when resolving a nodeset and getting the value of nodes. **It turns out that JavaRosa has some hard-coded relevance checks in both places to ignore nodes that are irrelevant when computing a nodeset and to return `null` values on irrelevant fields.** The goal would be to reduce upstream computations by managing fewer nodes and values.

This has to be further studied and discussed before doing anything that would affect this behavior

TBD

#### Expected behavior
Relevance should work exactly the same as in v2.17.0

#### Other information 
TBD

ggalmazor:
Closes #541

This PR is built on top of the `Scenario` improvements in #539. The first commit is `
Implement toString in QuickTriggerable and DateData for easier debugging` at 2d3bb2bd

Created a new automated regression test

This PR solves the regression described in #541. The new test has full details about the regression and the scenario that leads to it.

While studying this we considered fixing the regression by stopping to propagate relevance updates into a group's descendant fields. This would ensure that just one relevance condition is ever evaluated for each node, which also aligns JavaRosa with the specs. Unfortunately, this would also undo the hard-coded optimizations described in the `Deviation from specs in relevance propagation` section in #541.

Another alternative fix involved removing from the relevance condition triggerable's targets those refs that would correspond to descendant nodes with their own relevance conditions. This would also ensure that we just evaluate one relevance condition per node, but this proved to be impossible without changing the form parsing process. `FormInstanceParser.applyInstanceProperties()` descends into nodes starting from the root, which means that we couldn't know whether a group's descendant node has a relevance condition or not because the method wouldn't have iterated through it yet.

N/A

Nope.

No.


dsrees:
# [Codecov](https://codecov.io/gh/opendatakit/javarosa/pull/542?src=pr&el=h1) Report
> Merging [#542](https://codecov.io/gh/opendatakit/javarosa/pull/542?src=pr&el=desc) into [master](https://codecov.io/gh/opendatakit/javarosa/commit/c845cacf6778c934c00d5b66ef7aa8a84ca9cd14&el=desc) will **increase** coverage by `0.10%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/opendatakit/javarosa/pull/542?src=pr&el=tree)

```diff
@@             Coverage Diff              @@
##             master     #542      +/-   ##
============================================
+ Coverage     52.90%   53.01%   +0.10%     
- Complexity     3111     3129      +18     
============================================
  Files           245      245              
  Lines         13340    13386      +46     
  Branches       2566     2574       +8     
============================================
+ Hits           7058     7097      +39     
- Misses         5464     5478      +14     
+ Partials        818      811       -7     
```


| [Impacted Files](https://codecov.io/gh/opendatakit/javarosa/pull/542?src=pr&el=tree) | Coverage Δ | Complexity Δ | |
|---|---|---|---|
| [...a/org/javarosa/xform/parse/FormInstanceParser.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EveGZvcm0vcGFyc2UvRm9ybUluc3RhbmNlUGFyc2VyLmphdmE=) | `66.76% <0.00%> (-4.18%)` | `109.00% <0.00%> (-5.00%)` | |
| [.../org/javarosa/core/reference/ReferenceManager.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EvY29yZS9yZWZlcmVuY2UvUmVmZXJlbmNlTWFuYWdlci5qYXZh) | `60.63% <0.00%> (-1.07%)` | `20.00% <0.00%> (-1.00%)` | |
| [...rg/javarosa/core/model/instance/TreeReference.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EvY29yZS9tb2RlbC9pbnN0YW5jZS9UcmVlUmVmZXJlbmNlLmphdmE=) | `75.53% <0.00%> (-0.69%)` | `107.00% <0.00%> (+2.00%)` | |
| [src/main/java/org/javarosa/core/model/FormDef.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EvY29yZS9tb2RlbC9Gb3JtRGVmLmphdmE=) | `70.08% <0.00%> (+0.42%)` | `156.00% <0.00%> (+1.00%)` | :arrow_down: |
| [...in/java/org/javarosa/xpath/expr/XPathFuncExpr.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EveHBhdGgvZXhwci9YUGF0aEZ1bmNFeHByLmphdmE=) | `70.42% <0.00%> (+0.45%)` | `209.00% <0.00%> (+1.00%)` | :arrow_down: |
| [...varosa/core/model/condition/EvaluationContext.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EvY29yZS9tb2RlbC9jb25kaXRpb24vRXZhbHVhdGlvbkNvbnRleHQuamF2YQ==) | `73.71% <0.00%> (+0.64%)` | `43.00% <0.00%> (+1.00%)` | :arrow_down: |
| [...in/java/org/javarosa/xpath/expr/XPathPathExpr.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EveHBhdGgvZXhwci9YUGF0aFBhdGhFeHByLmphdmE=) | `63.21% <0.00%> (+1.72%)` | `47.00% <0.00%> (+2.00%)` | :arrow_down: |
| [...a/org/javarosa/core/model/condition/Condition.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EvY29yZS9tb2RlbC9jb25kaXRpb24vQ29uZGl0aW9uLmphdmE=) | `35.13% <0.00%> (+2.70%)` | `10.00% <0.00%> (+1.00%)` | :arrow_down: |
| [...n/java/org/javarosa/xpath/expr/XPathArithExpr.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EveHBhdGgvZXhwci9YUGF0aEFyaXRoRXhwci5qYXZh) | `93.75% <0.00%> (+3.12%)` | `18.00% <0.00%> (+1.00%)` | :arrow_down: |
| [.../org/javarosa/xform/parse/ElementChildDeleter.java](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree#diff-c3JjL21haW4vamF2YS9vcmcvamF2YXJvc2EveGZvcm0vcGFyc2UvRWxlbWVudENoaWxkRGVsZXRlci5qYXZh) | `96.77% <0.00%> (+3.22%)` | `14.00% <0.00%> (+1.00%)` | :arrow_down: |
| ... and [5 more](https://codecov.io/gh/opendatakit/javarosa/pull/542/diff?src=pr&el=tree-more) | |

------

[Continue to review full report at Codecov](https://codecov.io/gh/opendatakit/javarosa/pull/542?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/opendatakit/javarosa/pull/542?src=pr&el=footer). Last update [c845cac...a562bb4](https://codecov.io/gh/opendatakit/javarosa/pull/542?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -398,6 +395,82 @@ public void finalizeTriggerables(FormInstance mainInstance) throws IllegalStateE
         return edges;
     }
 
+    private static Set<QuickTriggerable> getDependantTriggerables(FormInstance mainInstance, EvaluationContext ec, QuickTriggerable triggerable, Map<TreeReference, Set<QuickTriggerable>> triggerIndex) {
```
lognaturel: The corresponding methods should be removed from `FormInstanceParser` as they're no longer used, right?

ggalmazor:
> I think it would also be possible to sort the relevancy condition targets so that the most local one is always evaluated last. But this seems simpler. 

It turns out that it's all the way around: the outermost relevance condition can make a local field irrelevant regardless of that field's relevance status. `TriggerableDagTest.relevance_propagation_topmost_relevance_wins()` explores this in #543

> Exploring that alternative a bit leads me to notice that QuickTriggerableComparator is still in the code base but unused. Should we have a lint check for unused code?

Maybe we should. I'll search what checkstyle rule we need to add for this.

> When we last spoke, you mentioned a performance regression that I don't recall the details of. Does this also address it?

Can't recall exactly what it was. Maybe the evaluation of more than one relevance condition per field?

In the regression test I've added in this PR using the WHO VA form, I've verified that the relevance for the field at `/data/consented/illhistory/illdur/Id10120_0` is computed just once after setting the subject's death date. Before this PR, relevance would be computed more times, which leads to making the field relevant, thus the regression.

Maybe I was referring to this?

lognaturel:
> the outermost relevance condition can make a local field irrelevant regardless of that field's relevance status.

Ah yes, this is hard to reason about. What I was saying is something like to determine whether a field is relevant, start from its outermost container. If that container is irrelevant, stop. If that container is relevant, go one level deeper and repeat. If the field is reached (all of its containers are relevant), its relevance condition determines whether or not it is relevant. I think we're saying the same thing.

> Before this PR, relevance would be computed more times

I believe this is what I was remembering.

PR #545, opened 2020-02-24
Add name of "Select" option to error message

### LINKED ISSUE #131: "select question has no choices" error message is incomplete
Originally reported by: **Yaw Anokwa (Bitbucket: [yanokwa](https://bitbucket.org/yanokwa), GitHub: [yanokwa](https://github.com/yanokwa))**

----------------------------------------

**What steps will reproduce the problem?
**

Get a form that has a select1 with no choices. Let the select1 field be "my_select". Try to start a new form with that form

**What is the expected output? **

The error message should include the name of the field with the error, for example: "Select question 'my_select' has no choices"

** What do you see instead? **

The error message is incomplete: "Select question has no choices", it doesn't tell you which question.

----------------------------------------
- Bitbucket: https://bitbucket.org/javarosa/javarosa/issue/55

dave91087:
Also add new test, and associated resource

Closes #131

Ran automated test suites, including new test specifically for this case

Simple change

New error message in the case of a Select without any options. Anything relying on specific error text (hopefully nowhere!) may now error.

There is a form in test/resources that produces the expected exception

No

### DIFF
--- a/src/test/resources/org/javarosa/xform/parse/internal_empty_select.xml
+++ b/src/test/resources/org/javarosa/xform/parse/internal_empty_select.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <h:head>
+        <h:title>internal empty select</h:title>
+        <model>
+            <instance>
+                <internal_empty_select id="internal_empty_select">
+                    <first/>
+                    <meta>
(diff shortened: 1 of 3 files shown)


src/test/resources/org/javarosa/xform/parse/internal_empty_select.xml
```diff
@@ -0,0 +1,27 @@
...
+            <instance id="first">
+                <root>
+                </root>
+            </instance>
+            <bind nodeset="/internal_empty_select/first" type="select1"/>
+            <bind calculate="concat('uuid:', uuid())" nodeset="/internal_empty_select/meta/instanceID" readonly="true()" type="string"/>
+        </model>
+    </h:head>
+    <h:body>
+        <select1 ref="/internal_select_10/first">
```
lognaturel: The root node name here should be `internal_empty_select`!

dave91087: Should now be resolved

dave91087:
Thanks for the feedback @lognaturel .

I've hopefully resolved the issues 👍

lognaturel:
Thank you, @dave91087!

PR #543, opened 2020-02-10
Improve test suite

ggalmazor:
This PR only improves the test suite without changing the implementation codebase.

Noteworthy tests for future discussion:
- There are new tests to describe how relevance and constraint conditions work and the strange interaction between relevance and computations
- There are tests to describe the behavior discussed in #119, #135, and #540 
- There are a couple of smoke tests.
  - The test with the WHO VA form is more directed and has more assertions. This is a form without repeat groups but a complex network of relevance conditions
  - The test with the child vaccination form is more dynamic and has assertions only at a couple of places. This is a form with simple logic but nested repeats featuring infinite (outer) repeat groups.

#### What has been done to verify that this works as intended?
Run the automated tests

#### Why is this the best possible solution? Were any other approaches considered?
Doesn't apply

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
This PR shouldn't affect users in any way.

#### Do we need any specific form for testing your changes? If so, please attach one.
Nope.

#### Does this change require updates to documentation? If so, please file an issue [here]( https://github.com/opendatakit/docs/issues/new) and include the link below.
Nope.

PR #546, opened 2020-02-27
Issue 540 expansion of refs in nested repeats

### LINKED ISSUE #540: Relative references aren't correctly expanded in nested repeats

ggalmazor:
Closes #540

This PR is built on top of #543 and the first commit is b4959970

Added automated tests
Run the whole test suite

An alternative solution that was speculated on would be to change function evaluation to overwrite any multiplicity value at the last step in references that have to be used with functions that operate on nodesets rather than single nodes.

Users should be able to use relative parent references inside nested repeats and obtain the correct nodesets in return for example to correctly count nested repeat groups correctly from within those groups.

No.

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
+++ b/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
@@ -912,11 +912,20 @@ public class TriggerableDagTest {
         scenario.createNewRepeat("/data/outer[1]/inner");
         scenario.createNewRepeat("/data/outer[1]/inner");
 
-        assertThat(scenario.answerOf("/data/outer[0]/inner[0]/count"), is(intAnswer(1))); // Should be 3
-        assertThat(scenario.answerOf("/data/outer[0]/inner[1]/count"), is(intAnswer(1))); // Should be 3
-        assertThat(scenario.answerOf("/data/outer[0]/inner[2]/count"), is(intAnswer(1))); // Should be 3
-        assertThat(scenario.answerOf("/data/outer[1]/inner[0]/count"), is(intAnswer(1))); // Should be 2
-        assertThat(scenario.answerOf("/data/outer[1]/inner[1]/count"), is(intAnswer(1))); // Should be 2
+        // The following assertions have incrementally greater count values because
+        // we currently evaluate triggerables bound to the new repeat group only. This
+        // means that the actual count value only takes into account the existing repeat
+        // groups when creating the new instance.
+        //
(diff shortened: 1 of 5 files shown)


src/test/java/org/javarosa/core/model/TriggerableDagTest.java
```diff
@@ -913,10 +913,10 @@ public void relative_paths_in_nested_repeats() throws IOException {
         scenario.createNewRepeat("/data/outer[1]/inner");
 
         assertThat(scenario.answerOf("/data/outer[0]/inner[0]/count"), is(intAnswer(1))); // Should be 3
-        assertThat(scenario.answerOf("/data/outer[0]/inner[1]/count"), is(intAnswer(1))); // Should be 3
-        assertThat(scenario.answerOf("/data/outer[0]/inner[2]/count"), is(intAnswer(1))); // Should be 3
+        assertThat(scenario.answerOf("/data/outer[0]/inner[1]/count"), is(intAnswer(2))); // Should be 3
```
lognaturel: When you say "as we would expect" in the commit comment, do you mean "as the implementation suggests" rather than "as is correct?" Count calculations *should* update every time a repeat is added, right? Isn't it the case that on full-form recomputation all the inner counts will be 3 and the outer counts will be 2? And that if you delete the first inner repeat instance the count fields for the remaining instances will be updated to 2 but that if you delete the last inner repeat instance no update will happen?

The great news is that I think the contextualization fix is just right! I'll kick it around my brain a bit more to see if there are other test cases that would be worth adding. But this test still seems super confusing. The way that phase 1 of repeat instance creation is described [in the docs](https://github.com/opendatakit/javarosa/blob/master/dag.md) suggests to me that the counts should all be updated and I'm not really clear on why they're not. I think that should be added either to the docs or to the tests or perhaps both.

ggalmazor: >  Count calculations should update every time a repeat is added, right?

Yes, precisely. I forgot to mention that I was planning on separating the contextualization fix from the wrong behavior we're seeing in those counts into another issue, so that this fix could be merged.

Before this PR they would compute a 1. Now they compute the count at the time of the node's creation, which is still wrong, but better somehow.

Would you rather go all the way and try to fix the wrong count numbers in this PR?

lognaturel: Sounds fine to focus on the contextualization fix here. I haven't had a chance to come back and do an honest review yet! Let's just make sure that there's documentation of the issue of counts in repeat instances before a newly added one not getting re-evaluated counts.

PR #552, opened 2020-03-09
Allow expressions in digest arguments

### LINKED ISSUE #547: Add test with form for digest function and ensure expressions for the first parameter are parsed
Reported by @tiritea. This makes Validate fail so `digest` isn't accessible through XLSForm.

#### Problem description
The first argument to the `digest` function is cast to a `String` when it should be evaluated as an expression.

Write a form with some expression as the first argument to `digest`. 

#### Expected behavior
The expression should be evaluated and the digest of its result should be computed. Instead, it crashes.

lognaturel:
Closes #547

Wrote and ran automated tests.

The fix itself involved following the documentation in `XPathFuncExpr` so there weren't many decisions to make. I did consider also relaxing the checking for valid arguments and encoding and returning empty strings instead of throwing exceptions. Ultimately, I decided that was an unrelated change and not all that important since generally those will be fixed values. These exceptions is why the test has `if(/data/my-algorithm != '' and /data/my-encoding != '' ...`.

I also had to make a decision about where the test would go. @ggalmazor recently introduced a `regression` package so that was the alternative I considered. I put it in `XPathEvalTest` because I consider it critical functionality of `digest` but I can also see an argument for saying that it's implied and this is just a regression check.

Regression risk is just to the `digest` function. It's narrow enough that I think the risk is very low. It should just fix the bug.

[dynamic-digest.xml.zip](https://github.com/opendatakit/javarosa/files/4308999/dynamic-digest.xml.zip)

No.


lognaturel:
> Did you consider adding it to the DigestAlgorithmTest class?

No. 😄 Is that your preferred place? I'm not against it. I guess my first reaction is that "algorithm" makes it seem like it's a low level unit test so feels like a different level than what I'm adding here but I could parameterize the form def.

lognaturel:
@ggalmazor I made the suggested change and force pushed not to leave behind confusing history. Do you want to take another look and then merge if happy?

Freeedim:
Hi,

Thanks for the quick fix!

I am the user who reached out to @tiritea and his colleagues for this issue in the first place. But I am not familiar with JavaRosa coding at all. Since I don't really understand everything in the fix description in the original post and since I don't think users can test the fix already, I would just want to make sure:
Will we have the following behaviour if ${foo} is an empty string?
digest(${foo}, "md5", "hex") -> 'd41d8cd98f00b204e9800998ecf8427e'

ggalmazor:
Hi, @Freeedim!

I just verified that the example use you've provided matches what JavaRosa will produce. I was able to find an issue with the test, so this example was super useful. Thanks!!!

PR #553, opened 2020-03-10
Add section about benchmarks

lognaturel:
Every time I want to run benchmarks, I have to look up how to do it. I was running them with `./gradlew jmh` but @ggalmazor has let me know that he got inconsistent results running them that way and instead suggests first building a jar.


lognaturel:
I think we should probably recommend not using the host machine at all while benchmarks are running. What do you think, @ggalmazor?

PR #560, opened 2020-03-11
Add test to highlight recomputation bug

lognaturel:
This is a test-only change meant to document some new-to-me behavior. I was always under the impression that the final validation pass performed by e.g. Collect triggered all computations. That is not the case. I haven't done a deep dive yet but I believe that because the validation runs through user-visible fields, calculate-only fields will not be recomputed.

@ggalmazor does this match your expectations? Do you agree that this is a bug?

I also expected that if `/data/repeat` were a repeat, that I could do `count(/data/repeat)` to get the count of instances of that repeat. That expression always returns 1 as though the reference were to a specific repeat instance. I think that's a bug.

Given a node `/data/repeat/foo` in that same repeat, `count(..)` returns 1. I think that might be the right thing though I'm wondering whether `count` should always use a generic reference for the last step in the reference as we've previously discussed.

### DIFF
--- a/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
+++ b/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
@@ -929,6 +929,65 @@ public class TriggerableDagTest {
     }
 
     /**
+     * This test documents some bugs:
+     *  * calling the count function with an absolute reference to a repeat should return the count of repeat instances
+     *  but instead it always returns 1
+     *  * a count inside of a repeat instance is not recomputed as new instances are added
+     *  * calculations that don't involve user input are not recomputed on form validation
+     */
+    @Test
+    public void validation_only_triggers_calculations_involving_user_visible_fields() throws IOException {
+        Scenario scenario = Scenario.init("Some form", html(
+            head(
+                title("Some form"),
+                model(
+                    mainInstance(t("data id=\"some-form\"",
+                        t("user-input", "1"),
+                        t("repeat jr:template=\"\"",
+                            t("count"),
+                            t("multiplied-count"),
+                            t("some-field")
+                        )
+                    )),
+                    bind("/data/user-input").type("int"),
+                    bind("/data/repeat/count").type("int").calculate("count(../../repeat)"), // both .. and /data/repeat always give a count of 1
+                    bind("/data/repeat/multiplied-count").type("int").calculate("/data/user-input * count(../../repeat)")
+                )
+            ),
+            body(
+                input("/data/user-input"),
+                repeat("/data/repeat",
+                    input("/data/repeat/some-field"))
+            )));
+
+        scenario.createNewRepeat("/data/repeat");
+        scenario.createNewRepeat("/data/repeat");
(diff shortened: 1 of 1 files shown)


ggalmazor:
I think that my understanding of how JR behaves is consistent with the values we're getting after validation.

- The validation algorithm iterates over all relevant fields and re-commits their existing values.
- In the form you're using, this means the inputs for `/data/user-input` and `/data/repeat/some-field`
  - When the value for `/data/user-input` is re-committed, the triggerable in `/data/repeat/multiplied-count` is evaluated.

    This expression includes a `count` call, which produces a 3 because the main instance has 3 repeats at `/data/repeat`.

    This produces the all `3` values we're seeing.

    So far, so good.

  - `/data/repeat/some-field` is irrelevant because there are no expressions depending on this field.

- The values in all `/data/repeat/count` are not recomputed because the validation process doesn't create or delete repeat instances, which is the trigger for their calculate expressions.

> I also expected that if /data/repeat were a repeat, that I could do count(/data/repeat) to get the count of instances of that repeat. That expression always returns 1 as though the reference were to a specific repeat instance. I think that's a bug.

This is interesting! I didn't think about this when working on #546...

I think that we're seeing that `/data/repeat` is contextualized into `/data/repeat[x]` (x being the corresponding multiplicity of the new repeat instance that's being created), which will always produce a count of 1.

It looks like my fix from #546 only works with relative refs inside counts :/

> I'm wondering whether count should always use a generic reference for the last step in the reference as we've previously discussed.

I agree. We can always intervene during function evaluation to change how the nodesets are computed just for the `count` function, although it feels like it should be a last resort solution. I think  we should explore options to make the changes from #546 a bit smarter to support absolute refs too.

ggalmazor:
I've been exploring options for supporting absolute references in `count` arguments to avoid the "always 1" issue. Placing this in `XPathFuncExpr` seems to do the trick:

```java
for (int i = 0; i < args.length; i++) {
    if (name.equals("count")) {
        // We need to inspect the argument to ensure we will be working with a nodeset instead of a single node
        TreeReference argReference = ((XPathPathExpr) args[0]).getReference();

        // Ensure the last step is unbound
        if (argReference.size() > 0)
            argReference.setMultiplicity(argReference.size() - 1, TreeReference.INDEX_UNBOUND);

        // Resolve an absolute reference, in case the argument is relative or starts with a parent ref
        TreeReference absoluteArgRef = argReference.anchor(evalContext.getContextRef());

        // Ensure the evaluation context we use to evaluate the reference won't have a bound multiplicity
        // in the last step by using our absolute arg ref instead of the one belonging to the specific
        // node are currenlty computing this expression at.
        EvaluationContext ec = new EvaluationContext(evalContext, absoluteArgRef);

        argVals[i] = new XPathPathExprEval().eval(absoluteArgRef, ec);
    } else {
        argVals[i] = args[i].eval(model, evalContext);
    }
}
```

I'm still thinking about how we could avoid doing this at this layer by making `TriggerableDag.evaluateTriggerables` or `TreeReference.contextualize()` but after some digging around I'm not sure it's doable.

PR #561, opened 2020-03-12
On repeat creation, don't double-evaluate expressions

lognaturel:
When reviewing https://github.com/opendatakit/javarosa/pull/537, I noticed that some expressions were evaluated twice on repeat creation. This PR prevents that. 

Additionally, no tests failed when `triggerTriggerables` was removed and only the smoke test failed when `initializeTriggerables` was removed. This PR adds coverage to demonstrate that both are required.

Added and ran automated tests.

This is narrow and effective. I didn't consider alternatives.

The only expected user impact is a performance improvement on repeat addition but either there would have to be lots of expressions in the repeat or many repeat instances for it to be noticeable. There's a risk to all calculations that need to happen on repeat add. Clearly we didn't have much coverage in this area and if we can think of other meaningful tests to add we should do so.

See tests.

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
+++ b/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
@@ -929,6 +929,84 @@ public class TriggerableDagTest {
     }
 
     /**
+     * Excercises the triggerTriggerables call in createRepeatGroup.
+     */
+    @Test
+    public void adding_repeat_instance_triggers_triggerables_outside_repeat_that_reference_repeat_nodeset() throws IOException {
+        Scenario scenario = Scenario.init("Form", html(
+            head(
+                title("Form"),
+                model(
+                    mainInstance(t("data",
+                        t("count"),
+                        t("repeat jr:template=\"\"",
+                            t("string")
+                        )
+                    )),
+                    bind("/data/count").type("int").calculate("count(/data/repeat)"),
+                    bind("/data/repeat/string").type("string")
+                )
+            ),
+            body(
+                repeat("/data/repeat",
+                    input("/data/repeat/string")
+                )
+            )
+        ));
+
+        scenario.createNewRepeat("/data/repeat");
+        scenario.createNewRepeat("/data/repeat");
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/javarosa/core/model/TriggerableDagTest.java
```diff
@@ -928,6 +928,84 @@ public void relative_paths_in_nested_repeats() throws IOException {
...
+                )
+            ),
+            body(
+                repeat("/data/repeat",
+                    input("/data/repeat/string")
+                )
+            )
+        ));
+
+        scenario.next();
```
ggalmazor: Maybe consider changing that into this?

```java
scenario.createNewRepeat("/data/repeat");
scenario.createNewRepeat("/data/repeat");
```

src/test/java/org/javarosa/core/model/TriggerableDagTest.java
```diff
@@ -928,6 +928,84 @@ public void relative_paths_in_nested_repeats() throws IOException {
         assertThat(scenario.answerOf("/data/outer[1]/inner[1]/count"), is(intAnswer(2)));
     }
 
+    @Test
+    public void adding_repeat_instance_triggers_triggerables_outside_repeat_that_reference_repeat_nodeset() throws IOException {
```
ggalmazor: Maybe add a comment with the justification for this test? Something that points out that we didn't know whether the phase 1 of the DAG's algorithm for creating new repeats was justified?

ggalmazor:
@lognaturel, I'll let you merge the PR in case you want to change anything before that.

src/test/java/org/javarosa/core/model/TriggerableDagTest.java
lognaturel: Good call.

src/test/java/org/javarosa/core/model/TriggerableDagTest.java
lognaturel: Sounds good!

PR #563, opened 2020-03-19
Feature/add week to date format

### LINKED ISSUE #544: Add support for %W (week number) to format-date and format-date-time
See https://forum.opendatakit.org/t/odk-xform-spec-proposal-add-week-number-to-format-date-and-format-date-time/23401

macdude357:
Closes #544

Added unit test

This was a fairly trivial fix
Users can now use %W in their forms


src/main/java/org/javarosa/core/model/utils/DateUtils.java
```diff
@@ -73,14 +74,15 @@ public DateFields () {
         public int minute; //0-59
         public int second; //0-59
         public int secTicks; //0-999 (ms)
+        public int week; // 1-52
```
lognaturel, line 77: A year can have 53 weeks, as 2020 does.

src/main/java/org/javarosa/core/model/utils/DateUtils.java
```diff
@@ -89,20 +91,25 @@ public DateFields () {
...
+            return new DateFields(year, month, day, hour, minute, second, secTicks, dow, week);
         }
 
-        public boolean check () {
-            return (inRange(month, 1, 12) && inRange(day, 1, daysInMonth(month - MONTH_OFFSET, year)) &&
-                    inRange(hour, 0, 23) && inRange(minute, 0, 59) && inRange(second, 0, 59) && inRange(secTicks, 0, 999));
+        public boolean check() {
+            return (inRange(month, 1, 12) && inRange(day, 1, daysInMonth(month - MONTH_OFFSET, year))
+                    && inRange(hour, 0, 23) && inRange(minute, 0, 59) && inRange(second, 0, 59)
+                    && inRange(secTicks, 0, 999) && inRange(week, 1, 52));
```
lognaturel, line 112: The week range should be 1-53. See https://en.wikipedia.org/wiki/ISO_week_date.

src/test/java/org/javarosa/core/model/utils/test/DateUtilsFormatTests.java
```diff
@@ -0,0 +1,44 @@
...
+        assertEquals("13", week);
+        
+        week = DateUtils.format(DateFields.of(2018, 1, 1, 10, 20, 30, 400), "%W");
+        assertEquals("01", week);
+
+        // Week of year is based on what year the first thursday is. 1/1/2017 was a Sunday so 
+        // it's actually the 52nd week of the previous year.
+        week = DateUtils.format(DateFields.of(2017, 1, 1, 10, 20, 30, 400), "%W");
+        assertEquals("52", week);
+
```
lognaturel, line 41: Would be good to add a test for Dec 29, 2020 or some other date in the 53rd week.

macdude357: added

src/main/java/org/javarosa/core/model/utils/DateUtils.java
macdude357: fixed

src/main/java/org/javarosa/core/model/utils/DateUtils.java
macdude357: fixed

lognaturel:
Thanks so much, @macdude357! As I was doing a last check for this, I went back to the XPath 3.1 spec and realized that in the [table for the picture string](https://www.w3.org/TR/xpath-functions-31/#rules-for-datetime-formatting), `W` is said to return an unpadded integer by default. I went ahead and made that change since we've already gone back and forth.

PR #564, opened 2020-03-19
Add setvalue tests to document intended behavior, convert existing tests to Scenario

### LINKED ISSUE #573: Empty value for setvalue is not accepted.
Not sure if this is javarosa or validate.

`setvalue=""` is considered an error.

https://github.com/XLSForm/pyxform/blob/f64f4a00f12e9bb6f2d20e238b3566f8f2008096/pyxform/tests/test_expected_output/when_column_setvalue_4.xml

https://github.com/XLSForm/pyxform/issues/438

lognaturel:
Closes #573 

Ran the tests.

I didn't need to convert the existing tests but since I was in that brain space and it was quick to do, I figured now was a good time. What I want to avoid is confusion over where to add any future `setvalue` tests we may want.

I ended up also fixing #573 here which came in after I added the initial set of tests. The crash was happening in `Parser.verifyBaseExpr`. I considered trying to accept empty string expressions everywhere but I decided that it's too risky. In practice, I can't think of another place where an expression is expected and an empty string would make sense. For now, I explicitly handled it for `setvalue`'s `value` attribute. 

The only user-facing change is to the `setvalue` `value` attribute parsing. It should only fix the bug.

See https://github.com/getodk/javarosa/pull/564/commits/44919e51cbc06d985935b9009514f6e07ade7417#diff-5f18730f1a6f1b79e1074750c1bee0a7R237

Using readonly calculates to display computed values to users is a useful recipe that's not explicitly documented: https://github.com/opendatakit/docs/issues/1198


src/test/java/org/javarosa/core/model/actions/SetValueActionTest.java
```diff
@@ -0,0 +1,96 @@
...
+                    )),
+                    bind("/data/readonly-field").readonly("1"),
+                    t("setvalue event=\"odk-instance-first-load\" ref=\"/data/readonly-field\" value=\"now()\"")
+                )
+            ),
+            body(
+                input("/data/readonly-field")
+            )));
+
+        assertThat(scenario.answerOf("/data/readonly-field"), is(notNullValue()));
```
seadowg, line 213: Good to be able to assert that the value is correct here. Maybe inject current time somehow?

src/test/java/org/javarosa/core/model/actions/SetValueActionTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void setvalue_setsValueOfReadOnlyField() throws IOException {
+        Scenario scenario = Scenario.init("Setvalue readonly", html(
+            head(
+                title("Setvalue readonly"),
+                model(
+                    mainInstance(t("data id=\"setvalue-readonly\"",
+                        t("readonly-field")
+                    )),
+                    bind("/data/readonly-field").readonly("1"),
+                    t("setvalue event=\"odk-instance-first-load\" ref=\"/data/readonly-field\" value=\"now()\"")
```
lognaturel, line 206: Add DSL method.

src/test/java/org/javarosa/core/model/actions/SetValueActionTest.java
```diff
@@ -0,0 +1,96 @@
...
+        assertThat(scenario.answerOf("/data/destination2"), is(intAnswer(11)));
+    }
+
+    /**
+     * This test demonstrates that read-only is just a UI concern and that calculates are still run on read-only fields.
+     * Questions around this behavior came up while discussing setvalue. It's here for documentation purposes pending a
+     * better place to put it.
+     */
+    @Test
+    public void calculate_evaluatedOnReadonlyFieldWithUI() throws IOException {
```
seadowg, line 288: This should go in a `CalculateTest/`ReadOnlyTest` or something similar I think

src/test/java/org/javarosa/core/model/actions/SetValueActionTest.java
```diff
@@ -60,104 +42,135 @@ public void after() {
...
-            int eventType = -1;
-            while (eventType != FormEntryController.EVENT_QUESTION) {
-                eventType = formEntryController.stepToNextEvent();
-            }
+        assertThat(scenario.answerOf("/data/destination"), is(dateTimeAnswer(DATE_NOW)));
 
-            DateTimeUtils.setCurrentMillisFixed(todayPlusDays(i));
-            formEntryController.answerQuestion(new DecimalData(i + 1), true);
-        }
+        DateTimeUtils.setCurrentMillisFixed(DATE_NOW + DAY_OFFSET); // shift the current time so we can test whether the setvalue action was re-fired
```
seadowg, line 112: Maybe nice to use another field to test this rather than time. Time is hard.

lognaturel:
Thanks for the review, @seadowg! This is much better.

PR #565, opened 2020-04-04
Handle relative paths with predicates

### LINKED ISSUE #540: Relative references aren't correctly expanded in nested repeats

lognaturel:
Closes #540

The solution to #540 in #546 was not complete. In particular, the form in the issue still fails because relative paths with predicates fail to parse.

I added a narrow failing test to illustrate the issue, fixed the bug, and then also added an integration/regression test using actual forms with the indefinite repeat pattern [in the documentation](https://docs.opendatakit.org/form-logic/#repeating-as-long-as-a-condition-is-met). There's a lot going on in that example and it's the kind of thing that could be broken by changes in `jr:count`, relative references, predicates and more. Only the nested case exercises the bug fixed here but I added the single repeat case for completeness.

We've had this kind of problem a few times now where the solution to a problem comes with narrow tests and somehow we don't go back to verify the original form. QA helps with this but probably reviewers should also make sure that a test with a form is in place.

Added the tests and verified the original form from the issue.

This is the only approach I could come up with. See af53cd9 for an explanation.

This is intended only as a bug fix. It modifies the `getReference()` method which is used to parse any XPath expression in a form definition. However, there's only a change in behavior for `ref.getRefLevel() > 0` which means only if there's a relative reference. So the risk would be to relative references with predicates. Those didn't work at all before so I think that the worst that could happen is that they're not fully fixed.

See issue.

No.

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathPathExpr.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathPathExpr.java
@@ -200,8 +200,8 @@ public class XPathPathExpr extends XPathExpression {
             if (step.predicates.length > 0) {
                 List<XPathExpression> v = new ArrayList<XPathExpression>(step.predicates.length);
                 Collections.addAll(v, step.predicates);
-                //int level = ref.getRefLevel() > 0 ? i - ref.getRefLevel() : i; // refLevel represents parenting steps
-                ref.addPredicate(i, v);
+                int level = ref.getRefLevel() > 0 ? i - ref.getRefLevel() : i; // refLevel represents parenting steps
+                ref.addPredicate(level, v);
             }
         }
         return ref;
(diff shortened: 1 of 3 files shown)


lognaturel:
@ggalmazor if you happened to have some time, you'd definitely be the best positioned to review this!

lognaturel:
@seadowg Could you please take a look if @ggalmazor doesn't have a chance by the end of your day Monday?

lognaturel:
This is timely: a user just experienced the bug at https://forum.opendatakit.org/t/infinite-loop-and-dynamically-delete-already-selected-options-from-list/25902/3

ggalmazor:
Hi! I know I'm late to the party (sorry!), but I wanted to say that I've taken a look and the solution sounds good to me. On top of that, I think that the new tests explain the scenario pretty well too.

PR #566, opened 2020-04-08
Add `FormIndex#getPreviousLevel` and `FormEntryController#jumpToNewRepeat`

seadowg:
These were added in as helpers in Collect (as part of https://github.com/opendatakit/collect/pull/3656) recently but seemed like they should actually live in JavaRosa.

Wrote new tests for the new features.

Should just add the new features. Now existing code was actually changed so this should be low risk to merge.

### DIFF
--- a/src/main/java/org/javarosa/form/api/FormEntryController.java
+++ b/src/main/java/org/javarosa/form/api/FormEntryController.java
@@ -303,6 +303,10 @@ public class FormEntryController {
 
     public void jumpToNewRepeatPrompt() {
         FormIndex repeatIndex = getRepeatGroupIndex(getModel().getFormIndex(), getModel().getForm());
+        if (repeatIndex == null) {
+            return;
+        }
+
         int repeatDepth = repeatIndex.getDepth();
 
         do {
@@ -316,8 +320,11 @@ public class FormEntryController {
         if (element instanceof GroupDef && ((GroupDef) element).getRepeat()) {
             return index;
         } else {
-            FormIndex previousLevel = index.getPreviousLevel();
-            return getRepeatGroupIndex(previousLevel, formDef);
+            if (index.getPreviousLevel() != null) {
+                return getRepeatGroupIndex(index.getPreviousLevel(), formDef);
+            } else {
+                return null;
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/test/java/org/javarosa/core/test/Scenario.java
```diff
@@ -180,6 +180,17 @@ public ValidateOutcome getValidationOutcome() {
...
+     * When using the result of this method for test assertions against {@link FormIndex#getReference()} we need to
+     * specify multiplicity on all steps. For example the following would pass for a form index pointing at a
+     * question at the top level of the form:
+     *
+     * <code>
+     * assertThat(formIndex.getReference(), is(getRef("/data/question[0]")));
+     * </code>
+     *
+     * If we used just used <code>getRef("/data/question")</code> the test would fail as the multiplicity defaults
+     * to <code>0</code>.
```
lognaturel, line 193: That makes it sound like it should pass. I think what you mean is that when using `getRef`, the multiplicity for `question` would be unbound (-1) but that for any reference used "for real" the multiplicity would have defaulted to 0. 

Maybe it's worth it to explain that `getRef` has no way of knowing whether a node without a multiplicity is an unbounded repeat or a singular group or question so it assumes the former. Adding an explicit `[0]` forces it to build a reference for the later. I had to go through and figure that out again just now.

src/main/java/org/javarosa/form/api/FormEntryController.java
```diff
@@ -286,15 +288,47 @@ public FormIndex deleteRepeat() {
...
+     *
      * @param language
      */
     public void setLanguage(String language) {
         model.setLanguage(language);
     }
+
+    /**
+     * Jump to the prompt to add a new repeat for the repeat the controller is currently in. If the current
+     * position in the form is no in a repeat nothing will happen.
```
lognaturel, line 306: "no in" -> "not in"

seadowg: It works in Scots...

JK I'll fix.

src/test/java/org/javarosa/core/test/Scenario.java
seadowg: > Maybe it's worth it to explain that getRef has no way of knowing whether a node without a multiplicity is an unbounded repeat or a singular group or question so it assumes the former. Adding an explicit [0] forces it to build a reference for the later. I had to go through and figure that out again just now.

This is the best explanation we've had so far so I'm going to try and use it!

PR #567, opened 2020-04-08
Support triggers for relative references

### LINKED ISSUE #540: Relative references aren't correctly expanded in nested repeats

lognaturel:
Closes #540 

Maddeningly enough, #565 still didn't fully address #540 because the test forms used had absolute references rather than the relative ones that `pyxform` now produces.

Added and ran automated test that matches user-provided form with relative references. I also tried the user-provided forms at https://forum.opendatakit.org/t/infinite-loop-and-dynamically-delete-already-selected-options-from-list/25902 and https://forum.opendatakit.org/t/error-in-indefinite-repeat-group-inside-another-repeat-group/24729.

No other solutions considered. There's a comment on `Triggerable.getTriggers` that shows that someone already spotted that passing in a `null` context reference when the context reference is known didn't make much sense. We know exactly what the context node is and that's what we should pass in. Then, in `XPathConditional.getTriggers`, it follows that we should use the contextualized node.

This should only fix the bug. It does make changes to the building of the DAG which is complex code and generally risky. However, as with other interventions for #540, the changes mostly only relate to relative references (but see keep reading for an exception). Users of XLSForm (the majority of ODK users) would only run into these modified code paths if they had nested repeats in their forms so I think this is low-risk. Prior to this change, the `expr.getTriggers` call on a relative reference always would have led to an `IllegalArgumentException` in `XPathConditional.getTriggers` so the only risk there is if the fix isn't complete.

The change to use a contextualized reference to get the predicate context affects every reference with a predicate. Predicates are used with repeats or secondary instances. I think that the test coverage here is sufficient to give confidence to the change with repeats. Expressions involving secondary instances are read-only so would not be part of the DAG.

https://forum.opendatakit.org/t/infinite-loop-and-dynamically-delete-already-selected-options-from-list/25902 and https://forum.opendatakit.org/t/error-in-indefinite-repeat-group-inside-another-repeat-group/24729

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/condition/Triggerable.java
+++ b/src/main/java/org/javarosa/core/model/condition/Triggerable.java
@@ -103,12 +103,7 @@ public abstract class Triggerable implements Externalizable {
     public abstract boolean isCascadingToChildren();
 
     public Set<TreeReference> getTriggers() {
-        Set<TreeReference> relTriggers = expr.getTriggers(originalContextRef);
-        Set<TreeReference> absTriggers = new HashSet<>();
-        for (TreeReference r : relTriggers) {
-            absTriggers.add(r.anchor(originalContextRef));
-        }
-        return absTriggers;
+        return expr.getTriggers(originalContextRef);
     }
 
     public TreeReference contextualizeContextRef(TreeReference anchorRef) {
--- a/src/main/java/org/javarosa/core/model/condition/Triggerable.java
+++ b/src/main/java/org/javarosa/core/model/condition/Triggerable.java
@@ -103,7 +103,7 @@ public abstract class Triggerable implements Externalizable {
     public abstract boolean isCascadingToChildren();
 
     public Set<TreeReference> getTriggers() {
-        Set<TreeReference> relTriggers = expr.getTriggers(null);  /// should this be originalContextRef???
+        Set<TreeReference> relTriggers = expr.getTriggers(originalContextRef);
         Set<TreeReference> absTriggers = new HashSet<>();
         for (TreeReference r : relTriggers) {
             absTriggers.add(r.anchor(originalContextRef));
--- a/src/main/java/org/javarosa/xpath/XPathConditional.java
+++ b/src/main/java/org/javarosa/xpath/XPathConditional.java
@@ -124,8 +124,8 @@ public class XPathConditional implements IConditionExpr {
                 }
 
                 //we can't generate this properly without an absolute reference
-                if(!ref.isAbsolute()) { throw new IllegalArgumentException("can't get triggers for relative references");}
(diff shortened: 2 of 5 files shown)


lognaturel:
@seadowg I'm thinking we can do an interactive review tomorrow?

seadowg:
@lognaturel sounds good

lognaturel:
During interactive review with @seadowg, I noticed that the loop bounds in the modified `XPathConditional.getTriggers` didn't make sense -- we were setting the bounds using the relative reference and then using the loop index to index into the contextualized reference. The regression test passed because the predicate trigger `../inner_count` was also used outside the predicate. That meant it didn't matter if the predicate trigger was totally wrong. In fact, I found out I could comment out that entire loop and that no test would fail.

In 67eafbb, I added a narrow failing test where that loop was necessary. Then I also added a regression test at 8f1fb44 which allowed me to find another bug.

I believe that my statements about relative safety of this change from the PR description hold.

e997f68 is a change that affects every single expression in a form but I think it can confidently be verified. @seadowg, please see if you confirm. That is, it should be clear to you that the call on `anchor` that I removed always comes after another call on `anchor`. If there's any doubt, we can remove that commit.

During review, I mumbled some things about `current()` and convinced myself that the form design was flawed. Luckily, I was wrong. In `indefiniteRepeatJrCountExpression_inNestedRepeat_withRelativePaths_addsRepeatsUntilConditionMet`, when we have an expression like `../inner_repeat[position() = ../inner_count]/add_more`, we want the `../inner_count` predicate expression to be relative to the `../inner_repeat` expression it's on. That means using `../` is correct. Adding `current()` would make the expression relative to the bound node which would be `/data/outer_repeat/target_count`. That turns out to also be fine but it's not necessary and I think we should avoid use of `current()` unless absolutely needed since it's not part of XPath and is instead a XForms/XSLT addition that is (even) less familiar than core XPath.

@seadowg let me know if you'd prefer to talk through this again. I know there are a lot of details.

lognaturel:
After writing out my little spiel about `current()`, I realized that the original context was never saved when recursing to get predicate triggers and so `current()` expressions in predicates couldn't be correctly evaluated. The two commits above should hopefully help clarify what all this means.

lognaturel:
@seadowg and I talked through this one more time since there's a lot going on. We noticed that one of the tests wasn't fully checking what we wanted and I fixed that. He's agreed to take another ~15min dive into https://github.com/getodk/javarosa/pull/567/commits/e997f6861f944bd30444d431789c93d6e6a331e0 to verify that he's very confident `anchor` does always get called. If that's the case, we'll merge and go to QA.

lognaturel:
@seadowg made the agreed-upon dive and feels fairly confident. He gave me back an explanation that also matches my understanding and rationale.

PR #569, opened 2020-04-15
Allow zero arg normalize-space and string-length calls (#468)

### LINKED ISSUE #468: Add zero-argument variants for string-length and normalize-space functions
See https://www.w3.org/TR/1999/REC-xpath-19991116/#function-string-length

abdulazizali77:
Hi @lognaturel , PTAL

Closes #468 

Four extra test cases to verify normalize-space() and string-length() calls in the normal case and in the exceptional case

The new behaviour might demand some discussion. If the two functions are called without a model/context, an `XPathException` will be thrown, one alternative is to not handle the function if it is not called without a model/context and throw `XPathUnhandledException` instead.
The idiom of the ifblock might be up for discussion too, whether we want to have a separate block for the empty args case instead.

Preexiting behaviour with single argument has not been changed. A slim risk is that a preexisting application calling no args normalize-space and string-length expecting `XPathUnhandledException` will instead get a valid `XPathNodeSet` instead. 

None

https://github.com/opendatakit/docs/issues/1214

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -387,12 +387,22 @@ public class XPathFuncExpr extends XPathExpression {
             return toString(argVals[0]).startsWith(toString(argVals[1]));
         } else if (name.equals("ends-with") && args.length == 2) {
             return toString(argVals[0]).endsWith(toString(argVals[1]));
-        } else if (name.equals("string-length")) {
-            assertArgsCount(name, args, 1);
-            return stringLength(argVals[0]);
-        } else if (name.equals("normalize-space")) {
-            assertArgsCount(name, args, 1);
-            return normalizeSpace(argVals[0]);
+        } else if (name.equals("string-length") && args.length <= 1) {
+            Object arg;
+            if (args.length == 1) {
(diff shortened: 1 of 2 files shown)


lognaturel:
Thanks, @abdulazizali77! This is really great. Do you have prior experience with XPath?

> If the two functions are called without a model/context, an XPathException will be thrown

It's a good case to consider and have a test for but I don't think it's possible to achieve in a form. We don't expect that folks would be clients of this library outside of a form context. I think that what you've done with not explicitly handling the case and letting the exception come from the `eval` call makes sense.

What does bother me a bit is that if these functions are called with 2+ arguments the exception thrown will say that the function isn't handled. It was this way before and is consistent with other string functions so probably not something to deal with in this PR. As you've likely noticed, the way args checks are done is inconsistent with some functions using `assertArgsCount`, others using `checkArity`, others directly throwing exceptions, etc. 

> whether we want to have a separate block for the empty args case instead.

I like the nesting you've done.

abdulazizali77:
thanks @lognaturel! 
long time ago i did use xslt as just a consumer/user, ive had to reremind myself /reread the specs (sporadically) yet again to fix this :D

lognaturel:
Thanks again! And welcome back to the world of XML. 😉

PR #570, opened 2020-04-20
Fix unexpected prompt to add not relevant repeat instances

grzesiek2010:
Closes #https://github.com/getodk/collect/issues/2992

I tested the changes manually and added automated tests.

If repeatable group doesn't contain any questions or is marked as not relevant a user shouldn't be asked to add it. I fixed those two cases. There is one more possible case that could be fixed - the case where a group has children but all of them are not relevant. However I decided to leave it as is because it's a case that would require going through all children that also might be other nested groups evaluating them etc. So since it's a rare case and the current behavior is not a big problem I think it doesn't make sense to implement complex and risky code for it.

The change is related to reparable groups. It should just fix cases where a group is marked as not relevant or doesn't have any elements inside.

I used: 
[relevantRepeats.xlsx](https://github.com/getodk/javarosa/files/4502073/relevantRepeats.xlsx)

No.

### DIFF
--- a/src/test/java/org/javarosa/xform/parse/RepeatGroupsRelevanceTest.java
+++ b/src/test/java/org/javarosa/xform/parse/RepeatGroupsRelevanceTest.java
@@ -1,0 +1,39 @@
+package org.javarosa.xform.parse;
+
+import org.javarosa.core.model.FormDef;
+import org.javarosa.core.test.FormParseInit;
+import org.javarosa.form.api.FormEntryModel;
+import org.junit.Test;
+
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.junit.Assert.assertFalse;
+
+public class RepeatGroupsRelevanceTest {
+
+    @Test
+    public void isRepeatRelevantTest() {
+        FormParseInit fpi = new FormParseInit(r("repeat_groups_relevance.xml"));
+        FormEntryModel formEntryModel = fpi.getFormEntryModel();
+        FormDef formDef = fpi.getFormDef();
+
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+
+        // repeatable group made not relevant by the previous question
+        assertFalse(formDef.isRepeatRelevant(formDef.getChildInstanceRef(formEntryModel.getFormIndex())));
+
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+
+        // repeatable group marked as not relevant
+        assertFalse(formDef.isRepeatRelevant(formDef.getChildInstanceRef(formEntryModel.getFormIndex())));
+
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+        formEntryModel.setQuestionIndex(formEntryModel.incrementIndex(formEntryModel.getFormIndex()));
+
+        // empty repeatable group
+        assertFalse(formDef.isRepeatRelevant(formDef.getChildInstanceRef(formEntryModel.getFormIndex())));
+    }
+}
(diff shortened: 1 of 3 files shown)


lognaturel:
Great tests to guide this! We've moved towards using the `Scenario` class and putting form defs inline so the tests contain more context. I realize this isn't explicitly documented and have added an issue at #571. I've also made the changes in https://github.com/getodk/javarosa/compare/master...lognaturel:relevance-hm so you can cherry-pick them. Hopefully you'll agree that this makes the tests and their implications easier to understand. I also separated out the different cases that are being verified and renamed the test class to clearly identify what it relates to.

The underlying issue is that when there's a static `false()` expression for the repeat relevance, it's not included in the DAG where `getTriggerableForRepeatGroup` gets the relevance expression from.

This whole area of code seems problematic. The `isRepeatRelevant` method is called by `FormEntryModel.isIndexRelevant` which I'm pretty sure does a lot of extra work. In particular, it checks every ancestor. However, this should be taken care of by the way the main instance is initially constructed and the DAG. And then `FormDef.isRepeatRelevant` checks just the immediate parent which doesn't make much sense. If relevance didn't propagate, all ancestors would need to be checked.

I think the following would be a better implementation for `isRepeatRelevant`:
```
TreeElement node = mainInstance.resolveReference(repeatRef);
return node == null || mainInstance.resolveReference(repeatRef).isRelevant()
```

This is a bit of a strange fix because there is so much unnecessary stuff going on. But I suppose it's better to be conservative. @grzesiek2010 what was your reasoning for looking at the template's relevance instead of the repeat instance's relevance with
 `mainInstance.resolveReference(repeatRef).isRelevant()`? Did you carefully consider whether there could be a case where the template's relevance was false but the specific instance's relevance was true? I can't think of one off the top of my head but I haven't thought about it deeply.

> the case where a group has children but all of them are not relevant

This does seem a lot more likely than an empty group. I wonder whether the empty group case is even a good one to handle at all. It wasn't mentioned in getodk/collect#2992 and it feels like it would always be a form design mistake. If I made that mistake when designing a form it seems like I'd rather have the client prompt me so that I can know to fix it. What was your reasoning for adding that special case?

Happy to do an interactive review if you want to discuss further!

grzesiek2010:
I tested different cases using form like this one:
image

>  Did you carefully consider whether there could be a case where the template's relevance was false but the specific instance's relevance was true?

as I noticed this already works as expected `gr1` in the form above where relevance is abased on a previous question.

I used template's relevance for static values like in `gr2` where it's always no but after thinking about it more and taking into account what you said:
> I wonder whether the empty group case is even a good one to handle at all [...] it feels like it would always be a form design mistake

I could agree that we could both those cases (Empty groups and groups statically marked as not relevant) just ignore and treat as a form design mistake.
In such a case there is just one case we could consider: repeat groups where all the children are not relevant.

What do you think?

lognaturel:
> works as expected gr1 in the form above where relevance is based on a previous question.

Great! What I had in mind was some kind of expression that would change based on the context of the repeat instance. See https://github.com/getodk/javarosa/pull/581/commits/54fa3aea008efe55008861eb6fe6f14d61aaf5fa#diff-02a3eeffc5c34a1d10f7680159593ac9R173 for an example I came up with.

Thinking about this more, I think we should make the simplifying changes not only to `FormDef.isRepeatRelevant` but also to `FormEntryModel.isIndexRelevant`. I think I was able to come up with tests that give us good coverage for the change and I think `TreeElement.setRelevant`/`getRelevant` can be traced to verify that they handle changes to ancestor relevance. I've put an alternative PR at #581 and would love to hear what you think. Can you think of additional test cases? Is my reasoning ok?

> repeat groups where all the children are not relevant

To me, that also seems like the responsibility of the form designer to consider.

src/test/java/org/javarosa/xform/parse/RepeatGroupsRelevanceTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Test;
+
+import static org.javarosa.test.utils.ResourcePathHelper.r;
+import static org.junit.Assert.assertFalse;
+
+public class RepeatGroupsRelevanceTest {
+
+    @Test
+    public void isRepeatRelevantTest() {
+        FormParseInit fpi = new FormParseInit(r("repeat_groups_relevance.xml"));
```
lognaturel, line 15: One thing I did in #581 is split this form and test into several smaller ones at https://github.com/getodk/javarosa/pull/581/files#diff-02a3eeffc5c34a1d10f7680159593ac9R62.

I just want to point out that this is a "do as I say, not as I do" case because I did the same kind of monolithic test at https://github.com/getodk/javarosa/pull/578/commits/c954d0bc29e7c521702a42670de856a92859922c#diff-c9232740659aac312cd026fe65a452f8R24 and @seadowg had to tell me to split it up. I think the split has to be a later pass once we know what's going on and it's hard to remember to do that!

lognaturel:
@grzesiek2010 and I talked about this and are closing in favor of #581.

PR #572, opened 2020-04-23
Fixed: Deleting last repeat does not result in calculates

grzesiek2010:
Closes #https://github.com/getodk/collect/issues/3016

I added an automated test and tested this change manually with Collect.

It's just a bug fix.

It just fixes the but it shouldn't cause any problems so I'm not even sue if it requires testing... maybe just in case.

The form attached to the [issue](https://github.com/getodk/collect/issues/3016) would be good for testing if needed.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -206,6 +206,7 @@ public class TriggerableDag {
             }
         } else {
             triggerTriggerables(mainInstance, evalContext, deleteRef, new HashSet<>(0));
+            evaluateChildrenTriggerables(mainInstance, evalContext, deletedElement, false, new HashSet<>(0));
         }
     }
 
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -185,22 +185,27 @@ public class TriggerableDag {
         //After a repeat group has been deleted, the following repeat groups position has changed.
         //Evaluate triggerables which depend on the repeat group reference. Directly or indirectly.
         String repeatGroupName = deletedElement.getName();
-        for (int i = deletedElement.getMultiplicity(); i < parentElement.getChildMultiplicity(repeatGroupName); i++) {
-            TreeElement repeatGroup = parentElement.getChild(repeatGroupName, i);
+        boolean lastRepeatGroup = deletedElement.getMultiplicity() == parentElement.getChildMultiplicity(repeatGroupName);
+        if (!lastRepeatGroup) {
+            for (int i = deletedElement.getMultiplicity(); i < parentElement.getChildMultiplicity(repeatGroupName); i++) {
+                TreeElement repeatGroup = parentElement.getChild(repeatGroupName, i);
 
-            Set<QuickTriggerable> alreadyEvaluated = triggerTriggerables(mainInstance, evalContext, repeatGroup.getRef(), new HashSet<>(0));
-            publishSummary("Deleted", repeatGroup.getRef(), alreadyEvaluated);
+                Set<QuickTriggerable> alreadyEvaluated = triggerTriggerables(mainInstance, evalContext, repeatGroup.getRef(), new HashSet<>(0));
+                publishSummary("Deleted", repeatGroup.getRef(), alreadyEvaluated);
 
-            if (repeatGroup.getRef().equals(deleteRef)) {
-                // Evaluate the children triggerables only once, for the deleted repeat group.
-                //  Only children of the deleted repeat group have actually changed (they're gone) and thus calculations depend
-                //  on the must be re-evaluated, the following repeat groups have been shifted along with their children.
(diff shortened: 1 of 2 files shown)


src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -185,22 +185,27 @@ public void deleteRepeatGroup(FormInstance mainInstance, EvaluationContext evalC
...
-                //  Only children of the deleted repeat group have actually changed (they're gone) and thus calculations depend
-                //  on the must be re-evaluated, the following repeat groups have been shifted along with their children.
-                //  If there are calculations - regardless if inside the repeat or outside - that depend on the following
-                //  repeat group positions, they will fired cascade by the above code anyway.
-                //  Unit test for this scenario:
-                //  Safe2014DagImplTest#deleteThirdRepeatGroup_evaluatesTriggerables_indirectlyDependentOnTheRepeatGroupsNumber
-                evaluateChildrenTriggerables(mainInstance, evalContext, repeatGroup, false, alreadyEvaluated);
+        boolean lastRepeatGroup = deletedElement.getMultiplicity() == parentElement.getChildMultiplicity(repeatGroupName);
+        if (!lastRepeatGroup) {
+            for (int i = deletedElement.getMultiplicity(); i < parentElement.getChildMultiplicity(repeatGroupName); i++) {
```
grzesiek2010, line 190: The problem was that after deleting a group we navigate through other repeats and do what we need including recalculations. However if we delete the last group nothing happens (on master) because the code in this for loop won't be called.

lognaturel:
Great that this turns out to be "just" a bounds issue!

How did you pick where to write the tests? I don't thing this has anything to do with XForm parsing so I'm surprised to see it in that package. I think it should go in the existing `TriggerableDagTest` after the tests that verify deleting intermediary repeats. 

I described the current patterns for JavaRosa tests at https://github.com/getodk/javarosa/pull/570#issuecomment-618120624 but I imagine you did this before. In general, the more targeted a test can be, the more useful it is likely to be. I think that when we were creating separate test form documents for tests we got into a bad pattern of squeezing many tests into one form definition. With the form building DSL it's easier to have really small, targeted forms inline.

Here are the tests I would have written:

```
    @Test
    public void deleteLastRepeat_evaluatesTriggerables() throws IOException {
        Scenario scenario = Scenario.init("Delete last repeat instance", html(
            head(
                title("Delete last repeat instance"),
                model(
                    mainInstance(t("data id=\"delete-last-repeat-instance\"",
                        t("repeat-count"),

                        t("repeat",
                            t("question")),
                        t("repeat",
                            t("question")),
                        t("repeat",
                            t("question"))
                    )),
                    bind("/data/repeat-count").type("int").calculate("count(/data/repeat)")
                ),
            body(
                repeat("/data/repeat",
                    input("question"))
            ))));

        assertThat(scenario.answerOf("/data/repeat-count"), is(intAnswer(3)));

        scenario.removeRepeat("/data/repeat[2]");
        assertThat(scenario.answerOf("/data/repeat-count"), is(intAnswer(2)));
    }

    @Test
    public void deleteLastRepeat_evaluatesTriggerables_indirectlyDependentOnTheDeletedRepeat() throws IOException {
        Scenario scenario = Scenario.init("Delete last repeat instance", html(
            head(
                title("Delete last repeat instance"),
                model(
                    mainInstance(t("data id=\"delete-last-repeat-instance\"",
                        t("summary"),

                        t("repeat",
                            t("question", "a")),
                        t("repeat",
                            t("question", "b")),
                        t("repeat",
                            t("question", "c"))
                    )),
                    bind("/data/summary").type("string").calculate("concat(/data/repeat/question)")
                ),
                body(
                    repeat("/data/repeat",
                        input("question"))
                ))));

        assertThat(scenario.answerOf("/data/summary"), is(stringAnswer("abc")));

        scenario.removeRepeat("/data/repeat[2]");
        assertThat(scenario.answerOf("/data/summary"), is(stringAnswer("ab")));
    }
```

I put the 3 repeat instances in the form definition but adding them after like the other repeat delete tests do would also make sense.

You can see that the second test fails with your solution.

grzesiek2010:
> How did you pick where to write the tests? I don't thing this has anything to do with XForm parsing so I'm surprised to see it in that package. I think it should go in the existing TriggerableDagTest after the tests that verify deleting intermediary repeats.

I agree it was just a random temporary place since it was a draft pr.

> I described the current patterns for JavaRosa tests at #570 (comment) but I imagine you did this before. In general, the more targeted a test can be, the more useful it is likely to be. I think that when we were creating separate test form documents for tests we got into a bad pattern of squeezing many tests into one form definition. With the form building DSL it's easier to have really small, targeted forms inline.

Thanks I'll investigate and follow that new pattern.

grzesiek2010:
> As described above, the deletion of values inside of the last repeat should trigger calculations that depend on them. See the second test I wrote.

Ah ok I missed it. Fixed.

lognaturel:
One additional case I could think of is if there is only one repeat and that repeat is deleted. I've verified that works and I think it's ok not to have an explicit test for it.

PR #576, opened 2020-05-19
Augment geo function tests with string concatenation, use Scenario everywhere

lognaturel:
Documents behavior discussed [on the forum](https://forum.getodk.org/t/use-area-with-geopoint-not-geotrace-nor-geoshape/26516/3?u=ln)

Ran tests.

No alternatives considered. I moved the area tests to have consistent naming. I converted all those tests to `Scenario` and cleaned up form definitions that are no longer used. While there, I noticed that the `distance` test should also test string input.

No user-facing changes.

No.

https://github.com/getodk/docs/issues/1222


src/test/java/org/javarosa/core/util/GeoDistanceTest.java
```diff
@@ -112,4 +114,60 @@ public void distance_isComputedForGeoshape() throws IOException {
...
+                    bind("/data/point2").type("geopoint"),
+                    bind("/data/point3").type("geopoint"),
+                    bind("/data/point4").type("geopoint"),
+                    bind("/data/point5").type("geopoint"),
+                    bind("/data/concat").type("string").calculate("concat(/data/point1, ';', /data/point2, ';', /data/point3, ';', /data/point4, ';', /data/point5)"),
+                    bind("/data/distance").type("decimal").calculate("distance(/data/concat)")
+                )
+            ),
+            body(
+                input("/data/point1")
```
seadowg, line 143: Does this need to have an input?

lognaturel: Removed the entire body for this one and the corresponding area test. Thanks!

lognaturel:
> It looks like there is a one line drop in coverage (an if statement that's now never true).

What strange behavior this points to. The coverage before was accidental -- sometimes there was a point with all components of value 0. I have added a minimal test to document it.

lognaturel:
Also a good reminder of how skeptical we should be of coverage stats. We went from a branch being accidentally hit to a test that actually verifies and documents behavior and coverage didn't change.

PR #578, opened 2020-05-23
Consider instances in TreeReference comparison and set null instanceName for main instance on deserialization

### LINKED ISSUE #449: Referring to node in secondary instance with path that also exists in primary instance can result in a dependency cycle

lognaturel:
Closes #449

This is pretty high-risk and would be good to do an interactive review on.

Thought really hard about what was going on and tried to write as specific of tests as possible.

The change to `TreeReference` is the same as in #450 with the addition of an implementation for `hashCode`. It also avoids using a custom `Objects` class now that Collect has a `minSdkVersion` of 21. I feel good about this change and don't really see alternatives.

This unmasks some inconsistencies in how references in form definitions are resolved by JavaRosa and in how the primary instance was represented before and after serialization. The main symptom when we tried to release this previously is that constraints were broken - https://github.com/getodk/javarosa/issues/475. The reason for this is that before serialization, the `instanceName` of the primary instance root node was null and after deserialization, it was populated. In some code paths, this led to references to primary instance nodes with an explicit instance (e.g. `instance("My cool form")/data/a` instead of `/data/a`). This didn't matter before because instances were ignored when doing reference comparison. Constraints were broken because in the evaluation process there ended up being a comparison between `instance("My cool form")/data/a` and `/data/a` that used to pass but now fails. This comparison is made in `XPathPathExpr.getRefValue` to identify whether the latest value entered by the user should be used.

Even though it's been working forever, it's not right that the primary instance's `instanceName` would be different before and after serialization. I decided to explicitly set it on deserialization. This is the counterpart to `FormInstanceParser.buildInstanceStructure` which creates the primary instance initially and sets `instanceName` to `null`. I think that in an ideal world there wouldn't be this kind of special case for the primary instance name and some other mechanism would be used to identify that an instance is primary. However, that would require a lot of high-risk refactoring throughout the codebase.

The goal is for this to only unlock new behavior. However' it's high risk because as we saw with #450, it does touch anything having to do with `TreeReference` comparison or hashing. The places where `TreeReference`s are compared are:
`EvaluationContext.rescope`
`FormDef.getElementsFromReferences`
`FormDef.findQuestionByRef`
`FormInstanceParser.verifyRepeatMemberBindings` -- true case never hit
`TreeReference.intersect` -- true case never hit
`TriggerableDag.deleteRepeatGroup`
`XPathPathExpr.getRefValue`
`XPathPathExpr.pivot` -- not tested at all
`XPathReference.equals` -- false case never hit

One nice thing about `Scenario` that I realized writing up this list is that because it gives us a single entry point into tests, we can do things like immediately serialize all form defs on `Scenario.init`. I did this and ran the tests and one failed in `ChoiceNameTest`. I initially thought it could be a problem with reference comparison but I'm pretty sure it's just that the test wasn't actually mimicking client behavior.

The `jr:choice-name` function handler (see `FormDef.initEvalContext`) makes a call to populate dynamic choices, but only if they haven't already been populated. In the case of an unserialized form definition, the choices were populated on form initialization as an empty list. And then when the `jr:choice-name` call was evaluated, the choices weren't repopulated. After deserialization, though, the initialization step didn't happen. That meant that when the `jr:choice-name` call was evaluated, the choices were populated.

But Collect will explicitly call `populateDynamicChoices` on displaying the choice list for cities. So I updated the test to reflect this and now the behavior is the same with and without serialization.

See form at https://github.com/getodk/collect/issues/3316.

No.


src/main/java/org/javarosa/core/model/instance/DataInstance.java
```diff
@@ -23,13 +23,15 @@
...
  */
 public abstract class DataInstance<T extends AbstractTreeElement<T>> implements Persistable {
-
-
     /** The integer Id of the model */
     private int recordid = -1;
 
-    /** The name for this data model */
+    /**
+     * A unique name for this instance. The value of the id attribute in the case of a secondary instance or the title
```
lognaturel, line 30: It's strange to describe the how this field is used in this comment. However, I couldn't find a better place for it and it seemed valuable to me for making sense of debug contents. I can remove it if others disagree, though.

src/main/java/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -100,6 +100,10 @@
     private String namespace;
     private String namespacePrefix;
 
+    /**
+     * The name of the instance that this node is in. The value of the id attribute in the case of a secondary
```
lognaturel, line 104: Again, maybe this information shouldn't be here.

src/test/java/org/javarosa/core/model/instance/TreeReferenceSerializationTest.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.javarosa.core.util.XFormsElement.title;
+
+import java.io.IOException;
+import org.javarosa.core.test.Scenario;
+import org.javarosa.core.util.externalizable.DeserializationException;
+import org.junit.Test;
+
+public class TreeReferenceSerializationTest {
+    @Test
+    public void mainInstanceName_afterDeserialization_isNull() throws IOException, DeserializationException {
```
seadowg, line 24: `..isStillNull`

seadowg: Maybe split this up or make it clearer this is testing three things

seadowg:
Couple of comments and a merge conflict to fix here!

src/test/java/org/javarosa/core/model/instance/TreeReferenceSerializationTest.java
lognaturel: Thank you, that does make things a lot clearer!

PR #595, opened 2020-09-15
Fix relevance for prompts

seadowg:
Closes https://github.com/getodk/collect/issues/4090.

This changes how repeat relevance is calculated so that it checks the template as well as the node.

Added new tests and verified in Colllect.

There are a few things to improve:

- [x] The `@Ignored` `FormDefTest` test needs rewritten so it works with this new implementation
- [ ] We should remove `FormDef#isRepeatRelevant`

Should just fix the bug! Might be good to check any weird repeat relevance cases that can be thought of.

This is a good form to check the fix works: https://docs.google.com/spreadsheets/d/1pc9LPOR8s5C2uFjMj1JbuF6_PPLQ_Q2l6X6ytAZvul0/edit#gid=0


src/main/java/org/javarosa/form/api/FormEntryModel.java
```diff
@@ -383,19 +375,20 @@ public boolean isIndexRelevant(FormIndex index) {
...
         boolean isRepeatJuncture = (getEvent(index) == FormEntryController.EVENT_REPEAT_JUNCTURE);
 
-        boolean relevant;
         if (isAskNewRepeat) {
-            relevant = form.isRepeatRelevant(ref) && form.canCreateRepeat(ref, index);
+            if (!form.canCreateRepeat(ref, index)) {
+                return false;
+            }
+
+            return form.isRepeatRelevant(ref);
```
seadowg, line 383: We most likely need to make a change here to get the test pass. It would be great if we could grab the `GroupDef` for the index/reference and then call something like:

```java
groupDef.isNewRepeatRelevant()
```

This probably means adding the relevance expression to the `GroupDef` when we parse the form.

lognaturel:
I did a little more digging and realized that the repeat template is actually part of the `mainInstance` which is a tree of `TreeElement`s representing the instance being filled in. Templates are represented as `TreeElement`s with a multiplicity of `TreeReference.INDEX_TEMPLATE`. Like all `TreeElement`s, they keep track of their relevance. I think this leads to a straightforward change to `isRepeatRelevant`:

```
public boolean isRepeatRelevant(TreeReference repeatRef) {
    TreeElement repeatNode = mainInstance.resolveReference(repeatRef);

    if (repeatNode == null) {
        TreeElement templateNode = mainInstance.getTemplate(repeatRef);
        return templateNode.isRelevant();
    }

    return repeatNode.isRelevant();
}
```

It could also be written as a boolean expression but this feels clearer to me. It also doesn't really feel like it belongs in `FormDef`. It could go in `FormEntryModel`.

I don't think there's anything else from the logic at https://github.com/getodk/javarosa/pull/581/files#diff-b60f702426880ccc5f1b079019104bd4L537 that we need. To be extra safe, it might make sense to have a test where there's an outer group that's irrelevant and check that it's indeed impossible to add an instance of an inner repeat. But it shouldn't even be possible to navigate inside the outer group to try. Looking forward to seeing what you think, @seadowg.

Side note: I forgot the `mainInstance` has templates because I think of it as concrete. And also the multiplicity constants are in `TreeReference`.

lognaturel:
And to be extra clear, would be great if you could finalize the PR, @seadowg! Since this is a bug, I think we can do a point release. And I think it would be ok to change the public interface by getting rid of `FormDef.isRepeatRelevant`. We don't know of many non-ODK tools using this library. And for those that do, it's really unlikely they're using that particular method.

src/test/java/org/javarosa/core/model/RepeatTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static org.hamcrest.Matchers.is;
+import static org.javarosa.core.util.BindBuilderXFormsElement.bind;
+import static org.javarosa.core.util.XFormsElement.*;
+import static org.junit.Assert.assertThat;
+
+public class RepeatTest {
+
+    @Test
+    public void whenRepeatIsNotRelevant_isSkipped() throws Exception {
```
lognaturel, line 15: I think it'd be helpful for the name to be a little more specific. Maybe just adding `repeatPromptIsSkipped`.

seadowg:
> And I think it would be ok to change the public interface by getting rid of FormDef.isRepeatRelevant

@lognaturel I agree but there is test coverage to move around, so I think for this PR it'd be best to just add the implementation in `FormDef` for the moment.

seadowg:
@lognaturel ach it looks like the template solution doesn't work for dynamic relevance. I've added a test and will keep exploring.

src/main/java/org/javarosa/core/model/GroupDef.java
```diff
@@ -17,6 +17,9 @@
 package org.javarosa.core.model;
 
 import org.javarosa.core.model.actions.ActionController;
+import org.javarosa.core.model.condition.EvaluationContext;
```
lognaturel, line 20: Revert changes.

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -528,19 +528,30 @@ private void evaluateTriggerable(FormInstance mainInstance, EvaluationContext ev
         // the repeat are only triggered for the changed instance. This is important for performance.
         TreeReference contextRef = affectsAllRepeatInstances ? toTrigger.getContext() : toTrigger.getContext().contextualize(changedRef);
 
-        List<EvaluationResult> evaluationResults = new ArrayList<>(0);
         // In general, expansion will have no effect. It only makes a difference if affectsAllRepeatInstances is true in
         // which case the triggerable will be applied for every repeat instance.
-        for (TreeReference qualified : evalContext.expandReference(contextRef))
+        List<TreeReference> qualifiedReferences = evalContext.expandReference(contextRef);
+
+        // We want to up a repeat's template as well
```
lognaturel, line 535: update?

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -46,6 +46,8 @@
 import static org.javarosa.xform.parse.RangeParser.populateQuestionWithRangeAttributes;
 
 import java.io.ByteArrayInputStream;
+import java.io.DataInputStream;
```
lognaturel, line 49: Revert.

lognaturel (changes requested):

src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -173,6 +174,7 @@ public void repeatIsIrrelevant_whenGrandparentRelevanceSetToFalse() throws IOExc
     }
 
     @Test
+    @Ignore("Test needs rewritten")
```
lognaturel, line 177: This test could possibly have helped us notice the bug here. The relevance of the inner group is based on the position of the **outer group** but I think I convinced myself it was based on the position of the inner group (and that's what I was thinking when we talked, @seadowg).

In the original test, the first assertion is:
```
scenario.next();
scenario.next();
assertThat(scenario.refAtIndex(), is(getRef("/data/outer[0]/inner[1]")));
```

This is wrong because of the bug this PR fixes. Any inner group should be non-relevant based on the outer group's index. The form definition instance only has one instance of the inner repeat. It turns out the event emitted is actually a repeat prompt (because of this bug). The rest of the test makes sense.

Below is a more explicit proposal. I put two instances of the outer repeat and two inner instances in the first outer instance. That way we can check that initially none of the inner repeats in the first outer repeat are relevant and that there's no prompt. 

```java
Test
public void nestedRepeatRelevance_updatesBasedOnParentPosition() throws IOException {
    Scenario scenario = Scenario.init("Nested repeat relevance", html(
        head(
            title("Nested repeat relevance"),
            model(
                mainInstance(t("data id=\"nested-repeat-relevance\"",
                    t("outer",
                        t("inner",
                            t("q1")
                        ),
                        t("inner",
                            t("q1")
                        )
                    ),
                    t("outer",
                        t("inner",
                            t("q1")
                        )
                    ),
                    t("relevance-condition", "0")
                )),
                bind("/data/relevance-condition").type("string"),
                bind("/data/outer/inner").relevant("position(..) mod 2 = /data/relevance-condition")
            ),
            body(
                repeat("/data/outer",
                    repeat("/data/outer/inner",
                        input("/data/outer/inner/q1")
                    )
                ),
                input("/data/relevance-condition")
            ))));

    FormDef formDef = scenario.getFormDef();
    FormEntryModel formEntryModel = new FormEntryModel(formDef);

    // For ref /data/outer[0]/inner[0], the parent position is 1 so the boolean expression is false. That means
    // none of the inner groups in /data/outer[0] can be relevant.
    assertThat(formEntryModel.isIndexRelevant(scenario.indexOf("/data/outer[0]/inner[0]")), is(false));

    scenario.next();
    // The repeat is relevant but nothing inside it is.
    assertThat(scenario.refAtIndex(), is(getRef("/data/outer[0]")));

    scenario.next();
    assertThat(scenario.refAtIndex(), is(getRef("/data/outer[1]")));

    scenario.next();
    assertThat(scenario.refAtIndex(), is(getRef("/data/outer[1]/inner[0]")));

    scenario.next();
    assertThat(scenario.refAtIndex(), is(getRef("/data/outer[1]/inner[0]/q1[0]")));

    scenario.answer("/data/relevance-condition", "1");

    scenario.jumpToBeginningOfForm();
    scenario.next();
    assertThat(scenario.refAtIndex(), is(getRef("/data/outer[0]")));

    scenario.next();
    assertThat(scenario.refAtIndex(), is(getRef("/data/outer[0]/inner[0]")));
}
```

seadowg: Right this makes a lot more sense to me now. The way we were thinking about position before was hurting my brain. I think we can incorporate this in this PR.

lognaturel (approved):

PR #596, opened 2020-09-15
#506

### LINKED ISSUE #506: Space between output elements should be respected
#### Software versions 
JavaRosa v2.17 and prior.

#### Problem description
Space between output elements as in `<output value=" ../person/first_name "/> <output value=" ../person/last_name "/>` is not respected. That is, if `first_name` has value "Foo", `last_name` has value "Bar" and the two output elements are in a `label`, the `label` is displayed as "FooBar" rather than "Foo Bar".

[named-repeat.xml.zip](https://github.com/opendatakit/javarosa/files/3764296/named-repeat.xml.zip)

#### Expected behavior
Spaces should be respected.

macdude357:
Added test to demonstrate using non-breaking space to pad output.

Closes #506 

Added a unit test to demonstrate that using NBSP will achieve the intended effect


lognaturel (approved):

lognaturel:
Thanks!

PR #597, opened 2020-09-16
Bugfix/add test for sum

### LINKED ISSUE #562: Type conversion may not be as described by XPath spec
https://forum.opendatakit.org/t/how-to-compare-2-calculated-values/5588 for sample form.

It should be possible to sum fields of type string because they should automatically be coerced to numbers.

More context: https://github.com/XLSForm/xlsform.github.io/issues/71#issuecomment-224322744 and https://github.com/XLSForm/pyxform/issues/70#issuecomment-599660252

We should at least have tests to document the current behavior.

macdude357:
Closes #562 

Added tests to demonstrate the sum works properly with Strings


macdude357:
Based on my tests, it seems that sum is working properly with Strings.  Note that sum will not allow Strings with characters other than 0-9, - and .  so the user that reported the issue may have used a value like +12.0 or 12,000.01 which would get converted to 12.000.01 which is invalid as well.

PR #598, opened 2020-09-16
#555

### LINKED ISSUE #555: number(geopoint) returns empty string
<a href="https://github.com/mitchellsundt"></img></a> **Issue by [mitchellsundt](https://github.com/mitchellsundt)**
_Thursday Jul 09, 2015 at 19:39 GMT_
_Originally opened as https://github.com/opendatakit/opendatakit/issues/1006 (3 comment(s))_

---

Originally reported on Google Code with ID 1005

```
number(geopoint) should return a number. It returns an empty string,
```

Reported by `yanokwa@nafundi.com` on 2014-05-20 19:39:27

<hr>
- _Attachment: [geopoint_accuracy.xml](https://storage.googleapis.com/google-code-attachments/opendatakit/issue-1005/comment-0/geopoint_accuracy.xml)_

macdude357:
Added test to test XPathFuncExpr.toString, toNumber, etc when passing a
GeoPointData instance.

Closes #555

### DIFF
--- a/src/test/java/org/javarosa/xpath/expr/XPathFuncExprGeoPointDataTest.java
+++ b/src/test/java/org/javarosa/xpath/expr/XPathFuncExprGeoPointDataTest.java
@@ -1,0 +1,25 @@
+package org.javarosa.xpath.expr;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.javarosa.core.model.data.GeoPointData;
+import org.junit.Test;
+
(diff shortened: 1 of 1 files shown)


macdude357:
Seems like toNumber(geopoint) is working correctly (returning the accuracy value).  Even if the Geopoint is not valid, it returns 9999999.0.  I'm also a bit confused on the suggestion that GeoPointData should not implement IExprDataType but should be implemented by the data returned by GeoPointData.  I don't see how that could be since the data returned by GeoPointData are built in types.  It seems that the combination of the logic in XPathFuncExpr.toXXXXX and the GeoPointData.toXXXXX are effective.

PR #604, opened 2020-09-18
Group repeat relevance

### LINKED ISSUE #603: DAG does not compute relevance for all repeats
#### Software versions 
JavaRosa v3.0.1

#### Problem description
The form posted as part of https://github.com/getodk/collect/issues/4090 still prompts for adding `egg_photos` and `light_source` repeats even when their relevance expressions should evaluate to `false`. 

Run the form from https://github.com/getodk/collect/issues/4090, set a GPS coordinate, swipe forward.

#### Expected behavior
There should be no prompt to add repeats.

#### Other information 
[Troubleshooting notes](https://docs.google.com/document/d/1IhYsKQ5uYRxV6NEYIe7q2mYUT8XTAGX5KQ3hge1imAE/edit).

I see that on initialization, all the other repeats' relevance are computed in `evaluateTriggerable`. The triggerables for `egg_photos` and `light_source`'s relevances are not evaluated then.

lognaturel:
Closes #603

It is certainly possible to figure out the right contexts to fix #603 without separately caching repeat relevance but I don't think it's worth it. We now understand the purpose of that cache and have tests for the behavior it provides which is great. That has in turn given us insights into categories of forms we don't have enough test coverage around -- ones with shared expressions at different levels of the tree. I will be following up with some tests around that and am expecting there will be some more bugs found in that process.

This incorporates @grzesiek2010's change in #570 which tests the repeat template's relevance in the case of a static expression.

Verified in Collect with the form at https://github.com/getodk/collect/issues/4090. Added a test that demonstrates the issue that form uncovered.

I believe that all code removed by #581 to walk up the tree checking for parent relevance is redundant (taken care of by DAG updates). However, I don't feel confident enough to remove it for the `isRepeatRelevant` case and I've brought it back with a comment.

I went back and forth on the code at https://github.com/getodk/javarosa/pull/581/files#diff-1cf9266e21581c3fdea78db8d579c7ebL395 which explicitly walks the tree for non-repeat notes. I really tried to write a test that would show this is necessary but I can't do it. It still feels relatively safe to me but can bring it back if we don't want to take the risk.

The intentional change is fixing the bug described in #603. This brings back code that has been in production for a long time and that we now have extensive test coverage for so it should be low-risk. I think the highest risk decision is not to walk the tree for non-repeat nodes.

See issue.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -531,16 +531,25 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
     }
 
     public boolean isRepeatRelevant(TreeReference repeatRef) {
-        TreeElement repeatNode = mainInstance.resolveReference(repeatRef);
+        boolean relev = true;
 
-        if (repeatNode != null) {
-            return repeatNode.isRelevant();
-        } else {
-            // We are dealing with a repeat that doesn't exist yet so get
-            // relevance from the template
-            TreeElement template = mainInstance.getTemplate(repeatRef);
-            return template.isRelevant();
+        QuickTriggerable qc = dagImpl.getRelevanceForRepeat(repeatRef.genericize());
+        if (qc != null) {
+            relev = (boolean) qc.eval(mainInstance, new EvaluationContext(exprEvalContext, repeatRef));
         }
+
+        if (relev) {
+            // Must check the template in case of static relevance expressions that wouldn't be in the DAG (e.g. false()).
+            TreeElement templNode = mainInstance.getTemplate(repeatRef);
+            // Check the relevancy of the immediate parent. Unclear whether this is needed because in a normal form-filling
+            // flow the child would not be accessible if its parent is irrelevant.
+            TreeReference parentPath = templNode.getParent().getRef().genericize();
+            TreeElement parentNode = mainInstance
+                .resolveReference(parentPath.contextualize(repeatRef));
+            relev = parentNode.isRelevant() && templNode.isRelevant();
+        }
+
+        return relev;
     }
 
     public boolean canCreateRepeat(TreeReference repeatRef, FormIndex repeatIndex) {
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -90,6 +90,11 @@ public class TriggerableDag {
      */
     private final Map<TreeReference, Set<QuickTriggerable>> triggerablesPerTrigger = new HashMap<>();
 
+    /**
+     * An index to look up relevance conditions for each repeat. See buildRelevancePerRepeat.
+     */
+    private Map<TreeReference, QuickTriggerable> relevancePerRepeat = new HashMap<>();
+
     TriggerableDag(EventNotifierAccessor accessor) {
         this.accessor = accessor;
     }
@@ -149,6 +154,7 @@ public class TriggerableDag {
      */
     void finalizeTriggerables(FormInstance mainInstance, EvaluationContext ec) throws IllegalStateException {
         triggerablesDAG = buildDag(allTriggerables, getDagEdges(mainInstance, ec));
+        relevancePerRepeat = buildRelevancePerRepeat(mainInstance, triggerablesDAG);
     }
 
     /**
@@ -532,13 +538,6 @@ public class TriggerableDag {
         // which case the triggerable will be applied for every repeat instance.
         List<TreeReference> qualifiedReferences = evalContext.expandReference(contextRef);
 
-        // We want to update the repeat's template as well
-        TreeElement template = mainInstance.getTemplate(contextRef);
-        if (template != null) {
-            TreeReference templateRef = template.getRef();
-            qualifiedReferences.add(templateRef);
-        }
-
         List<EvaluationResult> evaluationResults = new ArrayList<>(0);
         for (TreeReference qualified : qualifiedReferences) {
             try {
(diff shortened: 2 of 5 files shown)


src/test/java/org/javarosa/core/model/RepeatTest.java
```diff
@@ -73,4 +73,35 @@ public void whenRepeatRelevanceIsDynamic_andNotRelevant_repeatPromptIsSkipped()
         int event = scenario.next();
         assertThat(event, is(FormEntryController.EVENT_END_OF_FORM));
     }
+
+    @Test
+    public void whenRepeatAndTopLevelNodeHaveSameRelevanceExpression_andExpressionEvaluatesToFalse_repeatPromptIsSkipped() throws Exception {
```
seadowg, line 78: This problem **is** unique to repeats right? I guess with normal nodes the problem just wouldn't matter?

src/test/java/org/javarosa/form/api/FormEntryModelTest.java
```diff
@@ -74,4 +74,51 @@ public void isIndexRelevant_respectsRelevanceOfOutermostGroup() throws IOExcepti
...
     }
+
+    /**
+     * Identical expressions in a form get collapsed to a single Triggerable and the Triggerable's context becomes
+     * its targets' highest common parent (see Triggerable.intersectContextWith). This makes evaluation in the context
+     * of repeats hard to reason about. This test shows that relevance is propagated as expected when a relevance expression
+     * is shared between a repeat and non-repeat.
+     */
+    @Test
+    public void whenRepeatAndNonRepeatShareRelevanceExpr_repeatContentRelevanceRespectsHierarchy() throws IOException {
```
seadowg, line 85: Not sure why we need this and the other test?

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -149,6 +154,7 @@ private QuickTriggerable findTriggerable(Triggerable t) {
      */
     void finalizeTriggerables(FormInstance mainInstance, EvaluationContext ec) throws IllegalStateException {
```
seadowg, line 155: I'm guessing this is called when the form is loading/starting up?

src/main/java/org/javarosa/core/model/TriggerableDag.java
```diff
@@ -645,6 +644,28 @@ void deleteRepeatInstance(FormInstance mainInstance, EvaluationContext evalConte
...
+     * by the DAG because repeat templates aren't updated by DAG recomputations. Additionally, the context intersection
+     * made in Triggerable.intersectContextWith means we may not evaluate against the correct context when the triggerable is
+     * applied.
+     */
+    private static Map<TreeReference, QuickTriggerable> buildRelevancePerRepeat(FormInstance mainInstance, Set<QuickTriggerable> triggerables) {
+        Map<TreeReference, QuickTriggerable> relevancePerRepeat = new HashMap<>();
+        for (QuickTriggerable triggerable : triggerables)
+            if (triggerable.isCondition())
+                for (TreeReference target : triggerable.getTargets())
+                    if (mainInstance.getTemplate(target) != null)
```
seadowg, line 660: If it has a template it's a repeat right? I think we need to express that in a clearer way throughout.

src/main/java/org/javarosa/core/model/condition/Triggerable.java
```diff
@@ -65,9 +65,10 @@
...
     private TreeReference contextRef;  //generic ref used to turn triggers into absolute references
 
-    // TODO Study why we really need this property. Looking at mutators, it should always equal the contextRef.
     /**
-     * The first context provided to this triggerable before reducing to the common root.
+     * The first context provided to this triggerable before reducing to the common root (see Triggerable.intersectContextWith).
+     * This is just the first bind encountered with the expression represented by this Triggerable so it's not clear how
+     * it's useful (and bind order for the same form definition affects its value).
      */
     private TreeReference originalContextRef;
```
seadowg, line 73: I don't think I understand what context is for at all

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -531,16 +531,25 @@ public void processResultOfAction(TreeReference refSetByAction, String event) {
...
-            TreeElement template = mainInstance.getTemplate(repeatRef);
-            return template.isRelevant();
+        QuickTriggerable qc = dagImpl.getRelevanceForRepeat(repeatRef.genericize());
+        if (qc != null) {
+            relev = (boolean) qc.eval(mainInstance, new EvaluationContext(exprEvalContext, repeatRef));
+        }
+
+        if (relev) {
+            // Must check the template in case of static relevance expressions that wouldn't be in the DAG (e.g. false()).
+            TreeElement templNode = mainInstance.getTemplate(repeatRef);
```
seadowg, line 543: Urgh I hate that we have to do this as well. Hate it.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -531,16 +531,25 @@ public void processResultOfAction(TreeReference refSetByAction, String event) {
...
+            // Check the relevancy of the immediate parent. Unclear whether this is needed because in a normal form-filling
+            // flow the child would not be accessible if its parent is irrelevant.
+            TreeReference parentPath = templNode.getParent().getRef().genericize();
```
seadowg, line 546: Did we need this step to pass any of the tests?

src/test/java/org/javarosa/form/api/FormEntryModelTest.java
lognaturel: You mean the other simpler one could be removed and both the general repeat prompt case and the collapsed triggers led case could be covered by this one?

src/main/java/org/javarosa/core/model/FormDef.java
lognaturel: I couldn’t express a test that failed without it and I tried. But I am feeling a little too nervous to remove it.

src/main/java/org/javarosa/core/model/TriggerableDag.java
lognaturel: Yes.

src/main/java/org/javarosa/core/model/condition/Triggerable.java
lognaturel: When you have relative expressions, you need to be able to convert them to concrete nodes. The evaluation context lets you do that. So if you have the path foo and you’re in /data/repeat[10], you’re actually referring to /data/repeat[10]/foo. 

The problem here is that the “template” concept is represented in such a way that contextualization can lead you to lose that you want to refer to the template.

src/test/java/org/javarosa/form/api/FormEntryModelTest.java
seadowg: Chatted about this and agreed to delete this test.

seadowg (approved):

src/test/java/org/javarosa/form/api/FormEntryModelTest.java
lognaturel: After looking at it some more, I remembered that this test here was an additional one I wrote to verify that the relevance was correctly updated at every level of a hierarchy. It's a bit more coupled with the implementation since it uses `isRepeatRelevant`.

I'm going to add some other tests around Triggerables with multiple targets of different contexts. When I do that, I'll figure out whether this test should go away or stay and if it stays, where it should go.

PR #610, opened 2020-10-20
Add test to demonstrate use of setvalue with literal value

### LINKED ISSUE #602: Literal strings for value attribute of the setvalue action don't work
#### Software versions 
JavaRosa v3 and most likely all prior

Add a `setvalue` action such as
`<setvalue event="odk-instance-first-load" ref="/data/my-node" value="a literal string"/>`

This can be done by creating an XLSForm with a value in the `default` column that has hyphens in it (e.g. `my-value`). `pyxform` will treat this as a dynamic default and create a `setvalue`.

#### Expected behavior
The literal string value should populate the target node when the event is fired.

#### Other information 
I think we may have introduced a special case for empty strings without realizing literal strings didn't work.

macdude357:
Added test to demonstrate how setval works with literal strings.

Closes #602 

Added unit test

Just a test change to demonstrate existing behavior

No code changes
No
Maybe update the docs for the tools that generate the forms to explain how to use setvalue with a literal

### DIFF
--- a/src/test/java/org/javarosa/xform/parse/XFormParserTest.java
+++ b/src/test/java/org/javarosa/xform/parse/XFormParserTest.java
@@ -424,6 +424,14 @@ public class XFormParserTest {
         assertThat(g3Element.getBind(), is(g3AbsRef));
     }
 
+    @Test
+    public void testSetValueWithStrings() throws IOException {
+        
+        Scenario scenario = Scenario.init("default_test.xml");
+        assertEquals("string-value", scenario.getAnswerNode("/data/string_val").getValue().getValue().toString());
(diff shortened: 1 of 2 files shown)


lognaturel (approved):
Thanks! We now favor the form-building DSL (defined by [XFormsElement](https://github.com/getodk/javarosa/blob/0f8d69edf49bcac6ca6bd1ecc1ea2c51a8c4b1c4/src/test/java/org/javarosa/core/util/XFormsElement.java)) when adding new tests so that the form definition and assertions are in the same place. Naturally, https://github.com/getodk/javarosa/issues/571 to actually document this has been languishing! Since all the other tests in this file have separate forms, I think it's fine.

PR #612, opened 2020-10-21
Re-evaluate calculates that may have multiple dependencies

### LINKED ISSUE #611: On repeat creation, some calculates are computed before all of the calculates they reference
#### Software versions 
JavaRosa v3

#### Problem description
Expressions in repeats that reference a calculate which is triggered by repeat addition AND some other calculate do not get computed correctly on repeat addition. This is because the whole cascade based on repeat addition is computed initially but some of the dependencies may not already have been triggered.

This test currently fails because `/data/repeat/other` doesn't yet have a value when `/data/repeat/concatenated` is evaluated:
```
    @Test
    public void addingRepeat_updatesInnerCalculations_withMultipleDependencies() throws IOException {
        Scenario scenario = Scenario.init("Repeat cascading calc", html(
            head(
                title("Repeat cascading calc"),
                model(
                    mainInstance(t("data id=\"repeat-calcs\"",
                        t("repeat",
                            t("position"),
                            t("position_2"),
                            t("other"),
                            t("concatenated")
                        ))),
                    // position(..) means the full cascade is evaulated as part of triggerTriggerables
                    bind("/data/repeat/position").calculate("position(..)"),
                    bind("/data/repeat/position_2").calculate("/data/repeat/position * 2"),
                    bind("/data/repeat/other").calculate("2 * 2"),
                    // concat needs to be evaluated after other has a value
                    bind("/data/repeat/concatenated").calculate("concat(../position_2, '-', ../other)")),
                body(
                    repeat("/data/repeat",
                        input("/data/repeat/concatenated"))
                ))));

        scenario.next();
        scenario.next();
        assertThat(scenario.answerOf("/data/repeat[" + 0 + "]/concatenated"), CoreMatchers.is(stringAnswer("2-4")));

        scenario.next();
        scenario.createNewRepeat();

        scenario.next();
        assertThat(scenario.answerOf("/data/repeat[" + 1 + "]/concatenated"), CoreMatchers.is(stringAnswer("4-4")));
    }
```

#### Expected behavior
On repeat addition, inner calculations should be computed in DAG order.

#### Other information 
The change that led to this is that we now pass in the triggered calculations from `triggerTriggerables` as already triggered to the call on `initializeTriggerables` whereas we didn't previously:
https://github.com/getodk/javarosa/blob/v2.17.2/src/main/java/org/javarosa/core/model/Safe2014DagImpl.java#L102

lognaturel:
Closes #611

Added automated tests, verified with the form at https://forum.getodk.org/t/repeats-and-calculations-inside-nested-repeats-behaving-differently-in-odk-1-27-3-vs-1-28-2/30694. I debugged the DAG creation process to make sure everything was in the expected order. I debugged the test with v2.17 and the change with the triggered set and confirmed that the behavior was the same.

Once I saw that this was a change from v2.17, I did not consider any other approaches. 

Since this is bringing back old behavior, the risk is low.

See test.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/TriggerableDag.java
+++ b/src/main/java/org/javarosa/core/model/TriggerableDag.java
@@ -574,12 +574,13 @@ public class TriggerableDag {
         publishSummary("Created (phase 1)", createdRef, qtSet1);
 
         // initialize conditions for the node (and sub-nodes)
-        Set<QuickTriggerable> qtSet2 = initializeTriggerables(mainInstance, evalContext, createdRef, qtSet1);
+        Set<QuickTriggerable> qtSet2 = initializeTriggerables(mainInstance, evalContext, createdRef, new HashSet<>());
         publishSummary("Created (phase 2)", createdRef, qtSet2);
 
         Set<QuickTriggerable> alreadyEvaluated = new HashSet<>(qtSet1);
         alreadyEvaluated.addAll(qtSet2);
 
+        // TODO: add a test that fails without this or remove (all tests on v2.17 pre-DAG simplification pass)
         evaluateChildrenTriggerables(mainInstance, evalContext, createdElement, true, alreadyEvaluated);
     }
 
--- a/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
+++ b/src/test/java/org/javarosa/core/model/TriggerableDagTest.java
@@ -752,6 +752,78 @@ public class TriggerableDagTest {
     //endregion
 
     //region Adding or deleting repeats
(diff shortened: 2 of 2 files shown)


lognaturel:
@macdude357 I'd mentioned tagging you in the regression fix PR when we talked in Slack. Here it is but I now realize it needs a lot of context so up to you whether it makes sense to look at! @seadowg and I will do a quick interactive review anyway. The big thing is to make sure my test adequately captures the case and that there aren't any similar issues lurking nearby.

seadowg (approved):

PR #615, opened 2020-12-09
Support self-closing setvalue without value

lognaturel:
#564 added support for `setvalue` with an empty literal value or an empty `String` as the attribute value for `value`. However, it didn't add support for a self-closing `setvalue` tag without a `value` attribute which is what `pyxform` actually generates in the case of an empty `calculation`:
`<setvalue event="xforms-value-changed" ref="/data/malaria_age"/>`

Added test and verified that the generated XML matches the one generated by `pyxform`.

This is such a small change that I don't think there really are alternatives. It gets the desired behavior -- if there isn't an explicit value for a `value` attribute or as a child, then we clear the target.

This loosens the requirements for `setvalue`. It could possibly allow more form design mistakes through but I think that risk is acceptable.

See test.

https://github.com/getodk/docs/issues/1287

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -786,11 +786,12 @@ public class XFormParser implements IXFormParserFunctions {
 
         registerActionTarget(treeref);
         if (valueExpression == null) {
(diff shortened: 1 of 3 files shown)


seadowg (approved):

seadowg:
Looks good!

PR #619, opened 2021-01-29
Cache and reuse dynamic choice lists

lognaturel:
Speeds up displaying previously-filtered choice lists.
Closes getodk/collect#4363

Added and ran tests, tried it in Collect.

We've talked about many possible performance improvements in this area. Most would only work in some cases. Adding caching works for any choice filter expression and gives us O(1) performance in many cases. Initially computing a choice list is O(N) and can still be slow but once it's computed, we don't have to do it again. I think this is a relatively low-risk, high impact change that we want to do no matter what other improvements we make later.

The intent is just to speed up displaying filtered choice lists that were previously filtered and didn't change. There's risk around all things dynamic choice lists.

I initially thought that this would increase memory usage but I believe that it makes no difference because a choice list was always saved in `ItemsetBinding` previously, it just was immediately discarded.

See tests.

No.


src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,323 @@
...
+            )
+        );
+    }
+    //endregion
+
+    //region Select in repeat
+    // When a dynamic select is in a repeat, the itemsets for all repeat instances are represented by the same ItemsetBinding.
+    @Test
+    public void selectInRepeat_withPredicate_evaluatesChoiceListForEachRepeatInstance() throws Exception {
+        Scenario scenario = Scenario.init("Select from repeat", html(
```
dcbriccetti, line 147: Did you mean for the form name to be `Select in repeat`?

lognaturel: Ah yes, indeed!

src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -56,54 +70,198 @@
...
+        return null;
     }
 
     public void localeChanged(String locale, Localizer localizer) {
-        if (choices != null) {
-            for (int i = 0; i < choices.size(); i++) {
-                choices.get(i).localeChanged(locale, localizer);
+        if (latestFilteredChoiceList != null) {
+            for (int i = 0; i < latestFilteredChoiceList.size(); i++) {
+                latestFilteredChoiceList.get(i).localeChanged(locale, localizer);
```
dcbriccetti, line 264: New-style `for` might be easier to read.
```
for (SelectChoice choice : latestFilteredChoiceList) {
    choice.localeChanged(locale, localizer);
}
```

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,323 @@
...
+                        item("bbb", "BB"))),
+                body(
+                    input("/data/filter"),
+                    select1Dynamic("/data/select", "instance('choices')/root/item[starts-with(value,/data/filter)]")
+                )
+            )));
+
+        scenario.answer("/data/filter", "a");
+        List<SelectChoice> choices = scenario.choicesOf("/data/select");
+        assertThat(scenario.choicesOf("/data/select"), sameInstance(choices));
```
dcbriccetti, line 250: This actually works! Nice.

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+    //region Multi-language
+    @Test
+    public void multilanguage() throws Exception {
+        Scenario scenario = Scenario.init("Select from repeat", html(
+            head(
+                title("Multilingual dynamic select"),
+                model(
+                    t("itext",
+                        t("translation lang=\"fr\"",
```
dcbriccetti, line 188: Any idea why we‘re using `\"` instead of `'`? Both forms seem to generate double-quoted attribute values. I find code with escaped double quotes harder to read.

lognaturel: I didn't realize that'd do the right thing. 👍 will update.

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
lognaturel: Yeah, seems like a reasonable way to test for the behavior we want, right?!

src/main/java/org/javarosa/core/model/ItemsetBinding.java
lognaturel: Yes, can update.

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
dcbriccetti: image

Here’s where the attribute value delimiters are ignored, in XFormsElement. Later they are replaced with double quotes.

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,323 @@
...
+                            t("text id=\"choices-0\"",
+                                t("value", "A (fr)")),
+                            t("text id=\"choices-1\"",
+                                t("value", "B (fr)")),
+                            t("text id=\"choices-2\"",
+                                t("value", "C (fr)"))
+                        ),
+                        t("translation lang=\"en\"",
```
dcbriccetti, line 196: On my phone, filling out this form, the French versions of the strings are presented to me, unless I modify the form, switching the order of the two `translation` elements. Is this expected?

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,323 @@
...
+        scenario.answer("/data/repeat[1]/filter", "bb");
+        assertThat(scenario.choicesOf("/data/repeat[0]/select").size(), is(2));
+        assertThat(scenario.choicesOf("/data/repeat[1]/select").size(), is(1));
+    }
+    //endregion
+
+    //region Multi-language
+    @Test
+    public void multilanguage() throws Exception {
+        Scenario scenario = Scenario.init("Select from repeat", html(
```
dcbriccetti, line 183: Might want to check all the form names.

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,323 @@
...
+                            t("repeat",
+                                t("select"),
+                                t("filter")))),
+
+                    instance("choices",
+                        item("a", "A"),
+                        item("aa", "AA"),
+                        item("b", "B"),
+                        item("bb", "BB"))),
+                body(
```
dcbriccetti, line 162: Is it OK for the body to be inside the head?

dcbriccetti: image

If it is invalid, then perhaps we’d modify the DSL to disallow that at compile time.

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
lognaturel: Yes, see [the spec](https://getodk.github.io/xforms-spec/#languages): "A default=”” attribute can be added to a <translation> to make it the default language, otherwise the first listed is used as the default."

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
lognaturel: Huh. I'm shocked that works. It's pretty twisted.

Looks like we introduced that badness in a test at some point and then kept copying it everywhere. Good catch. Let me know if there's a less verbose way to introduce `Head/BodyXFormsElement`.

dcbriccetti: Maybe this?

image

```
    class HeadXFormsElement extends TagXFormsElement {
        public HeadXFormsElement(XFormsElement[] children) {
            super("h:head", emptyMap(), Arrays.asList(children));
        }
    }

    class BodyXFormsElement extends TagXFormsElement {
        public BodyXFormsElement(XFormsElement[] children) {
            super("h:body", emptyMap(), Arrays.asList(children));
        }
    }
```

lognaturel: 🧐 🧐 🧐 There was a reason I thought extending `TagXFormsElement` wasn't what I wanted at the time but now I have no idea what it might have been.

src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -56,54 +70,198 @@
...
+        return latestFilteredChoiceList;
+    }
+
+    /**
+     * Returns a map:
+     *  - keys: the references that are triggers for the nodeset expression
+     *  - values: current values at those references
+     *
+     * Returns null if the nodeset expression has any triggers that are unbounded references because there's no single
+     * value we could track in that case.
```
dcbriccetti, line 157: Something’s bothering me about using null to indicate this. But maybe I don’t have an improvement using Java 7. I assume returning an empty map would have a different meaning. A checked exception, maybe?

dcbriccetti: I’d like to better understand how this is tested. Here are the call stacks where null is returned.
```
Called with /data/select[0] from selectFromRepeat_whenRepeatAdded_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatAdded_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatAdded_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatAdded_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatRemoved_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatRemoved_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatRemoved_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatRemoved_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatChanged_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatChanged_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatChanged_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatChanged_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatChanged_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_whenRepeatChanged_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectWithRepeatAsTrigger_recomputesChoiceListAtEveryRequest -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectWithRepeatAsTrigger_recomputesChoiceListAtEveryRequest -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectWithRepeatAsTrigger_recomputesChoiceListAtEveryRequest -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectWithRepeatAsTrigger_recomputesChoiceListAtEveryRequest -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectWithRepeatAsTrigger_recomputesChoiceListAtEveryRequest -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectWithRepeatAsTrigger_recomputesChoiceListAtEveryRequest -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_withPredicate_whenPredicateTriggerChanges_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_withPredicate_whenPredicateTriggerChanges_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_withPredicate_whenPredicateTriggerChanges_updatesChoices -> Scenario.choicesOf -> FormEntryPrompt.getAnswerValue -> FormEntryPrompt.getSelectChoices -> ItemsetBinding.getChoices
Called with /data/select[0] from selectFromRepeat_withPredicate_whenPredicateTriggerChanges_updatesChoices -> Scenario.choicesOf -> ItemsetBinding.getChoices
```

dcbriccetti: We discussed and have an improvement.

lognaturel:
I ran `DynamicSelectUpdateTest`, `SelectOneChoiceFilterTest`, `SelectMultipleChoiceFilterTest` and `XPathFuncExprRandomizeTest` with coverage and am happy with the line coverage (with the understanding that doesn't mean all concepts are covered). The lines that lack coverage are:
- error handling for unlikely corner cases
- related to "copy mode" which I know nothing about and did my best to just not touch
- have TODO comments because I suspect the code does nothing at all

I think the last thing pending is for @dcbriccetti to flag areas he's not confident about so we can compare notes and see whether there are additional areas of risk I might have missed.

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,398 @@
...
+
+                    instance("choices",
+                        item("a", "A"),
+                        item("aa", "AA"),
+                        item("b", "B"),
+                        item("bb", "BB")))),
+            body(
+                repeat("/data/repeat",
+                    select1Dynamic("/data/repeat/select", "instance('choices')/root/item[starts-with(value,current()/../filter)]"),
+                    input("filter"))
```
dcbriccetti, line 306: Should the filter input precede the select?

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,398 @@
...
+        assertThat(choices.size(), is(2));
+        // Because of the repeat trigger in the count expression, choices should be recomputed every time they're requested
+        assertThat(scenario.choicesOf("/data/select"), not(sameInstance(choices)));
+    }
+
+    //region Caching for selects in repeat
+    // When a dynamic select is in a repeat, the itemsets for all repeat instances are represented by the same ItemsetBinding.
+    @Test
+    public void selectInRepeat_withRefToRepeatChildInPredicate_evaluatesChoiceListForEachRepeatInstance() throws Exception {
+        Scenario scenario = Scenario.init("Select in repeat", html(
```
dcbriccetti, line 288: Perhaps this is expected, but when I preview this form in ODK Central, I get 
    TypeError: e is null
(This isn’t blocking me.)

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,398 @@
...
+/**
+ * Integration tests to verify that the choice lists for "dynamic selects" (selects with itemsets rather than inline items) are updated when
+ * dependent values change.
+ *
+ * See also:
+ * - {@see SelectOneChoiceFilterTest}
+ * - {@see SelectMultipleChoiceFilterTest} for coverage of dynamic select multiples
+ * - {@see XPathFuncExprRandomizeTest} for coverage of choice list updates when randomization is specified
+ */
+public class DynamicSelectUpdateTest {
```
dcbriccetti, line 53: image
Should we have a test case where the cache is not used because the random seed differs? (The picture shows code coverage.)

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,398 @@
...
+                    input("filter"))
+            )));
+
+        scenario.answer("/data/repeat[0]/filter", "a");
+        scenario.answer("/data/repeat[1]/filter", "a");
+        List<SelectChoice> repeat0Choices = scenario.choicesOf("/data/repeat[0]/select");
+        List<SelectChoice> repeat1Choices = scenario.choicesOf("/data/repeat[1]/select");
+
+        // The trigger keys are /data/repeat[0]/filter and /data/repeat[1]/filter which means no caching between them
+        assertThat(repeat0Choices, not(sameInstance(repeat1Choices)));
```
dcbriccetti, line 315: Yes. New logging confirms this as the reason the cache was not used:

```
14:26:45.534 [Test worker] DEBUG org.javarosa.core.model.ItemsetBinding - Not using cached value because:
14:26:45.534 [Test worker] DEBUG org.javarosa.core.model.ItemsetBinding -   triggers differ: {/data/repeat[2]/filter=StringData{s='a'}}, {/data/repeat[1]/filter=StringData{s='a'}}
```

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,398 @@
...
+        assertThat(choices, containsInAnyOrder(
+            choice("aa", "A"),
+            choice("aaa", "AA")));
+
+        scenario.answer("/data/filter", "aa");
+        assertThat(choices, containsInAnyOrder(
+            choice("aa", "A"),
+            choice("aaa", "AA")));
+        // Even though the list happens to be unchanged, it should have been recomputed because the trigger value changed
+        assertThat(scenario.choicesOf("/data/select"), not(sameInstance(choices)));
```
dcbriccetti, line 247: Confirmed:

```
15:17:56.835 [Test worker] DEBUG org.javarosa.core.model.ItemsetBinding -   triggers differ: {/data/filter=StringData{s='aa'}}, {/data/filter=StringData{s='a'}}
```

src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -22,22 +31,27 @@
...
-import org.slf4j.LoggerFactory;
 
 public class ItemsetBinding implements Externalizable, Localizable {
-    private static final Logger logger = LoggerFactory.getLogger(ItemsetBinding.class);
+    // Temporarily cached filtered list (not serialized)
+    private List<SelectChoice> latestFilteredChoiceList;
+
+    // Values needed to determine whether the cached list should be expired (not serialized)
+    private Map<TreeReference, IAnswerData> latestTriggerValues;
+    private Long latestRandomizeSeed;
```
dcbriccetti, line 48: More than once I’ve confused `latestTriggerValues` with `currentTriggerValues`. Perhaps more distinct names would help? Maybe `previous…` or `cached…` instead of `latest…`?

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
lognaturel: It doesn't matter for the tests but I've changed it everywhere so the forms make sense to a human. 😬

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
lognaturel: It looks like this is not something Enketo (web form engine) supports but it's esoteric and it works in Collect so I'm not worried about it.

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
lognaturel: Did you see that the tests in `XPathFuncExprRandomizeTest` do exercise that case? I ended up putting a reference to that test class in the header comment for `DynamicSelectUpdateTest`. Do you think it's important to have explicit tests in both places?

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
```diff
@@ -0,0 +1,398 @@
...
+        List<SelectChoice> repeat1Choices = scenario.choicesOf("/data/repeat[1]/select");
+        assertThat(repeat0Choices, sameInstance(repeat1Choices));
+
+        scenario.answer("/data/filter", "bb");
+        assertThat(scenario.choicesOf("/data/repeat[0]/select").size(), is(1));
+    }
+
+    @Test
+    public void selectInNestedRepeat_withPredicateWithOuterRef_returnsCachedChoiceListForSameOuterRepeatInstance() throws Exception {
+        Scenario scenario = Scenario.init("Select in repeat", html(
```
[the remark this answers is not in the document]
lognaturel: Looks like the same Enketo bug found at https://github.com/getodk/javarosa/pull/619#discussion_r591877575. I think it's ok to let it go and just be glad JavaRosa/Collect now has coverage.

src/main/java/org/javarosa/core/model/ItemsetBinding.java
lognaturel: Thank you, I like `cached`!

src/test/java/org/javarosa/core/model/DynamicSelectUpdateTest.java
dcbriccetti: Oh, sorry, I missed that!

dcbriccetti (approved):
Looks good. Thanks for addressing all my concerns.

lognaturel:
Thanks so much for all the great feedback and checks!

src/main/java/org/javarosa/form/api/FormEntryPrompt.java
```diff
@@ -318,7 +292,8 @@ public String getHelpText() {
 
 
     /**
-     * Attempts to return the specified Item (from a select or 1select) text.
+     * Attempts to return the specified item text from a STATIC select or select1. Does NOT
+     * support dynamic selects ({@see ItemsetBinding}).
```
lognaturel, line 296: This comment suggests I thought this was not applicable to dynamic selects. However, it's now crashing for dynamic selects so something is not right: https://github.com/getodk/javarosa/pull/625#issuecomment-805488046

lognaturel: Ah. There's Collect code that directly calls `getSelectItemText` regardless of whether the select is static or dynamic. It seems that used to work and doesn't anymore.

src/main/java/org/javarosa/form/api/FormEntryPrompt.java
```diff
@@ -95,63 +94,52 @@ public IAnswerData getAnswerValue() {
...
+                        //will no longer be an option this go around
+                        if (choice != null) {
+                            selection.add(choice.selection());
                         }
                     }
-                    //if it's a dynamic question, then there's a good choice what they selected last time
-                    //will no longer be an option this go around
-                    if(choice != null)
-                    {
-                        selection.add(choice.selection());
```
lognaturel, line 142: I think this used to ensure there was always a `SelectionChoice` for the `Selection`.

PR #626, opened 2021-03-26
Bind select values to SelectChoice(s) and implement equals for IAnswerData subclasses

lognaturel:
Follow up to #619 which fixes two issues that were discovered when using in Collect. I'm going to get this to QA right away so we can see if there's anything else strange like this we might have missed.

Added and ran tests, tried in Collect with form [here](https://drive.google.com/drive/u/0/folders/1QwX7E-u19jzXvzwkOpj2Y4X-xb9ihd1f).

For binding the select values, it makes the most sense to me to do when we're possibly updating the answer data anyway. I believe that is also the only way to avoid another loop through all the choices.

This replaces [code](https://github.com/getodk/javarosa/pull/619/files#diff-626065115f07d638a026633b80d89eb37f3ec2f769c6d4f58ab3376993ea1056L142) in `getAnswerValue` that used to do this binding. It was really unclear that this was meaningful behavior and I think the new structure makes it a lot clearer. Plus, now there's a test.

When it comes to implementing `equals` for `IAnswerData` subclasses, my first instinct was to introduce an abstract class that defines `equals` using `getValue()`. However, there's so much weirdness with serialization that this scared me. I went with individual implementations and I only did the types that I think are likely to be involved in filter expressions on selects. 

There's a risk around `IAnswerData` comparison. It feels very safe to me to consider two `IAnswerData` with the same underlying values as equal but I've been surprised enough times that it makes me slightly nervous. That's why I didn't do it for more types than I needed to.

I think there's less risk around the select choice binding. Still, there could be some unexpected side effects that would affect all forms with dynamic selects (internal or from external files).

See tests

### DIFF
--- a/src/main/java/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/main/java/org/javarosa/core/model/ItemsetBinding.java
@@ -17,6 +17,7 @@ import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IConditionExpr;
 import org.javarosa.core.model.data.IAnswerData;
 import org.javarosa.core.model.data.MultipleItemsData;
+import org.javarosa.core.model.data.SelectOneData;
 import org.javarosa.core.model.data.StringData;
 import org.javarosa.core.model.data.helper.Selection;
 import org.javarosa.core.model.instance.DataInstance;
@@ -114,14 +115,14 @@ public class ItemsetBinding implements Externalizable, Localizable {
             throw new XPathException("Could not find references depended on by" + nodesetRef.getInstanceName());
         }
 
-        Map<String, Boolean> currentAnswersInNewChoices = initializeCurrentAnswerMap(formDef, curQRef);
+        Map<String, SelectChoice> currentAnswersInNewChoices = initializeCurrentAnswerMap(formDef, curQRef);
 
         List<SelectChoice> choices = new ArrayList<>();
         for (int i = 0; i < filteredItemReferences.size(); i++) {
             SelectChoice choice = getChoiceForTreeReference(formDef, formInstance, i, filteredItemReferences.get(i));
             choices.add(choice);
             if (currentAnswersInNewChoices != null && currentAnswersInNewChoices.containsKey(choice.getValue())) {
-                currentAnswersInNewChoices.put(choice.getValue(), true);
+                currentAnswersInNewChoices.put(choice.getValue(), choice);
             }
         }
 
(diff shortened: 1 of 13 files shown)


src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -200,53 +201,60 @@ private SelectChoice getChoiceForTreeReference(FormDef formDef, DataInstance for
...
      */
-    private static MultipleItemsData getFilteredSelections(MultipleItemsData selections, Map<String, Boolean> shouldKeepSelection) {
+    private static MultipleItemsData getFilteredAndBoundSelections(MultipleItemsData selections, Map<String, SelectChoice> selectChoicesToKeep) {
         List<Selection> newSelections = new ArrayList<>();
         for (Selection oldSelection : (List<Selection>) selections.getValue()) {
             String key = oldSelection.choice != null ? oldSelection.choice.getValue() : oldSelection.xmlValue;
-            if (shouldKeepSelection.get(key)) {
-                newSelections.add(oldSelection);
+            if (selectChoicesToKeep.get(key) != null) {
+                newSelections.add(selectChoicesToKeep.get(key).selection());
```
dcbriccetti, line 257: May I suggest
```
if (selectChoicesToKeep.containsKey(key)) {
    newSelections.add(selectChoicesToKeep.get(key).selection());
}
```

or

```
SelectChoice choice = selectChoicesToKeep.get(key);
if (choice != null) {
    newSelections.add(choice.selection());
}
```

src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -200,53 +201,60 @@ private SelectChoice getChoiceForTreeReference(FormDef formDef, DataInstance for
...
     }
 
     /**
      * @param selections          an answer to a multiple selection question
-     * @param shouldKeepSelection maps each value that could be in @{code selections} to a boolean representing whether
-     *                            or not it should be kept
-     * @return a copy of {@code selections} without the values that were mapped to false in {@code shouldKeepSelection}
+     * @param selectChoicesToKeep maps each value that could be in @{code selections} to a SelectChoice if it should be bound
+     *                            or null if it should be removed.
+     * @return a copy of {@code selections} without the values that were mapped to false in {@code selectChoicesToKeep} and
```
dcbriccetti, line 249: Should the comment still say “mapped to false”?

src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -200,53 +201,60 @@ private SelectChoice getChoiceForTreeReference(FormDef formDef, DataInstance for
...
 
-    // Build a map with keys for each value in the current answer. This will allow us to remove answers that are no
-    // longer available for selection because of an updated filter.
-    private Map<String, Boolean> initializeCurrentAnswerMap(FormDef formDef, TreeReference curQRef) {
-        Map<String, Boolean> currentAnswersInNewChoices = null;
+    /**
+     * Build a map with keys for each value in the current answer. This will allow us to remove answers that are no
+     * longer available for selection because of an updated filter and to bind selection(s) in the IAnswerData to Selection
+     * objects. The latter is necessary to get label text for the answers.
+     */
```
dcbriccetti, line 208: I like how we’re slowly migrating these non-doc-comments to real, first class doc comments. I know you thought of me when you did this. :-)

src/main/java/org/javarosa/core/model/ItemsetBinding.java
lognaturel: I think the first would crash without other changes because we are keeping keys that shouldn't exist anymore and mapping them to null. But maybe we could remove the keys instead -- will look at that.

src/main/java/org/javarosa/core/model/ItemsetBinding.java
lognaturel: What a pleasant way to say I made a mistake. 😂

src/main/java/org/javarosa/core/model/ItemsetBinding.java
lognaturel: Yes, yes I did. 😊

src/main/java/org/javarosa/core/model/ItemsetBinding.java
lognaturel: I explored keys instead of having them map to null and it's not practical. We currently initialize the map keys based on an old answer which could contain unavailable choices. All values are initially null. We then iterate over all the currently-available choices and set values for any choices that are in the old answer map. Removing keys mapped to null would have to be a separate step and it doesn't seem like it adds clarity. I went with your second suggestion.

While exploring this, I noticed some of my naming didn't make as much sense with the new mapping to `SelectChoice` objects (replacing booleans). I tried to improve that and add a few more comments where I had to do some extra thinking to figure out what was going on.

lognaturel:
This has now been verified by QA at https://github.com/getodk/collect/pull/4490. I'd appreciate a finalized review to see if you catch anything else, @dcbriccetti, but I'm going to merge so we can get a snapshot into Collect.

PR #628, opened 2021-04-05
616 CSV parsing improvements-WIP

### LINKED ISSUE #616: Support quotes, escaped commas, and other CSV formatting in external CSV files
#### Software versions 
All

#### Problem description
CSVs are split on commas with no accounting for standard CSV formatting.

[select-one-file.zip](https://github.com/getodk/javarosa/files/5700809/select-one-file.zip)

#### Expected behavior
Standard CSV formatting is handled.

dcbriccetti:
Closes #616

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,7 @@ dependencies {
     // Be sure to update dependencies in pom.xml to match
     implementation 'joda-time:joda-time:2.10.10'
     implementation 'org.slf4j:slf4j-api:1.7.31'
+    implementation 'org.apache.commons:commons-csv:1.6'
     compileOnly 'net.sf.kxml:kxml2:2.3.0'
     testImplementation 'ch.qos.logback:logback-classic:1.2.3'
     testImplementation 'junit:junit:4.13.2'
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,11 @@
        <scope>compile</scope>
     </dependency>
     <dependency>
+       <groupId>org.apache.commons</groupId>
+       <artifactId>commons-csv</artifactId>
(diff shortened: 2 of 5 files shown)


lognaturel:
Thanks for thinking to use the lib. That is much better than rolling our own.

Unfortunately I don't remember much about how the Android API level checker works! It looks more like a mechanical issue getting the lib recognized than an actual failure which is good.

lognaturel:
@dcbriccetti This is starting to feel more pressing. Would you have a chance to finish soon or do you prefer to hand off?

dcbriccetti:
> @dcbriccetti This is starting to feel more pressing. Would you have a chance to finish soon or do you prefer to hand off?

Apologies for the delay. Resuming now.

lognaturel:
Thank you! I did recently update the Android API checker. If it still fails after a rebase on master, I can look into it.

dcbriccetti:
Hmm, the failing tests all call `configureReferenceManagerIncorrectly`, for what that’s worth. So far I don’t see that the failures are related to this change.

lognaturel:
Thanks for doing most of the legwork here.

The test failures are because there was a `FileNotFoundException` expected that is now a `FileSystemException`. It's related to the change at https://github.com/getodk/javarosa/commit/847865c78512af9b65211ea332d23e17c119fa07. I'll close this and reopen a PR off my branch.

lognaturel:
Unfortunately, this would require API 26. We have a min API level of 21. Maybe we can use an older version of `commons-csv`, I'm not sure.

lognaturel:
v1.4 works. New PR at #644.

PR #635, opened 2021-06-24
Use placeholder secondary instance when file not found

### LINKED ISSUE #444: Fail gracefully when an external secondary instance is not found
#### Problem description
Currently, a missing external secondary instance results in failure to parse a form. This is a problem because clients like Validate (https://github.com/opendatakit/validate/pull/66) or Aggregate don't need to have those external instances to do what they need to do. In fact, even form-filling clients like Collect may not need those files if, for example, they are used in optional questions. 

Try to parse a form such as [nigeria_wards_external.xml.txt](https://github.com/opendatakit/javarosa/files/3263461/nigeria_wards_external.xml.txt) that refers to external secondary instances. An exception will be thrown from `verifyItemsetBindings`. Something will also need to be done in `populateDynamicChoices` for clients that do something with the choices.

#### Expected behavior
No exception is thrown. Clients have a way of detecting that the instance files were unavailable and can choose what to do.

lognaturel:
Closes #444 

Note: I had included another piece in the issue which is letting the client explicitly know that an external file couldn't be found. It feels less important now that I've tried the various cases in Collect so I'm inclined to close entirely for now.

I have confirmed that this allows us to remove the `ReferenceManager` config done in https://github.com/getodk/collect/pull/3523. This means only launching a form for filling in the foreground would require the correct `ReferenceManager` config and addresses https://github.com/getodk/collect/issues/4634

@seadowg and I discussed possibly using a new class to identify an unresolved external secondary instance. I got to a mostly-working implementation with an `UnresolvedExternalDataInstance` but backed out of it because all external instances would then be serialized then deserialized as `UnresolvedExternalDataInstance` forever in Collect, even if the file did actually exist and subsequently gets resolved. That seemed too strange. I tried to improve things a bit from what I had before by introducing an explicit placeholder node and using that to determine whether a placeholder was used to avoid validation. This whole line of exploration led me to add several tests around de/serialization.

I did the Gradle wrapper upgrade so I could import the module into Collect. While there I did some dependency updates. These seem low-risk to me so it felt like good housekeeping.

This can be tried in Collect from https://github.com/getodk/collect/pull/4643


src/main/java/org/javarosa/core/model/data/helper/Selection.java
```diff
@@ -120,7 +120,7 @@ public String getValue () {
         if (xmlValue != null && xmlValue.length() > 0) {
             return xmlValue;
         } else {
-            throw new RuntimeException("don't know xml value! perhaps selection was stored as index only and has not yet been linked up to a formdef?");
+            throw new RuntimeException("This choice is not valid. If you are using an external file, please make sure it has the expected columns (usually name and label).");
```
lognaturel, line 123: This is going to be extremely rare. The only case I can think of it happening in is https://github.com/getodk/javarosa/pull/635/commits/bc2a8f26efefda828f9774f59f6bd92384e02116#diff-99f7f5318b0381156d5214c96eb7c81a107e2b9e73565eabcc606a96c7246b23R203

src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -38,27 +44,36 @@ private ExternalDataInstance(TreeElement root, String instanceId, String path) {
...
      */
     public static ExternalDataInstance build(String instanceSrc, String instanceId)
-        throws IOException, UnfullfilledRequirementsException, XmlPullParserException, InvalidStructureException, InvalidReferenceException {
-        TreeElement root = parseExternalInstance(instanceSrc, instanceId);
+        throws IOException, UnfullfilledRequirementsException, XmlPullParserException, InvalidStructureException {
+        TreeElement root;
+        try {
+            root = parseExternalInstance(instanceSrc, instanceId);
+        } catch (FileNotFoundException | InvalidReferenceException e) {
+            logger.info("External instance not found, falling back to placeholder");
```
lognaturel, line 67: I decided to leave this in because the other logging around form load is often helpful from Collect.

seadowg (approved):

PR #646, opened 2021-09-30
Update the model even when a cached choice list can be used

### LINKED ISSUE #642: Attempt to invoke virtual method ‘java.lang.String.org.javarosa.core.model.SelectChoice.getTextID()’ on a null object reference

lognaturel:
Closes #642, replaces #643

#643 re-introduced performance issues. See https://github.com/getodk/javarosa/pull/643#issuecomment-923601362 for a description of when performance would matter.

This solution is similar to https://github.com/getodk/javarosa/commit/7d483f14153b5bb9b26e20b7c7fdc3e4f2093bc1 but with a narrower slice of work when cached choices are returned.

Added and ran a test. Verified that all other tests pass, in particular those that ensure performance in cases with repeats (the ones that failed in #643).

Conceptually, there are two things we need to do for any select: 
- make sure that the answer doesn't contain choices that are no longer available
- make sure the answer is bound to the corresponding choice objects so that we can display the selected choice using its label
We don't want to do this on e.g. form load because it would mean having to iterate through all of the choices which we may never need to do. Doing this binding as late as possible is good for performance.

Semantically it feels like making sure the answer is synced with the choice list should happen any time the answer is requested. However, I don't think that's practical because the answers are represented as `SelectOneData` and `SelectMultipleData` objects which have no connection to the form definition or the currently filtered list of choices. I can't think of a way to kick off that syncing. What we've been doing which is updating the answer when the choice list is requested seems like the next best thing.

@grzesiek2010 says at https://github.com/getodk/javarosa/pull/643#issuecomment-912543031

> it's kind of strange that a method called getChoices() not only returns a list of choices but also performs such an additional operations.

True, but caching is inherently stateful. We don't want clients of `getChoices` to have to keep track of caching, we want all the right updates to be triggered by choices being requested. We could do some kind of hook or listener but fundamentally I think the result would be the same and it's not worth a huge refactor at this time.

I briefly explored doing the update when the answer label is requested (currently in `getSelectItemText`). But that's not quite right, there are cases such as in validation where the choice list would be requested but the answer label would not.

The intent is to narrowly fix the issue with editing forms with filtered selects in repeats. This feels like a quite safe change. It does mean that select answers are now updated any time a choice list is requested so there's risk there. We have very high test coverage, though.

See issue.

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/SelectOneChoiceFilterTest.java
+++ b/src/test/java/org/javarosa/core/model/SelectOneChoiceFilterTest.java
@@ -131,6 +131,7 @@ public class SelectOneChoiceFilterTest {
         // If we set level2 to "aa", form validation passes. Currently, clearing a choice only updates filter expressions
         // that directly depend on it. With this form, we could force clearing the third level when the first level is cleared
         // by making the level3 filter expression in the form definition reference level1 AND level2.
+        scenario.answer("/data/level1", "b");
         scenario.answer("/data/level2", "bb");
 
         validate = scenario.getValidationOutcome();
--- a/src/main/java/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/main/java/org/javarosa/core/model/ItemsetBinding.java
@@ -92,6 +92,8 @@ public class ItemsetBinding implements Externalizable, Localizable {
         // Return cached list if possible
         if (cachedFilteredChoiceList != null && allTriggerRefsBound && Objects.equals(currentTriggerValues, cachedTriggerValues)
             && Objects.equals(currentRandomizeSeed, cachedRandomizeSeed)) {
+            updateQuestionAnswerInModel(formDef, curQRef);
+
             return randomize && cachedRandomizeSeed == null ? shuffle(cachedFilteredChoiceList) : cachedFilteredChoiceList;
         }
 
@@ -116,10 +118,10 @@ public class ItemsetBinding implements Externalizable, Localizable {
 
         Map<String, SelectChoice> selectChoicesForAnswer = initializeAnswerMap(formDef, curQRef);
 
-        List<SelectChoice> choices = new ArrayList<>();
+        List<SelectChoice> filteredChoiceList = new ArrayList<>();
         for (int i = 0; i < filteredItemReferences.size(); i++) {
             SelectChoice choice = getChoiceForTreeReference(formDef, formInstance, i, filteredItemReferences.get(i));
-            choices.add(choice);
+            filteredChoiceList.add(choice);
             if (selectChoicesForAnswer != null && selectChoicesForAnswer.containsKey(choice.getValue())) {
                 // Keys with values that don't get set here will have null values and must be filtered out of the answer.
                 selectChoicesForAnswer.put(choice.getValue(), choice);
@@ -128,13 +130,13 @@ public class ItemsetBinding implements Externalizable, Localizable {
 
         updateQuestionAnswerInModel(formDef, curQRef, selectChoicesForAnswer);
 
-        cachedFilteredChoiceList = randomize ? shuffle(choices, currentRandomizeSeed) : choices;
+        cachedFilteredChoiceList = randomize ? shuffle(filteredChoiceList, currentRandomizeSeed) : filteredChoiceList;
 
         // TODO: write a test that fails if this is removed. It looks like a no-op because it's not accessing the shuffled collection.
         if (randomize) {
             // Match indices to new positions
-            for (int i = 0; i < choices.size(); i++)
-                choices.get(i).setIndex(i);
+            for (int i = 0; i < filteredChoiceList.size(); i++)
+                filteredChoiceList.get(i).setIndex(i);
         }
 
         //init localization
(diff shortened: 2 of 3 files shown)


src/test/java/org/javarosa/core/model/SelectOneChoiceFilterTest.java
```diff
@@ -131,6 +131,7 @@ public void clearingValueAtLevel1_ShouldClearValuesAtLevels2And3() {
         // If we set level2 to "aa", form validation passes. Currently, clearing a choice only updates filter expressions
         // that directly depend on it. With this form, we could force clearing the third level when the first level is cleared
         // by making the level3 filter expression in the form definition reference level1 AND level2.
+        scenario.answer("/data/level1", "b");
```
lognaturel, line 134: A side-effect of this change is that it's no longer possible to set a value for the second level of a cascade if the first level isn't set. That's a good thing but in practice, I don't think it will make any difference for users since they'd have to visit the first level in order to change its value.

lognaturel:
@seadowg I'm not sure whether you'll want to review or wait until @grzesiek2010 gets back. I think it'd be good to get to QA sooner than later. Hopefully it will be an easy review.

seadowg (dismissed):
Both solutions I've seen of this make sense to me, but I take @lognaturel's point that this preserves the performance benefits of caching for multiple different selects that use the same list. This is a pretty small change and the failing test (it's very specific, but I think it has to be) demonstrates this fixing the issue as far as I can see. I think it'd be good to get this to QA now!

seadowg:
I've tested this with Collect and it looks like there is a failing test - `CatchFormDesignExceptionsTest#whenFormHasDesignErrors_explanationDialogShouldBeDisplayedAndTheFormShouldBeClosed`. I'm thinking this fix to caching might be covering for a form design problem which might be ok? Either way, I'll think it's worth @lognaturel investigating before merge.

lognaturel:
After this change, the first two cases described by https://github.com/getodk/collect/issues/4750 are no longer problems because the answer gets updated earlier. That means if there's a calculate that produces a choice that exists in the choice list, it will be bound appropriately. If there's a calculate that produces a choice not in the choice list, then it will be cleared. I think all that is ok, it just makes it more pressing to address https://github.com/XLSForm/pyxform/issues/546 so users get feedback when they make the mistake of putting a calculate on a user-editable select.

I believe we could still get a test to verify what the now-failing test did with something like https://github.com/getodk/collect/issues/4170. For now, let's get a PR up for QA and I can work on that test.

seadowg:
> I'd like to have a quick play with using mvn install to see if we could go that way.

After a bit of investigation it looks like we could set it up, so you can just do `mvn install` locally and then pull that version in Collect. We just need to make a change to our `pom.xml` here so that it's possible to override the version from the command line and also make a change to Collect, so it uses the `mavenLocal()` repo. The nice thing about our new `Dependencies` object is that we could define a `javarosa_local` dependency alongside `javarosa` and so switching in Collect would be really simple. I'll put PRs together tomorrow.

lognaturel:
@seadowg I think this just needs your approval so I can merge and update Collect!

PR #649, opened 2021-11-01
Set version to 3.3.0

lognaturel:
Prepare for 3.3.0 release. This has gone through QA but was not actually part of the beta. The biggest risk would be that the issues addressed aren't fully. We can follow up with a point release if needed.


yanokwa (approved):
LGTM. I'm assuming you will do the GH release and actually publish it with mvn. If you need me to do either of those, let me know.

lognaturel:
Both should now be done. Please do a quick sanity check.

PR #652, opened 2021-11-21
Fixes for #637 Validation of dateTime widget…

### LINKED ISSUE #637: Validation of dateTime widget against now() fails
#### Software and hardware versions 
Collect v2012.2

#### Problem description
A constraint of `. <= now()` on a dateTime widget doesn't constrain when I enter today's date plus an hour later than now. Putting a day later than now works. It would seems only the date is being checked.

In my debugging, when I output the dateTime value in a note, it only shows the date. Using the jump button shows me both date and time. The XML that is persisted shows `<rdt_start_time>2021-06-15T01:33:00.000-07:00</rdt_start_time>`.

[datetime.xlsx](https://github.com/getodk/collect/files/6738146/datetime.xlsx) is the form I used.

#### Other information 
Enketo shows the datetime, but also seems to check the constraint against only the date.

dimwight:
Resolves #637

Form defined by [datetime637.xlsx](
https://github.com/getodk/javarosa/files/7576116/datetime637.xlsx
) reproduces the issue and verifies the fixes. 

Fixes are self-evident?

Corrects previous misbehaviour; no regression risk?

As above.

No?

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -699,7 +699,8 @@ public class XPathFuncExpr extends XPathExpression {
                 val = NaN;
             }
         } else if (o instanceof Date) {
-            val = (double) DateUtils.daysSinceEpoch((Date) o);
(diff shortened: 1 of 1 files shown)


dimwight:
Oops, missed a test that relies on the current code.

PR #655, opened 2022-01-25
Add geojson parsing support

lognaturel:
Work towards https://github.com/getodk/collect/issues/4943

This PR adds support for a geojson external secondary instance with:
- a .geojson extension
- a `FeatureCollection` at the top level
- all features with `Point` geometry

If the file is parsed successfully, then an external secondary instance is available for use by clients. At that point there's no trace of it having been geojson previously. This matches the XML and CSV external instance implementations.

Ran it in Collect!
Screenshot_1643130113

First I looked to Collect to see how json is parsed there. The org.json library is imported so I thought that's what was being used but it's not (https://github.com/getodk/collect/pull/4997). Once I found that out I looked at gson or jackson for parsing. I went with Jackson because it's slightly more broadly used and Gson is [in maintenance mode](https://github.com/google/gson/issues/1821). Some of the folks who worked on Gson are working on moshi at Square but I'd rather go with the venerable Jackson for now.

I then had to pick a parsing strategy. I decided to stream then using data binding for each feature to reduce memory usage (compared to parsing the whole thing in memory). This matches what the CSV parser does.

This is isolated. The only risk to existing behavior is that I messed up the conditional that chooses which parser to use.

https://test.getodk.cloud/#/projects/149/forms/geojson

[geojson.zip](https://github.com/getodk/javarosa/files/7935766/geojson.zip)

getodk/docs#1423


src/main/java/org/javarosa/core/model/instance/ExternalDataInstance.java
```diff
@@ -68,10 +68,12 @@ public static ExternalDataInstance build(String instanceSrc, String instanceId)
         return new ExternalDataInstance(root, instanceId, instanceSrc);
     }
 
-    private static TreeElement parseExternalInstance(String instanceSrc, String instanceId) throws IOException, InvalidReferenceException, InvalidStructureException, XmlPullParserException, UnfullfilledRequirementsException {
+    private static TreeElement parseExternalInstance(String instanceSrc, String instanceId)
+        throws IOException, InvalidReferenceException, InvalidStructureException, XmlPullParserException, UnfullfilledRequirementsException {
         String path = getPath(instanceSrc);
-        return instanceSrc.contains("file-csv") ?
-            CsvExternalInstance.parse(instanceId, path) : XmlExternalInstance.parse(instanceId, path);
+        return instanceSrc.contains("file-csv") ? CsvExternalInstance.parse(instanceId, path)
```
lognaturel, line 74: I considered pulling this out into a more conventional factory so it could be tested but I don't see a big benefit given how everything else is structured at the moment.

src/main/java/org/javarosa/core/util/GeoUtils.java
```diff
@@ -98,6 +99,16 @@ public static double calculateDistance(List<LatLong> points) {
         return totalDistance;
     }
 
+    public static String geojsonToOdkGeometry(JsonNode geojsonGeometry) throws IOException {
```
lognaturel, line 102: I would have preferred a String parameter but I couldn't quickly find a way to do that.

src/main/java/org/javarosa/core/model/instance/GeoJsonExternalInstance.java
```diff
@@ -0,0 +1,59 @@
...
+        TreeElement item = new TreeElement("item", multiplicity);
+
+        String geometry = GeoUtils.geojsonToOdkGeometry(feature.get("geometry"));
+        TreeElement geoField = new TreeElement("geometry", 0);
+        geoField.setValue(new UncastData(geometry));
+        item.addChild(geoField);
+
+        JsonNode properties = feature.get("properties");
+
+        for (Iterator<Map.Entry<String, JsonNode>> i = properties.fields(); i.hasNext();) {
```
lognaturel, line 50: `forEachRemaining` fails our Android lint check. I marginally prefer this to a `while`.

lognaturel:
Apologies for all the force pushes! @seadowg we'd talked about leaving the node-based parsing but I looked at the CSV implementation and it streams so I changed my mind. It seemed like we'd want to go for lower memory usage sooner than later no mater what.

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
```diff
@@ -0,0 +1,151 @@
...
+        try {
+            GeoJsonExternalInstance.parse("id", new ByteArrayInputStream(GEOMETRY_COLLECTION_GEOJSON.getBytes(StandardCharsets.UTF_8)));
+            fail("Exception expected");
+        } catch (IOException e) {
+            // expected
+        }
+    }
+
+    @Test
+    public void parse_parsesMultipleFeatures() throws IOException {
```
seadowg, line 42: The breakup of these tests is very TDD. I like it. It might feel simpler now you've written them to combine `parse_parsesMultipleFeatures` and `parse_addsGeometryAsChild` - that'll give you a kind of "core behaviour" test. I'll leave that up to you though!

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
```diff
@@ -0,0 +1,151 @@
...
+    @Test
+    public void getOdkCoordinates_convertsPointGeoJsonGeometry() throws IOException {
+        String point = "{ \"type\": \"Point\", \"coordinates\": [ 102, 0.5 ] }";
+        ObjectMapper objectMapper = new ObjectMapper();
+        GeojsonGeometry geometry = objectMapper.readValue(point, GeojsonGeometry.class);
+        assertThat(geometry.getOdkCoordinates(), is("102 0.5 0 0"));
+    }
+
+    @Test
+    public void getOdkCoordinates_throwsException_ifGeometryNotPoint() throws IOException {
```
seadowg, line 73: This test (and the one above) should live in `GeojsonGeometryTest` rather than here I reckon. Alternatively, you don't need `getOdkCoordinates_convertsPointGeoJsonGeometry` (as it's covered by your `parse` tests) and `getOdkCoordinates_throwsException_ifGeometryNotPoint` could be covered here with a test that provides JSON that will cause the fail. The latter is more correct™️ in my opinion, as it would ensure that the exception bubbles up correctly.

src/main/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstance.java
```diff
@@ -0,0 +1,71 @@
...
+                multiplicity++;
+            }
+        }
+
+        return root;
+    }
+
+    private static void validatePreamble(JsonParser jsonParser) throws IOException {
+        if (jsonParser.nextToken() != JsonToken.START_OBJECT) {
+            throw new IOException("GeoJSON file must contain a top-level Object");
```
seadowg, line 54: It looks this line isn't covered by a test

src/main/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstance.java
```diff
@@ -0,0 +1,71 @@
...
+        jsonParser.nextToken();
+        if (!(propertyName.equals("type") && jsonParser.getValueAsString().equals("FeatureCollection"))) {
+            throw new IOException("GeoJSON file must contain a top-level FeatureCollection");
+        }
+
+        jsonParser.nextToken();
+        propertyName = jsonParser.getCurrentName();
+
+        if (!(propertyName.equals("features") && jsonParser.nextToken() == JsonToken.START_ARRAY)) {
+            throw new IOException("GeoJSON FeatureCollection must contain an array of features");
```
seadowg, line 68: It looks this line isn't covered by a test

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
```diff
@@ -0,0 +1,38 @@
...
+    private String type;
+    private ArrayList<String> coordinates;
+
+    public String getType() {
+        return type;
+    }
+
+    public String getOdkCoordinates() {
+        if (!(type.equals("Point") && coordinates.size() == 2)) {
+            throw new UnsupportedOperationException("Only Points are currently supported");
```
seadowg, line 33: Not going to be a problem for long, but maybe we should just skip unsupported elements rather than exploding? That would let us test in Collect with `.geojson` files without worrying about it crashing the app.

seadowg (changes requested):
This is looking good to me after an initial scan through. There are a few pieces of testing that could be improved but the overall structure is nice and simple and makes sense!

build.gradle
```diff
@@ -31,12 +31,16 @@ sourceCompatibility = '1.8'
 
 dependencies {
     // Be sure to update dependencies in pom.xml to match
-    implementation 'joda-time:joda-time:2.10.10'
-    implementation 'org.slf4j:slf4j-api:1.7.31'
+    implementation 'joda-time:joda-time:2.10.13'
+    implementation 'org.slf4j:slf4j-api:1.7.33'
```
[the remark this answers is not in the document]
lognaturel: Yes, I tend to update point releases when I see they're available. I can do in other PR if you prefer.

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+        try {
+            geometry.getOdkCoordinates();
+            fail("Exception expected");
+        } catch (UnsupportedOperationException e) {
+            // expected
+        }
+    }
+
+    private static final String GEOMETRY_COLLECTION_GEOJSON = "{\n" +
```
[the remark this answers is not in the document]
lognaturel: Pulled out a file since that addresses your comment about the `protected` method too. I used the constant because it was easier for me to jump to as I was writing the tests but I also see your point about the visual noise. Too bad we can't use Java 15 `"""`.

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
lognaturel: I think we likely will have a release with only points. In that case it feels better to fail for reasons similar to the "fast external itemsets" bug you identified -- it's not great if a subset of user features are silently filtered out. I don't feel strongly about it but that was my reasoning for exploding.

seadowg: That makes sense. Let's use a checked exception then, so there's not a risk of Collect missing this and it resulting in a crash.

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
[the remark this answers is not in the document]
seadowg: :neckbeard: Kotlin has multiline strings with support all the way back to Java 6 (https://kotlinlang.org/docs/coding-conventions.html#strings).

build.gradle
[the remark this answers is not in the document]
seadowg: That's fine! Just checking.

seadowg (approved):

seadowg (changes requested):
Looking great! As I commented inline, I think we should add a checked exception to deal with the unsupported feature type case.

PR #656, opened 2022-02-09
Replace install local script with Gradle task

seadowg:
Closes #


build.gradle
```diff
@@ -66,12 +66,33 @@ jar {
     }
 }
 
-// Useful for testing
-task explodedJar(type: Copy) {
```
seadowg, line 70: I don't think we need this now we have `installLocal`.

lognaturel (approved):
Sorry about delay. Saw it, loved it, swamped. A wonderfully simple solution. 😍

PR #658, opened 2022-03-07
Geojson improvements

lognaturel:
- Closes https://github.com/getodk/collect/issues/5051
- Adds support for toplevel `id` if it exists. Toplevel `id` is mentioned in the GeoJSON spec so it takes precedence over an `id` property.
- Adds additional flexibility: ignore unknown toplevel elements and accept no explicit properties (though it won't work as choices for a select)

Added tests.

I was initially going to address the issue with a proguard rule in Collect but I realized that really we should be validating the `type` as we do other similar validations.

While validating the type, I realized that properties could be empty. This won't work for selects because we need properties for label and underlying value but we shouldn't crash. While I was there, I implemented support for toplevel `id` which we'd marked as an extension.

This is isolated to the geojson secondary instance implementation so the only regression risks are there. Test coverage is high so it should be quite safe.

See tests.

https://github.com/getodk/docs/issues/1423

### DIFF
--- a/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
+++ b/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
@@ -28,6 +28,8 @@ public class GeojsonFeature {
     private GeojsonGeometry geometry;
     private Map<String, String> properties;
 
+    private String id;
+
     public TreeElement toTreeElement(int multiplicity) throws IOException {
         if (!type.equals("Feature")) {
             throw new IOException("Item of type " + type + " found but expected item of type Feature");
@@ -43,6 +45,16 @@ public class GeojsonFeature {
             for (String property : properties.keySet()) {
                 TreeElement field = new TreeElement(property, 0);
                 field.setValue(new UncastData(properties.get(property)));
+                item.addChild(field);
+            }
+        }
+
+        if (id != null) {
+            if (!item.getChildrenWithName("id").isEmpty()) {
+                item.getChildrenWithName("id").get(0).setValue(new UncastData(id));
+            } else {
+                TreeElement field = new TreeElement("id", 0);
+                field.setValue(new UncastData(id));
                 item.addChild(field);
             }
         }
--- a/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
+++ b/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
@@ -39,10 +39,12 @@ public class GeojsonFeature {
         geoField.setValue(new UncastData(geometry.getOdkCoordinates()));
         item.addChild(geoField);
 
-        for (String property : properties.keySet()) {
-            TreeElement field = new TreeElement(property, 0);
-            field.setValue(new UncastData(properties.get(property)));
-            item.addChild(field);
+        if (properties != null) {
+            for (String property : properties.keySet()) {
+                TreeElement field = new TreeElement(property, 0);
+                field.setValue(new UncastData(properties.get(property)));
+                item.addChild(field);
+            }
         }
 
         return item;
--- a/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
+++ b/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
@@ -17,12 +17,14 @@
 package org.javarosa.core.model.instance.geojson;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import java.io.IOException;
 import java.util.Map;
 import org.javarosa.core.model.data.UncastData;
 import org.javarosa.core.model.instance.TreeElement;
 
 @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class GeojsonFeature {
     private String type;
     private GeojsonGeometry geometry;
--- a/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
+++ b/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
@@ -29,18 +29,22 @@ public class GeojsonFeature {
     private Map<String, String> properties;
 
     public TreeElement toTreeElement(int multiplicity) throws IOException {
+        if (!type.equals("Feature")) {
+            throw new IOException("Item of type " + type + " found but expected item of type Feature");
+        }
+
         TreeElement item = new TreeElement("item", multiplicity);
 
         TreeElement geoField = new TreeElement("geometry", 0);
         geoField.setValue(new UncastData(geometry.getOdkCoordinates()));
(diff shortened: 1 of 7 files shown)


src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
```diff
@@ -124,4 +124,9 @@ public void parse_throwsException_whenGeometryNotPoint() {
             // expected
         }
     }
+
+    @Test
+    public void parse_ignoresUnknownToplevelProperties() throws IOException {
+        TreeElement featureCollection = GeoJsonExternalInstance.parse("id", r("feature-collection-extra-toplevel.geojson").toString());
```
seadowg, line 130: Did this crash before or is there a missing assertion?

src/test/resources/org/javarosa/core/model/instance/geojson/feature-collection-extra-toplevel.geojson
```diff
@@ -0,0 +1,21 @@
...
+        {
+            "type": "Feature",
+            "geometry": {
+                "type": "Point",
+                "coordinates": [
+                    102,
+                    0.5
+                ]
+            },
+            "foo": "bar",
```
seadowg, line 13: Good to make this not exactly the same as the property in `properties` that we do want. It took me a good 5 minutes to understand what was happening here!

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
```diff
@@ -28,6 +28,8 @@
     private GeojsonGeometry geometry;
     private Map<String, String> properties;
 
+    private String id;
```
seadowg, line 33: What's the goal of parsing this out? Is it so it can be referenced in forms (questions/choice filters etc)?

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
```diff
@@ -94,6 +94,12 @@ public void parse_addsAllOtherPropertiesAsChildren() throws IOException {
         assertThat(featureCollection.getChildAt(1).getChild("special-property", 0).getValue().getValue(), is("special value"));
     }
 
+    @Test
+    public void parse_addsFeaturesWithNoProperties() throws IOException {
+        TreeElement featureCollection = GeoJsonExternalInstance.parse("id", r("feature-collection-no-properties.geojson").toString());
+        assertThat(featureCollection.getChildAt(0).getNumChildren(), is(1));
```
seadowg, line 115: Does this result in a crash in Collect if the name/label is missing or is there a graceful error?

seadowg (changes requested):

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
```diff
@@ -17,12 +17,14 @@
...
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import java.io.IOException;
 import java.util.Map;
 import org.javarosa.core.model.data.UncastData;
 import org.javarosa.core.model.instance.TreeElement;
 
 @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
lognaturel, line 27: Without this, unknown properties would cause a crash.

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
lognaturel: Yes and yes. 😫 Fixed and force pushed the last commit. Please review that one again, it was really a disaster.

src/test/resources/org/javarosa/core/model/instance/geojson/feature-collection-extra-toplevel.geojson
lognaturel: Good call. It almost looks like I wandered off before finishing this last commit, sorry. Renamed the property to `ignored` and force pushed.

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonFeature.java
lognaturel: It's to support a top-level `id` because that has meaning in the geojson spec. Other top-level properties are ignored (per that dreadful commit). Commit details:

> From GeoJSON spec: "If a Feature has a commonly used identifier, that identifier SHOULD be included as a member of the Feature object with the name "id", and the value of this member is either a JSON string or number." [https://datatracker.ietf.org/doc/html/rfc7946\#section-3.2](https://datatracker.ietf.org/doc/html/rfc7946/#section-3.2)

We give priority to the toplevel id since it's part of the spec but fall back to `properties/id` since it's also commonly used (see https://github.com/qgis/QGIS/issues/40876, for example).

The `id` node in JavaRosa's internal representation (which could come from top level or `properties/id`) is going to be the default underlying value for selects: https://github.com/XLSForm/pyxform/issues/591 But yeah, you could also use it arbitrarily in a form.

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
lognaturel: The error is caught but I wouldn't call it graceful:

device-2022-03-10-100306

Let me think about improving that.

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
```diff
@@ -84,6 +95,34 @@ public void parse_addsAllOtherPropertiesAsChildren() throws IOException {
...
+        assertThat(featureCollection.getChildAt(0).getNumChildren(), is(4));
+        assertThat(featureCollection.getChildAt(0).getChild("id", 0).getValue().getValue(), is("fs87b"));
+
+    }
+
+    @Test
+    public void parse_prioritizesTopLevelId() throws IOException {
+        TreeElement featureCollection = GeoJsonExternalInstance.parse("id", r("feature-collection-id-twice.geojson").toString());
+        assertThat(featureCollection.getChildAt(0).getNumChildren(), is(4));
+        assertThat(featureCollection.getChildAt(0).getChild("id", 0).getValue().getValue(), is("fs87b"));
```
seadowg, line 110: Using more test-tailored values here like "top-level-id" instead of a realistic one ("fs87b") would really help the readability of the test given that the JSON lives externally. That goes for the above test as well.

seadowg (approved):
Looks good! I've left one comment on a tweak that would be nice to sneak in before merging.

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
lognaturel: Thank you, a nice improvement!

PR #660, opened 2022-03-23
Expose item child nodes in SelectChoice

lognaturel:
Provides direct access to item children from a `SelectChoice`.

Tests.

It's the best I could come up with. It makes the most sense to me that the `SelectChoice` itself would provide direct access to children of the item node that the choice is based on. Once reaching that conclusion there aren't a whole lot of options.

This should be additive and have no effect on users. I don't think there's a regression risk but I think that it might be an incomplete solution. Technically an `item` could have a complex structure (nested nodes). I've never seen it but an XML or geojson document could have it. This doesn't give access to that more complex structure. The only use for it we have right now is to access the well-known `geometry` child so I think it's ok.

Tests.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/main/java/org/javarosa/core/model/ItemsetBinding.java
@@ -191,15 +191,20 @@ public class ItemsetBinding implements Externalizable, Localizable {
         // Provide a default value if none is specified
         value = value != null ? value : "dynamic:" + i;
 
-        TreeElement copyNode = null;
+        SelectChoice choice = new SelectChoice(
+            label,
+            value,
+            labelIsItext,
+            item.getInstanceName() != null
+                ? (TreeElement) formDef.getNonMainInstance(item.getInstanceName()).resolveReference(item)
+                : formDef.getMainInstance().resolveReference(item));
+
+        choice.setIndex(i);
+
         if (copyMode) {
-            copyNode = formDef.getMainInstance().resolveReference(copyRef.contextualize(item));
+            choice.copyNode = formDef.getMainInstance().resolveReference(copyRef.contextualize(item));
         }
 
-        SelectChoice choice = new SelectChoice(label, value, labelIsItext);
-        choice.setIndex(i);
-        if (copyMode)
-            choice.copyNode = copyNode;
         return choice;
     }
 
--- a/src/main/java/org/javarosa/core/model/SelectChoice.java
+++ b/src/main/java/org/javarosa/core/model/SelectChoice.java
@@ -30,6 +30,11 @@ public class SelectChoice implements Externalizable, Localizable {
     public TreeElement copyNode;
 
     /**
+     * The node that this choice represents. Not serialized.
+     */
+    private TreeElement item;
+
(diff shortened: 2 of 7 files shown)


src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -191,15 +191,20 @@ private SelectChoice getChoiceForTreeReference(FormDef formDef, DataInstance for
...
+            label,
+            value,
+            labelIsItext,
+            item.getInstanceName() != null
+                ? (TreeElement) formDef.getNonMainInstance(item.getInstanceName()).resolveReference(item)
+                : formDef.getMainInstance().resolveReference(item));
+
+        choice.setIndex(i);
+
         if (copyMode) {
```
lognaturel, line 204: This `copyMode` stuff isn't tested or documented. I think we should remove all of it in the near-ish future. For now I wanted to at least group it together so it's out of the way.

src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -29,6 +29,11 @@
      */
     public TreeElement copyNode;
 
+    /**
+     * The node that this choice represents. Not serialized.
+     */
+    private TreeElement item;
```
lognaturel, line 35: I briefly considered trying to use `copyNode` for this. But like I said in another comment, I think we should remove all that stuff instead.

src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -75,6 +81,16 @@ public String getValue() {
         return value;
     }
 
+    public String getChild(String childName) {
```
lognaturel, line 84: I also considered `getProperty`. I went with `getChild` because it best matches the underlying tree representation. `Property` makes the most sense for `geojson` and `column` makes the most sense for CSV.

src/test/java/org/javarosa/core/model/SelectChoiceTest.java
```diff
@@ -62,4 +69,54 @@ public void value_should_continue_being_an_empty_string_after_deserialization()
...
+    @Test
+    public void getChild_returnsNull_whenChoicesAreFromSecondaryInstance_andRequestedChildDoesNotExist() {
+        setUpSimpleReferenceManager(r("external-select-geojson.xml").getParent(), "file");
+
+        Scenario scenario = Scenario.init("external-select-geojson.xml");
+        assertThat(scenario.choicesOf("/data/q").get(1).getChild("non-existent"), CoreMatchers.is(nullValue()));
+    }
+
+    @Test
+    public void getChild_updates_whenChoicesAreFromRepeat() throws IOException {
```
lognaturel, line 91: Dude. You could, like, have a repeat that has a field called `geometry`. And then build a select with a `map` appearance on top of that. I don't know why exactly but that makes my brain explode a little bit. 🤯

I almost only populated the `item` field in the case of an external secondary instance because this kind of very dynamic situation scared me. But I can't actually think of a case where it wouldn't work.

PR #663, opened 2022-03-29
Get additional children

lognaturel:
Adds access to additional children from an itemset item in support of https://github.com/getodk/collect/issues/4948

Tests!

Comments inline for a few alternatives I considered.

Additive change. 

I did put all the `copyNode` stuff on a deprecation path. It's not documented. If ever we have a use case for it we can bring it back with tests.

See tests.

Not user-facing so no.

### DIFF
--- a/src/main/java/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/main/java/org/javarosa/core/model/ItemsetBinding.java
@@ -210,7 +210,8 @@ public class ItemsetBinding implements Externalizable, Localizable {
             labelIsItext,
             item.getInstanceName() != null
                 ? (TreeElement) formDef.getNonMainInstance(item.getInstanceName()).resolveReference(item)
-                : formDef.getMainInstance().resolveReference(item));
+                : formDef.getMainInstance().resolveReference(item),
+            labelRef.getNameLast());
 
         choice.setIndex(i);
 
--- a/src/main/java/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/main/java/org/javarosa/core/model/ItemsetBinding.java
@@ -55,16 +55,13 @@ public class ItemsetBinding implements Externalizable, Localizable {
     public TreeReference nodesetRef;   //absolute ref of itemset source nodes
     public IConditionExpr nodesetExpr; //path expression for source nodes; may be relative, may contain predicates
     public TreeReference contextRef;   //context ref for nodesetExpr; ref of the control parent (group/formdef) of itemset question
-       //note: this is only here because its currently impossible to both (a) get a form control's parent, and (b)
+       //note: this is only here because it's currently impossible to both (a) get a form control's parent, and (b)
        //convert expressions into refs while preserving predicates. once these are fixed, this field can go away
 
     public TreeReference labelRef;     //absolute ref of label
     public IConditionExpr labelExpr;   //path expression for label; may be relative, no predicates
     public boolean labelIsItext;       //if true, content of 'label' is an itext id
 
-    public boolean copyMode;           //true = copy subtree; false = copy string value
-    public IConditionExpr copyExpr;    // path expression for copy; may be relative, no predicates
-    public TreeReference copyRef;      //absolute ref to copy
     public TreeReference valueRef;     //absolute ref to value
     public IConditionExpr valueExpr;   //path expression for value; may be relative, no predicates (must be relative if copy mode)
 
@@ -74,6 +71,22 @@ public class ItemsetBinding implements Externalizable, Localizable {
     public boolean randomize = false;
     public XPathNumericLiteral randomSeedNumericExpr = null;
     public XPathPathExpr randomSeedPathExpr = null;
+
+    /**
+     * @deprecated No tests and no evidence it's used.
+     */
+    @Deprecated
+    public boolean copyMode;           //true = copy subtree; false = copy string value
+    /**
+     * @deprecated No tests and no evidence it's used.
+     */
+    @Deprecated
+    public IConditionExpr copyExpr;    // path expression for copy; may be relative, no predicates
+    /**
+     * @deprecated No tests and no evidence it's used.
+     */
+    @Deprecated
+    public TreeReference copyRef;      //absolute ref to copy
 
     /**
      * Returns a list of <code>SelectChoice</code> objects based on the data in the model. If any of the itemset expression
--- a/src/main/java/org/javarosa/core/model/SelectChoice.java
+++ b/src/main/java/org/javarosa/core/model/SelectChoice.java
@@ -42,7 +42,10 @@ public class SelectChoice implements Externalizable, Localizable {
      * 'copy' answer mode, this points to the node to be copied if this
      * selection is chosen this field only has meaning for dynamic choices, thus
      * is unserialized
+     *
+     * @deprecated No tests and no evidence it's used.
      */
+    @Deprecated
     public TreeElement copyNode;
 
     /**
@@ -122,7 +125,6 @@ public class SelectChoice implements Externalizable, Localizable {
 
         return index;
     }
-
 
     public void localeChanged(String locale, Localizer localizer) {
         // TODO Review this commented block
--- a/src/main/java/org/javarosa/core/model/SelectChoice.java
+++ b/src/main/java/org/javarosa/core/model/SelectChoice.java
@@ -1,5 +1,8 @@
 package org.javarosa.core.model;
 
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
 import org.javarosa.core.model.data.helper.Selection;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.services.locale.Localizable;
@@ -10,16 +13,28 @@ import org.javarosa.core.util.externalizable.Externalizable;
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.xform.parse.XFormParseException;
 
-import java.io.DataInputStream;
-import java.io.DataOutputStream;
-import java.io.IOException;
-
 public class SelectChoice implements Externalizable, Localizable {
-
-    private String labelInnerText;
-    private String textID;
-    private boolean isLocalizable;
+    /** The literal value from the child that is used as a value. For inline selects ("static"), this child always has
+     * the name "value". For selects from itemsets ("dynamic"), the child node to use is specified in the form definition
+     * (e.g. country_code)
+     */
     private String value;
+
+    /**
+     * If this choice is from a non-localizable item, the literal value from the child that is used as a label.
+     * For inline selects ("static"), this child always has the name "label". For selects from itemsets ("dynamic"), the
+     * child node to use is specified in the form definition (e.g. the_human_friendly_name)
+     */
+    private String labelInnerText;
+
+    /**
+     * If this choice is from a localizable item, the literal text from the child that is used as a label. This will be
+     * used to look up the localized label based on the current language.
+     */
+    private String textID;
+
+    private boolean isLocalizable;
+
     private int index = -1;
 
     /**
@@ -31,7 +46,7 @@ public class SelectChoice implements Externalizable, Localizable {
     public TreeElement copyNode;
 
     /**
-     * The node that this choice represents. Not serialized.
+     * For selects from itemsets ("dynamic"), the node that this choice represents. Not serialized.
      */
     private TreeElement item;
 
--- a/src/main/java/org/javarosa/core/model/SelectChoice.java
+++ b/src/main/java/org/javarosa/core/model/SelectChoice.java
@@ -3,6 +3,8 @@ package org.javarosa.core.model;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.LinkedHashMap;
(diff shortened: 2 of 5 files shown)


src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -197,7 +210,8 @@ private SelectChoice getChoiceForTreeReference(FormDef formDef, DataInstance for
             labelIsItext,
             item.getInstanceName() != null
                 ? (TreeElement) formDef.getNonMainInstance(item.getInstanceName()).resolveReference(item)
-                : formDef.getMainInstance().resolveReference(item));
+                : formDef.getMainInstance().resolveReference(item),
+            labelRef.getNameLast());
```
lognaturel, line 214: This won't work in the case of a nested reference which I believe would be possible when using an XML secondary instance. That would look like:

```
        <itemset nodeset="instance('cities')/root/item[country= /data/loc/country ]">
            <value ref="name"/>
            <label ref=➡️"basic_deets/human_readable_stuff/label"⬅️/>   
        </itemset>
```

I've never seen it in use. I think we should start with this and handle any issues users bring up.

src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -10,31 +15,51 @@
 import org.javarosa.core.util.externalizable.PrototypeFactory;
 import org.javarosa.xform.parse.XFormParseException;
 
-import java.io.DataInputStream;
-import java.io.DataOutputStream;
-import java.io.IOException;
-
 public class SelectChoice implements Externalizable, Localizable {
+    /** The literal value from the child that is used as a value. For inline selects ("static"), this child always has
```
lognaturel, line 19: As I wrote in the commit message, this structure is really confusing. Really, there are multiple types of choices (localized dynamic, non-localized dynamic, localized static, non-localized static). I wrote the comments to untangle myself and I think they can be helpful for an eventual refactor.

src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -91,15 +111,33 @@ public String getValue() {
...
+
         return null;
     }
 
+    public Map<String, String> getAdditionalChildren() {
+        if (item == null) {
+            throw new IllegalStateException("Can only get the child of a choice from a select from itemset");
+        }
+
+        Map<String, String> children = new LinkedHashMap<>();
```
lognaturel, line 131: I think some users will want this to be preserve the order the attributes were added in, hence the `LinkedHashMap`. I didn't test for order. Should I?

src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -91,15 +111,33 @@ public String getValue() {
...
+        for (int i = 0; i < item.getNumChildren(); i++) {
+            TreeElement child = item.getChildAt(i);
+            if (!child.getRef().getNameLast().equals(labelRefName)) {
```
lognaturel, line 134: This *only* filters out whatever the `label` ref is. I briefly considered filtering out `geometry` but that's static so it feels like Collect should do it. I initially was going to filter out the `value` as well but actually it can be helpful to see, I think.

src/test/java/org/javarosa/core/model/SelectChoiceTest.java
```diff
@@ -119,4 +123,103 @@ public void getChild_updates_whenChoicesAreFromRepeat() throws IOException {
...
+        scenario.answer("/data/repeat[0]/special-property", "AA");
+
+        assertThat(scenario.choicesOf("/data/select").get(0).getValue(), is("a"));
+        Map<String, String> children = scenario.choicesOf("/data/select").get(0).getAdditionalChildren();
+        assertThat(children.keySet(), hasSize(2));
+        assertThat(children, hasEntry("value", "a"));
+        assertThat(children, hasEntry("special-property", "AA"));
+
+        scenario.answer("/data/repeat[0]/special-property", "changed");
+        children = scenario.choicesOf("/data/select").get(0).getAdditionalChildren();
```
lognaturel, line 201: I hadn't really thought about this previously -- because `getChild` and `getAdditionalChildren` use the display text of the children, they're not dynamic. That is, you have to call `getChild` and `getAdditionalChildren` again after the form changes to get new values. I think that's ok -- I can't think of a usage where we'd hold on to one of these while other parts of the form are changing but I still wanted to flag it.

The alternative would be to keep a reference to an `AnswerData` object instead. That would reflect changes to the model.

lognaturel:
@seadowg May this be of value to you. 🤞🤞😬😬

src/test/java/org/javarosa/core/model/SelectChoiceTest.java
```diff
@@ -119,4 +123,103 @@ public void getChild_updates_whenChoicesAreFromRepeat() throws IOException {
         scenario.answer("/data/repeat[0]/special-property", "changed");
         assertThat(scenario.choicesOf("/data/select").get(0).getChild("special-property"), is("changed"));
     }
+
+    @Test
+    public void getChild_throwsException_whenCalledOnAChoiceFromStaticSelect() throws IOException {
```
seadowg, line 128: This will result in a crash on "static" selects in Collect that add the `map` appearance right? I'm still unhappy with the fact that it doesn't work (although I understand that it's an existing issue) but I think we should handle this differently with one of these options:

1. `getChild`/`getAdditionalChildren` throws a checked exception that Collect can anticipate (rather than a `RuntimeException`) and handle with an error message
1. `getChild`/`getAdditionalChildren` return null/empty and Collect disables the "Select place" button for "static" selects (and shows an error on the widget)

src/test/java/org/javarosa/core/model/SelectChoiceTest.java
```diff
@@ -119,4 +123,103 @@ public void getChild_updates_whenChoicesAreFromRepeat() throws IOException {
...
+                                t("special-property")),
+                            t("filter"),
+                            t("select"))))),
+            body(
+                repeat("/data/repeat",
+                    input("value"),
+                    input("label"),
+                    input("special-property")),
+                input("filter"),
+                select1Dynamic("/data/select", "../repeat")
```
seadowg, line 188: Wait... you can use a repeat as the "itemset" for a select?

seadowg (changes requested):

src/test/java/org/javarosa/core/model/SelectChoiceTest.java
lognaturel: https://github.com/getodk/javarosa/pull/660/files#r832740208 😄

src/main/java/org/javarosa/core/model/SelectChoice.java
lognaturel: I don't disagree with you. 😅 I added them in because in the code they're currently only described as static vs dynamic as far as I can tell. Part of the reason I wrote these comments is because you had mentioned static vs dynamic coming from the Collect side of things and I wasn't totally sure that it was inline vs itemset.

Can remove but wanted to clarify that it provides value to me and connects this to other parts of the code.

src/test/java/org/javarosa/core/model/SelectChoiceTest.java
lognaturel: Shoot, I meant to call this out but forgot. My reasoning was that once you're here it's too late and that it should be a client's responsibility to avoid calling these methods in contexts where they're not applicable. These methods only make sense when there's an itemset. So e.g. the `map` appearance which relies on them only makes sense when there's an itemset.

BUT. Do clients know whether they're dealing with an inline or itemset-based select?

> Collect disables the "Select place" button for "static" selects 

Is that possible? If so, then the runtime exception continues to make sense to me.

seadowg: > Is that possible? If so, then the runtime exception continues to make sense to me.

Huh, good question. I'm not sure. I'm still unclear on why a "static" select can't have children - it seems incredibly confusing, and I think the situation will be exacerbated by the `map` appearance (people will assume it works with "static" selects). If that's something we intend to "fix" then I think returning null/empty alone is the best course of action for the moment.

src/main/java/org/javarosa/core/model/SelectChoice.java
lognaturel: > List<Pair<String, String>>

Ok, excellent.

seadowg (changes requested):

src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -43,39 +68,34 @@ public SelectChoice() {
...
+        this(labelID, labelInnerText, isLocalizable, null, null);
     }
 
     public SelectChoice(String labelID, String labelInnerText, String value, boolean isLocalizable) {
-        this(labelID, labelInnerText, value, isLocalizable, null);
+        this(labelID, labelInnerText, value, isLocalizable, null, null);
     }
 
-    public SelectChoice(String labelOrID, String Value, boolean isLocalizable, TreeElement item) {
+    public SelectChoice(String labelOrID, String Value, boolean isLocalizable, TreeElement item, String labelRefName) {
```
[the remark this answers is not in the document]
lognaturel: Does the new code build `SelectChoice`s? The `TreeElement` parameter was just added in support of this work for use by `ItemsetBinding` so I thought adding the new param would have no effect. Maybe it's just for tests? If so, I think we should edit the tests and consider the two signature additions atomic for this JR release. If Collect is building `SelectChoice`s, we should talk about it.

src/main/java/org/javarosa/core/model/SelectChoice.java
lognaturel: Unless there's a better way to get something like a tuple, I prefer to keep the `LinkedHashMap` and add a test on order.

src/main/java/org/javarosa/core/model/SelectChoice.java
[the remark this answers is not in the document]
lognaturel: (I agree about your broader point about not changing the public API. But this one I specifically considered!!)

src/main/java/org/javarosa/core/model/SelectChoice.java
seadowg: Really? I can't even find a `Pair` type in later Java versions. We could bring the Kotlin standard lib into JavaRosa for this? That'll replace Apache Commons in most places I'd imagine.

lognaturel: I was thinking maybe we start introducing Kotlin in the next release and maybe it's even a major version bump? Or do you really think it's something casual we just do whenever?

src/main/java/org/javarosa/core/model/SelectChoice.java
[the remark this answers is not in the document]
seadowg: Yeah, it's in tests which of course still matter! They don't compile just as hard as non-test code 😆. I do think you're right that we can just consider the changes we've made over these two PRs as "one change" though, so it's not actually a big deal.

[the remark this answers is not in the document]
lognaturel: Yes, totally, I should have called it out.

src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -91,15 +111,33 @@ public String getValue() {
     }
 
     public String getChild(String childName) {
-        if (item != null) {
-            TreeElement child = item.getChild(childName, 0);
-            if (child != null) {
-                return child.getValue().getDisplayText();
-            }
+        if (item == null) {
```
lognaturel, line 114: One thing I find hard about JR is that it can be unclear what's a deliberate precondition vs a case that wasn't explicitly considered vs ...? I think this precondition is now much clearer even though it's equivalent to the terser code it replaces. (As I mentioned, this thinking about preconditions is what had nudged me in the direction of runtime exceptions.)

lognaturel:
Thanks for all the good discussion! I think the two things left are 
- Do you still want me to remove the "static" and "dynamic" from https://github.com/getodk/javarosa/pull/663#discussion_r838755927? (Fine by me but want you to read my rationale. Maybe you'll have alternative suggestions after that)
- Kotlin? Stick to LinkedHashMap and add order test? https://github.com/getodk/javarosa/pull/663#discussion_r838809626

src/main/java/org/javarosa/core/model/SelectChoice.java
seadowg: Oh I mean just use the standard library (`org.jetbrains.kotlin:kotlin-stdlib`) as an alternative to Apache Commons. We use it from Java a lot in Collect (before eventually converting).

lognaturel: 🤦🏻‍♀️

build.gradle
```diff
@@ -38,6 +38,8 @@ dependencies {
     // Upgrade to version higher than 1.4 when Collect minSDK >= 26
     implementation 'org.apache.commons:commons-csv:1.4'
 
+    implementation 'org.jetbrains.kotlin:kotlin-stdlib:1.6.10'
```
seadowg, line 41: It begins...

seadowg (approved):

PR #667, opened 2022-04-07
Add tests for geojson integer ids

### LINKED ISSUE #666: Support integer id from geojson files
From [the geojson spec](https://datatracker.ietf.org/doc/html/rfc7946): 
> If a Feature has a commonly used identifier, that identifier SHOULD be included as a member of the Feature object with the name "id", and the value of this member is either a JSON string or number.

Can we make `GeojsonFeature`'s `id` field have type `Object`?

lognaturel:
Closes #666 😈

A few additional test cases to verify that integer ids are supported. I also verified in Collect and didn't see any problems.

### DIFF
--- a/src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
+++ b/src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
@@ -123,6 +123,14 @@ public class GeoJsonExternalInstanceTest {
     }
 
     @Test
+    public void parse_allowsIntegerId() throws IOException {
+        TreeElement featureCollection = GeoJsonExternalInstance.parse("id", r("feature-collection-integer-id.geojson").toString());
+
+        assertThat(featureCollection.getChildAt(0).getNumChildren(), is(4));
+        assertThat(featureCollection.getChildAt(0).getChild("id", 0).getValue().getValue(), is("77"));
+    }
+
+    @Test
     public void parse_ignoresUnknownToplevelProperties() throws IOException {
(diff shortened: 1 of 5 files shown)


src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
```diff
@@ -110,6 +110,16 @@ public void parse_prioritizesTopLevelId() throws IOException {
...
     }
 
+    @Test
+    public void parse_allowsIntegerId() throws IOException {
+        TreeElement featureCollection = GeoJsonExternalInstance.parse("id", r("feature-collection-integer-id.geojson").toString());
+
+        assertThat(featureCollection.getChildAt(0).getNumChildren(), is(4));
+        assertThat(featureCollection.getChildAt(0).getChild("id", 0).getValue().getValue(), is("77"));
+
+        assertThat(featureCollection.getChildAt(1).getChild("id", 0).getValue().getValue(), is("78"));
```
seadowg, line 120: Unless I'm missing something, there is no reason to have more than one feature in the JSON to check that integer IDs get parsed correctly.

seadowg (changes requested):

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
lognaturel: This is not my best habit. While I like targeted tests, I'm also very paranoid about the 1-element case being special. You're right that it's really not necessary here and I've removed it.

seadowg (approved):

PR #669, opened 2022-04-11
Trigger xforms-value-changed after recomputation

### LINKED ISSUE #668: xforms-value-changed events fired before dependent calculations are recomputed
Currently `xforms-value-changed` events are fired before dependent calculations are recomputed. This means that if you have `fieldA` which is user-entered, then `calcA` which produces a value based on `fieldA`, it's not possible to have a `setvalue` action triggered by `fieldA` which uses `calcA`: `calcA` will not be updated yet.

For a practical example, see [this form](https://docs.google.com/spreadsheets/d/1ZFE7aLMrQ7FJ8kjH1oN-41p8EsYGTA4Rpo82RFGl92k/edit#gid=0). For the two `text` fields we'd like to use the calculate values but we can't because of this bug.

lognaturel:
Closes #668

Tests.

I don't see any alternatives without a big change in the structure of the code.

I think this is low risk? It is a change in behavior but I don't think the current behavior is usable.

In [this form](https://docs.google.com/spreadsheets/d/1ZFE7aLMrQ7FJ8kjH1oN-41p8EsYGTA4Rpo82RFGl92k/edit#gid=0), change the triggered calculations for the text fields to use `point_name` and `point_type`.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -394,12 +394,13 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
         setAnswer(data, node);
 
         QuestionDef currentQuestion = findQuestionByRef(ref, this);
+
+        Collection<QuickTriggerable> qts = triggerTriggerables(ref);
+
         if (valueChanged && currentQuestion != null) {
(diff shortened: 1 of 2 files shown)


lognaturel:
Perhaps someone could be using the current behavior to save the prior value of a calculate before it updates. I'm not sure how to reason about that.

seadowg (approved):

PR #671, opened 2022-04-11
Allow fields in group to trigger action outside

### LINKED ISSUE #670: Field in group can't trigger an action outside that group
#### Software versions 
all

#### Problem description
Field in group can't trigger an action outside that group

https://forum.getodk.org/t/triggers-not-working-on-groups-on-odk-collect/34460/6

```java
    @Test
    public void setvalueInGroup_setsValueOutsideOfGroup() throws IOException {
        Scenario scenario = Scenario.init("Setvalue", html(
            head(
                title("Setvalue"),
                model(
                    mainInstance(t("data id=\"setvalue\"",
                        t("g",
                            t("source")),
                        t("destination")
                    )),
                    bind("/data/g/source").type("int"),
                    bind("/data/destination").type("int")
                )
            ),
            body(
                group("/data/g",
                    input("/data/g/source",
                        setvalueLiteral("xforms-value-changed", "/data/destination", "7")
                    ))
            )));

        scenario.answer("/data/g/source", "foo");
        assertThat(scenario.answerOf("/data/destination"), is(intAnswer(7)));
    }

    @Test
    public void setvalueOutsideGroup_setsValueInGroup() throws IOException {
        Scenario scenario = Scenario.init("Setvalue", html(
            head(
                title("Setvalue"),
                model(
                    mainInstance(t("data id=\"setvalue\"",
                        t("source"),
                        t("g",
                            t("destination"))
                    )),
                    bind("/data/source").type("int"),
                    bind("/data/g/destination").type("int")
                )
            ),
            body(
                input("/data/source",
                    setvalueLiteral("xforms-value-changed", "/data/g/destination", "7")
                ))
        ));

        scenario.answer("/data/source", "foo");
        assertThat(scenario.answerOf("/data/g/destination"), is(intAnswer(7)));
    }
   ```
#### Expected behavior
Both tests should pass (the first currently fails).

### PR DESCRIPTION
**Pull request by lognaturel**
Closes #670

I don't understand the code this PR removes (and its original author may not have either) so I'm nervous about it. Putting something up for other eyes to hopefully look at.

Tests to add:
- [ ] `setvalue` in repeat with target outside

### CODE REVIEW

File: src/main/java/org/javarosa/core/model/actions/SetValueAction.java [hunk 1]
```diff
@@ -68,43 +68,24 @@ public void handle(XFormParser p, Element e, Object parent) {
...
-        if (contextRef != null){
-            //Note: right now we're qualifying then testing parentage to see whether
-            //there was a conflict, but it's not super clear whether this is a perfect
-            //strategy
-            if (!contextRef.isAncestorOf(targetReference, false)) {
-                return null;
-            }
-        }
-
-        //TODO: either the target or the value's node might not exist here, catch and throw reasonably
```
lognaturel, line 87: As far as I can tell, this can never happen. In fact, as I wrote elsewhere, I think `FormInstanceParser.verifyActions` is being too restrictive at parse time.

src/main/java/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -68,43 +68,24 @@ public void handle(XFormParser p, Element e, Object parent) {
...
-                throw new NullPointerException(failMessage);
-            } else if (references.size() > 1) {
-                throw new XPathTypeMismatchException("XPath nodeset has more than one node [" + references + "]; Actions can only target a single node reference. Refine path expression to match only one node.");
-            } else {
-                targetReference = references.get(0);
+        List<TreeReference> references = context.expandReference(targetReference);
+        if (references.size() == 0) {
+            // If after finding our concrete reference it is a template, this action is outside of the
+            // scope of the current target, so we can leave.
+            if (model.getMainInstance().hasTemplatePath(target)) {
```
lognaturel, line 81: I have been unable to come up with a form that would result in either an expanded reference set of size 0 or a reference with a template path. It feels safer to leave these cases in for now, though.

src/main/java/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -113,18 +94,14 @@ public TreeReference processAction(FormDef model, TreeReference contextRef) {
             //an unbound template, so see if such a reference could exist. Unfortunately this
             //won't be included in the above walk if the template is nested, since only the
             //top level template retains its subelement templates
-            if(model.getMainInstance().hasTemplatePath(target)) {
+            if (model.getMainInstance().hasTemplatePath(target)) {
```
lognaturel, line 97: Same as above -- I tried to come up with a form def that would lead to a null resolved reference and couldn't do it but feel it's safer to leave the case in for now.

src/main/java/org/javarosa/core/model/actions/SetValueAction.java
lognaturel: For 0 refs, I believe FormInstanceParser.verifyActions makes that impossible. 

TODO: try a predicate that does lead to a node being found at parse time but not at event time.

src/test/java/org/javarosa/core/model/actions/OdkNewRepeatEventTest.java
```diff
@@ -137,6 +138,43 @@ public void newRepeatInstance_doesNotTriggerActionOnUnrelatedRepeat() throws IOE
         assertThat(scenario.answerOf("/data/repeat2[2]/q1").getDisplayText(), is("barbaz"));
     }
 
+    @Test
+    public void newRepeatInstance_usesPreviousInstanceAsDefault() throws IOException {
```
lognaturel, line 142: Name implies this is what always happen, should use "canUse" instead

src/test/java/org/javarosa/core/model/actions/OdkNewRepeatEventTest.java
```diff
@@ -137,6 +138,43 @@ public void newRepeatInstance_doesNotTriggerActionOnUnrelatedRepeat() throws IOE
...
+        Scenario scenario = Scenario.init("Default from prior instance", html(
+            head(
+                title("Default from prior instance"),
+                model(
+                    mainInstance(t("data id=\"default-from-prior-instance\"",
+                        t("repeat",
+                            t("node"))
```
lognaturel, line 149: Don't call this `node`.

src/main/java/org/javarosa/core/model/actions/SetValueAction.java
lognaturel: I can get hit the `references.size() == 0` and then `hasTemplatePath` code path here if I **comment out the contents of `verifyActions`** and run the following:

```
   @Test
    public void foo() throws IOException {
        Scenario scenario = Scenario.init("Setvalue into repeat", html(
            head(
                title("Setvalue into repeat"),
                model(
                    mainInstance(t("data id=\"setvalue-into-repeat\"",
                        t("dest_index", "1"),
                        t("source"),
                        t("repeat",
                            t("destination")
                        )
                    ))
                )
            ),
            body(
                input("/data/dest_index"),
                input("/data/source",
                    setvalue("xforms-value-changed", "/data/repeat[position()=/data/dest_index]/destination", "/data/source")),
                repeat("/data/repeat",
                    input("/data/repeat/destination")
                ))));

        scenario.createNewRepeat("/data/repeat");
        scenario.createNewRepeat("/data/repeat");
        scenario.createNewRepeat("/data/repeat");

        scenario.answer("/data/source", "foo");
        assertThat(scenario.answerOf("/data/repeat[0]/destination").getDisplayText(), is("foo"));

        scenario.answer("/data/dest_index", "7");
        scenario.answer("/data/source", "bar");
    }
```

Again, I think `verifyActions` is doing more than it should be, but it makes sense to me to wait for a user to report.

lognaturel:
@seadowg and I had a good chat about this. He's going to chew on it and see if he thinks the tests I added are meaningful and whether he can think of cases we should add.

Whether or not he gets feedback to me by then, I'll merge tomorrow and do a beta.

seadowg:
@lognaturel just a heads-up that I won't have a chance to review again today

lognaturel:
Merging so we can do the beta. @seadowg will take a look before we release!

PR #674, opened 2022-04-14
Fix publishing to OSSRH

yanokwa:
Based on https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#Publishing-using-Apache-Maven

Closes #674

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,36 +1,27 @@
 name: Publish
 
 on: push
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

yanokwa:
Closes #674

lognaturel:
We'll just have to be extra careful about the versions we put in the pom and gradle files because whatever it is will be published.

PR #681, opened 2022-05-04
Increase robustness of geojson parsing by allowing extra properties and null values

lognaturel:
Closes issues described on forum starting [here](https://forum.getodk.org/t/odk-collect-v2022-2-beta-select-from-map-geojson-datasets/36913/17?u=ln). The existing code naively assumed that the geojson top level object would have just a `type` followed by `features` in that order and with no other properties.

Added tests based on forum reports. Note that the test case provided includes another nested property with name `type` which was a really good inclusion. One thing to think about in review is whether there are other tricky structures like that which we should test for.

I briefly considered using an `ObjectMapper` instead of the parser. But I do think it's better to be able to stream the parsing. I think this is much more robust but it's possible there are still cases I haven't thought of.

This does change the geojson parsing strategy quite a bit so there's risk. Our test coverage is good, though, so I don't think it's high.

See tests.

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
+++ b/src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
@@ -146,8 +146,8 @@ public class GeoJsonExternalInstanceTest {
     }
 
     @Test
-    public void parse_allowsNulls() throws IOException {
-        TreeElement featureCollection = GeoJsonExternalInstance.parse("id", r("feature-collection-with-nulls.geojson").toString());
+    public void parse_allowsNullFeaturePropertyValues() throws IOException {
+        TreeElement featureCollection = GeoJsonExternalInstance.parse("id", r("feature-collection-with-null.geojson").toString());
         assertThat(featureCollection.getChildAt(0).getNumChildren(), is(4));
         assertThat(featureCollection.getChildAt(0).getChild("extra", 0).getValue().getDisplayText(), is(""));
     }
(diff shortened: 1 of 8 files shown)


src/test/resources/org/javarosa/core/model/instance/geojson/feature-collection-with-nulls.geojson
```diff
@@ -0,0 +1,7 @@
+{
+    "type": "FeatureCollection",
+    "extra": null,
```
seadowg, line 3: Why do we have a top level `"extra"` set to `null` here as well?

src/test/java/org/javarosa/core/model/instance/geojson/GeoJsonExternalInstanceTest.java
```diff
@@ -132,4 +144,11 @@ public void parse_throwsException_whenGeometryNotPoint() {
             // expected
         }
     }
+
+    @Test
+    public void parse_allowsNulls() throws IOException {
```
seadowg, line 149: Probably best to be explicit that we mean null feature properties here.

seadowg (changes requested):
As well as the inline comments:

* I do feel like the parsing feels a little "too close to the metal" here. Why not use a proper object mapper here?
* The `.geojson` example files would be a lot easier to read if they were formatted. You can add `.geojson` as a JSON file type in IntelliJ using [File type associations](https://www.jetbrains.com/help/idea/creating-and-registering-file-types.html) and then auto format the files.

lognaturel:
> Why not use a proper object mapper here?

Streaming means we don't have to have two in-memory representations of the file. These files can get big so it seems worth it. It also better matches the CSV implementation. More on that in the original PR (https://github.com/getodk/javarosa/pull/655). I'd rather keep this unless we have a really good reason not to.

> The .geojson example files would be a lot easier to read if they were formatted. 

I like the semantic formatting of those files. But you're right that it's annoying to have files that aren't auto formatted, fixed, thanks.

PR #685, opened 2022-05-24
Stop HTML encoding unicode characters when serializing XForm

seadowg:
Closes https://github.com/getodk/collect/issues/4554

New test and verified it fixed the issue using Collect.

We could attempt to only stop HTML encoding for `formid` and `version`, but that would actually be more difficult as we'd have to change the way we serialize - `KXmlSerializer` only allows us to switch Unicode escaping on or off for the whole document (based on the desired output encoding). This solution also removes weirdness with Emoji (and other high digit Unciode characters) being encoded to two HTML characters which causes problems downstream for Central.

People doing analysis on submission XML directly will be affected as they might expect HTML encoded files (although this wasn't the case for Enketo submissions anyway). There could be other unforeseen consequences, so I'd advocate for us getting this into an early Collect beta and then announcing the change on the forum.

Any form that has Unicode characters anywhere and also submissions that include Unicode characters (like an Emoji in a text question)

### DIFF
--- a/src/main/java/org/javarosa/xform/util/XFormSerializer.java
+++ b/src/main/java/org/javarosa/xform/util/XFormSerializer.java
@@ -19,14 +19,13 @@ package org.javarosa.xform.util;
 import org.kxml2.io.KXmlSerializer;
 import org.kxml2.kdom.Document;
 import org.kxml2.kdom.Element;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayOutputStream;
 import java.io.DataOutputStream;
-import java.io.OutputStreamWriter;
 import java.io.UnsupportedEncodingException;
-import java.io.Writer;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import java.nio.charset.StandardCharsets;
 
 /* this is just a big dump of serialization-related code */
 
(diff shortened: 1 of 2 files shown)


lognaturel:
To evaluate risk and potential downstream effects, @seadowg will look into:
- Whether instances written by Collect are currently UTF-8 encoded
- Whether instances written by either Enketo or Collect have anything in the `encoding` XML root attribute

lognaturel:
Strangely, someone just reported running into a related issue. Enketo can't parse an XML submission with complex emojis. That led me to check the char encoding of submission files and unfortunately `file` says `charset=us-ascii`.

lognaturel:
There's something really weird going on. A submission that fails has `&#55357;&#56852;`. But when I try to add the 😔 emoji my XML has `&#128532;` in it.

Similarly, the issue you saw was with 👍, right? When I add that in Collect it is `&#128077;` and works end-to-end. You got `&#55357;&#56397;`. Seems like it's device related? This explains why we haven't heard of this issue much. I was pretty sure I'd tried emoji before and seen people use them.

lognaturel:
I agree with your reasoning. Let's make sure this is in our next beta.

Happy to write the forum post unless you'd rather take this one all the way now that you've done a deep dive!

lognaturel (approved):

seadowg:
> Happy to write the forum post unless you'd rather take this one all the way now that you've done a deep dive!

Can do! I'll write it up as part of the release notes for the v2022.3 Beta so that it can go in Github Releases and the Forum (for maximum 👀).

PR #691, opened 2022-09-15
Expose created entities after finalizing form

seadowg:
Work towards https://github.com/getodk/collect/issues/5231
Blocked by #692 

A form with an `entities:create` element will now have an entity exposed (built from the `saveto` attribute questions) through the `FormEntryModel` after finalization:

```kotlin
formEntryController.finalizeFormEntry()
val entities = formEntryController.getModel().getExtras().get(Entities.class).getEntities()
```

Instead of being included in the standard parse/runtime code, the entity features are available as a "plugin" that's added by wrapping the `XFormParserFactory` and adding a `FormPostProcessor` to `FormEntryController`:

```kotlin
XFormUtils.setXFormParserFactory(EntityXFormParserFactory(XFormParserFactory()))
val controller = FormEntryController(XFormUtils.getFormFromFormXml("myForm.xml", null)).also {
    it.addPostProcessor(EntityFormFinalizationProcessor())
}
```

For all this to work, the PR adds a few features to allow these kinds of "plugins":

* `XFormParser` can have `BindAttributeProcessor`, `FormDefProcessor` and `ModelAttributeProcessor` objects added to it to allow a plugin to inspect attributes on `bind` elements, the model attributes and to inspect parsed `FormDef` objects respectively
* Arbitrary data can be attached to and retrieved from `FormDef` objects through `getExtras`
* Extras on `FormDef` are retained when through the serialization process that clients uses for caching (using `writeExternal`/`readExternal`)
* Data can also be attached to `FormEntryModel` through model with a similar interface (although this data is not ready for serialization as the `FormEntryModel` is not expected to be serialized)
* As shown in the above example, post processors (run as form finalization) can be added to `FormEntryController` through `addPostProcessor` 
* `XFormParser.parse` now throws a specific (non-runtime) exception and the old `XFormParseException` has been deprecated (it seemed like too much work to remove this as part of the PR)
* The above change meant a lot of files were touched to change the `throws` declaration on test methods

New tests.

We initially looked at a solution that just added entity parsing/processing into existing JavaRosa classes. This felt like it was adding a lot of noise, which is why we experimented with trying out a more "plugin" oriented approach. I really like where this has ended up - it should let us extract other features in similar ways (like geometry) which will help us declutter `XFormParser`.

Should just add the new feature! Very little existing code was changed.

Entity forms (there is a spec link in the issue).

I imagine we will want to update docs, but not sure we're ready quite yet.


seadowg:
@lognaturel I'd like to do some tidying up on the code and have a bit of a deeper think about the touchpoints between Collect and JavaRosa. It would be good to get initial feedback on this before that though.

src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
```diff
@@ -89,11 +89,15 @@ FormInstance parseInstance(Element e, boolean isMainInstance, String name, Map<S
 
             List<TreeElement> meta = root.getChildrenWithName("meta");
             if (!meta.isEmpty()) {
-                List<TreeElement> entities = meta.get(0).getChildrenWithName("entities:entity");
+                List<TreeElement> entity = meta.get(0).getChildrenWithName("entities:entity");
```
lognaturel, line 92: Will need to add support for the `http://www.opendatakit.org/xforms/entities` namespace and a way to get children by namespace and name.

lognaturel (commented):
I think this is a very fine spike!

One high-level feeling I have looking at this is I wish it weren't so entangled with existing  machinery. I think of the entities concept as a wrapper/layer around the existing ODK XForms spec. It would be really nice if the code could reflect that and not touch everything. On the other hand, given that we've decided to add this layer as part of the form spec I don't see a good way to achieve that goal without entirely sacrificing performance.

lognaturel:
Maybe split out the API checker change?

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1921,6 +1927,14 @@ private void parseBind(Element element) {
     private void addBinding(DataBinding binding) {
         bindings.add(binding);
 
+        IDataReference reference = binding.getReference();
+        Optional<TreeElement> maybeSaveTo = binding.getAdditionalAttributes().stream().filter(treeElement -> treeElement.getName().equals("saveto")).findFirst();
```
lognaturel, line 1931: Should eventually be a known attribute rather than going to `getAdditionalAttributes`

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1004,7 +1008,16 @@ public void preloadInstance(TreeElement node) {
 
     public boolean postProcessInstance() {
         actionController.triggerActionsFromEvent(Actions.EVENT_XFORMS_REVALIDATE, elementsWithActionTriggeredByToplevelEvent, this);
-        return postProcessInstance(mainInstance.getRoot());
+        boolean instanceChanged = postProcessInstance(mainInstance.getRoot());
+
+        List<Pair<String, String>> fields = saveTos.stream().map(saveTo -> {
+            IDataReference reference = saveTo.getFirst();
+            String answer = mainInstance.resolveReference(reference).getValue().getDisplayText();
```
lognaturel, line 1015: I think we actually want the underlying value not the display text. This makes a different for selects.

This won't be great performance-wise. Probably fine to start here but I think eventually we'll want to do a single parse pass to get everything out. Might not be the first part of instance save we want to optimize, though, it's pretty slow already.

src/main/java/org/javarosa/xform/parse/XFormParser.java
seadowg: "saveto" should not remain in `getAdditionalAttributes()` (clients shouldn't be able to access it). You should be able to call `binding.getSaveTo()`.

seadowg: Or, maybe `saveto` should live somewhere else entirely and not in `getAdditionalAttributes()` or on the `DataBinding`.

src/main/java/org/javarosa/core/model/FormDef.java
seadowg: Do you have a scenario we could use for a failing test for this? I've tried writing one with selects (`fillingFormWithSelectSaveTo_andWithCreate_savesValuesCorrectlyToEntity`), but it's happy.

src/main/java/org/javarosa/entities/EntityFormPostProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Nullable
+    private String getDatasetToCreateWith(FormInstance mainInstance) {
+        TreeElement root = mainInstance.getRoot();
+        TreeElement meta = root.getChild("meta");
+        if (meta != null) {
+            TreeElement entity = meta.getChild("entity");
+
+            if (entity != null && entity.getNamespace().equals(ENTITIES_NAMESPACE)) {
```
seadowg, line 54: I feel like we should probably be checking this namespace (or searching with it) for all these elements. I'm thinking it would be good to look at unit tests for `EntityFormPostProcessor` to drive that out, but not sure how worth it is given how lax JavaRosa tends to be around namespaces usually?

src/main/java/org/javarosa/entities/EntityFormPostProcessor.java
```diff
@@ -39,4 +40,25 @@ public void processForm(FormEntryModel formEntryModel) {
...
+            List<TreeElement> entity = meta.get(0).getChildrenWithName("entity")
+                .stream()
+                .filter(node -> node.getNamespace().equals("http://www.opendatakit.org/xforms/entities"))
+                .collect(Collectors.toList());
+
+            if (!entity.isEmpty()) {
+                List<TreeElement> create = entity.get(0).getChildrenWithName("create");
+
+                if (!create.isEmpty()) {
+                    return create.get(0).isRelevant();
```
seadowg: @lognaturel from our discussion [in the spec](https://docs.google.com/document/d/1-O_wT0JZZFaWvzHl_krrzERJcnrdGMKP55ZZ_hxV0gU/edit#), it seems like we actually don't want this check yet as we won't be included it (support for relevance on `create`) in the spec. Is that right? I think it'll be confusing to include if it's not something we document or that Central supports.

lognaturel: I think the important thing is to check for the existence of the `create` node *in the instance to be submitted*. I was previously thinking that it would exist if relevant and not exist if non-relevant but now I realize that may not be true depending on when the pruning of non-relevant nodes happens.

seadowg: > Central is only looking for create on submission so it will implicitly support conditional creation right away.

Ah ok. I think we should just include it in the spec then so it's clear that's happening and so Central/Collect are both consistent to what you've documented.

> previously thinking that it would exist if relevant and not exist if non-relevant but now I realize that may not be true depending on when the pruning of non-relevant nodes happens

It looks like we don't "remove" the non-relevant nodes until serialization: https://github.com/getodk/javarosa/blob/1b4a7b42ee599f32eaad3fe9118178d009782dc6/src/main/java/org/javarosa/model/xform/XFormSerializingVisitor.java#L179.

Given that, I think we do this check to be able to expose entities at "runtime".

lognaturel (commented):
I really love how the separation turned out! It's really nice to have all entities-related implementation in one place and it feels like the new hooks could provide broader value.

I think the theme in my feedback is "naming is hard" but particularly it's helpful if we can use names that clearly identify whether something is an abstract declaration (an entity declaration, a form def) vs a concrete thing (an actual built entity, a filled form instance).

src/test/java/org/javarosa/xform/parse/EntitiesTest.java
```diff
@@ -0,0 +1,351 @@
...
+        scenario.answer("Tom Wambsgans");
+
+        scenario.finalizeInstance();
+        List<Entity> entities = scenario.getFormEntryController().getModel().getAttachment(EntitiesAttachment.class).getEntities();
+        assertThat(entities.size(), equalTo(1));
+        assertThat(entities.get(0).fields, equalTo(asList(new Pair<>("name", "Tom Wambsgans"))));
+    }
+
+    @Test
+    public void mustUseCorrectNamespace() throws IOException {
```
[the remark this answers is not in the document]
lognaturel: Ah, I see, it's only checked on create. So still two places to add and maybe the test should be split.

src/main/java/org/javarosa/entities/EntityFormPostProcessor.java
lognaturel: > I think we should just include it in the spec

You mean make conditional create part of the initial spec, right? That sounds good to me!

seadowg: > You mean make conditional create part of the initial spec, right?

Yup!

src/main/java/org/javarosa/core/util/externalizable/ExtWrapExternalizable.java
```diff
@@ -0,0 +1,48 @@
+package org.javarosa.core.util.externalizable;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+
+public class ExtWrapExternalizable extends ExternalizableWrapper {
```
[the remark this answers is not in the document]
seadowg: Ah I hadn't seen `ExtWrapBase` - will have a look at it.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -89,6 +94,7 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
     public static final int TEMPLATING_RECURSION_LIMIT = 10;
 
     private static EventNotifier defaultEventNotifier = new EventNotifierSilent();
+    private Map<String, Externalizable> parseAttachments = new HashMap<>();
```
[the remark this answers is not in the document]
seadowg: Make sense. I'll rethink the naming on this API and also consider some docs comments.

src/main/java/org/javarosa/entities/Entity.java
```diff
@@ -0,0 +1,16 @@
+package org.javarosa.entities;
+
+import kotlin.Pair;
+
+import java.util.List;
+
+public class Entity {
+
+    public final String dataset;
+    public final List<Pair<String, String>> fields;
```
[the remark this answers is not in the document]
seadowg: I'm thinking `properties` then. Might be good for us to do a quick "glossary" session to try and get some common parlance going?

src/main/java/org/javarosa/entities/internal/EntityFormParseAttachment.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class EntityFormParseAttachment implements Externalizable {
```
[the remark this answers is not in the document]
seadowg: I think the "attachments" potentially need their own abstraction. Something like Android's `Bundle` (often used in objects to hold "extras") might work.

src/test/java/org/javarosa/xform/parse/EntitiesTest.java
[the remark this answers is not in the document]
seadowg: Yeah, I think we should look at unit tests for `EntityPostProcessor` for this (or pull out something else that parses out the dataset/create). Coming back to this, I think we should do that now rather than file issues.

src/main/java/org/javarosa/form/api/FormPostProcessor.java
```diff
@@ -0,0 +1,6 @@
+package org.javarosa.form.api;
+
+public interface FormPostProcessor {
```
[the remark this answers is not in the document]
lognaturel: `FormInstancePostProcessor` `FormEntryPostProcessor` `FormEntryFinalizer`

src/main/java/org/javarosa/form/api/FormEntryController.java
```diff
@@ -198,6 +202,16 @@ public int stepToPreviousEvent() {
         return stepEvent(false);
     }
 
+    public void finalizeForm() {
```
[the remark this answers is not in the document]
lognaturel: `finalizeEntry`

lognaturel:
To add: 💥 if spec version doesn't match -- https://docs.google.com/document/d/1-O_wT0JZZFaWvzHl_krrzERJcnrdGMKP55ZZ_hxV0gU/edit#heading=h.pilfnyfgdjnn

src/main/java/org/javarosa/core/util/externalizable/ExtWrapExternalizable.java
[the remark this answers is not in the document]
seadowg: `ExtWrapBased` doesn't "externalize" the object's type meaning that it must be known at compile time. This doesn't work when externalizing a value that is being referenced by a super type like with `Extras`.

src/main/java/org/javarosa/xform/parse/StandardBindAttributesProcessor.java
```diff
@@ -118,16 +119,18 @@ DataBinding createBinding(IXFormParserFunctions parserFunctions, FormDef formDef
...
+                if (bindAttributeProcessor.getUsedAttributes().contains(new Pair<>(namespace, name))) {
                     bindAttributeProcessor.processBindingAttribute(name, element.getAttributeValue(i), binding);
                 }
             }
         });
 
         List<String> processorAttributes = bindAttributeProcessors.stream()
             .flatMap((Function<BindAttributeProcessor, Stream<String>>) bindAttributeProcessor -> {
-                return bindAttributeProcessor.getUsedAttributes().stream();
+                return bindAttributeProcessor.getUsedAttributes().stream().map(Pair::getSecond);
```
seadowg, line 133: We'll need to update `StandardBindAttributeProcessor` to only remove attributes with matching name **AND** namespace so we don't remove someone else's `saveto` here.

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -673,7 +682,19 @@ private void parseMeta(Element e) {
     }
 
     //for ease of parsing, we assume a model comes before the controls, which isn't necessarily mandated by the xforms spec
-    private void parseModel(Element e) {
+    private void parseModel(Element e) throws XFormParseException {
+        modelAttributeProcessors.stream().forEach(processor -> {
+            for (int i = 0; i < e.getAttributeCount(); i++) {
```
lognaturel, line 698: I'm always terrified to see any kind of nested iteration but I guess we don't really care too much about form parse time (within reason) and we generally expect at most a handful of attributes and processors.

src/test/java/org/javarosa/entities/internal/EntityFormParseProcessorTest.java
```diff
@@ -25,6 +26,76 @@
...
+            body(
+                input("/data/name")
+            )
+        );
+
+        EntityFormParseProcessor processor = new EntityFormParseProcessor();
+
+        XFormParser parser = new XFormParser(new InputStreamReader(new ByteArrayInputStream(form.asXml().getBytes())));
+        parser.addProcessor(processor);
+        parser.parse(null);
```
lognaturel, line 167: You don't want to add any kind of assertion here? I think this is fine but I feel like I've seen you throw back tests like this so want to give you a chance to consider it again in case you do care.

lognaturel (approved):
The new naming really helps pull things together for me. Thanks for investing in it. Liking the way some other things have moved, too.

src/test/java/org/javarosa/entities/internal/EntityFormParseProcessorTest.java
seadowg: I guess it came from writing these out in order (in a TDD red-green cycle). The only needed a test where something didn't explode to get the code where I wanted, but you're right that it's a bit confusing to read after the fact. I'll add a "this is working as expected" assertion.

PR #692, opened 2022-09-20
Simplify Android API checking

seadowg:
This PR removes the need for an extra dummy project to run Android lint. Instead, the JavaRosa gradle project can be conditionally configured as an Android Library (as opposed to a Java project) which allows Android tasks (such as `lintDebug` to be run against the code.


lognaturel:
That’s very nice! Did you do a quick sanity check to make sure it does fail as expected if using unsupported APIs?

seadowg:
> That’s very nice! Did you do a quick sanity check to make sure it does fail as expected if using unsupported APIs?

Yup! For example, if desugaring is removed it will fail due to the `stream()` calls.

PR #694, opened 2022-10-13
Handle empty saveto questions

seadowg:
This solves issues found in [collect/#5294](https://github.com/getodk/collect/pull/5294)

New tests

The only alternative was to store empty answers as `null` in their respective properties, but after discussing with the Central team, we felt it was easier to treat these cases as an empty string so that we can keep entity properties non-nullable.

Not a big change here. Should only affect questions with `saveto` attributes


src/main/java/org/javarosa/entities/EntityFormFinalizationProcessor.java
```diff
@@ -31,8 +32,13 @@ public void processForm(FormEntryModel formEntryModel) {
         if (dataset != null) {
             List<Pair<String, String>> fields = saveTos.stream().map(saveTo -> {
                 IDataReference reference = saveTo.getFirst();
-                String answer = mainInstance.resolveReference(reference).getValue().getDisplayText();
-                return new Pair<>(saveTo.getSecond(), answer);
+                IAnswerData answerData = mainInstance.resolveReference(reference).getValue();
+
+                if (answerData != null) {
```
lognaturel, line 37: Arg, `null`.

lognaturel (approved):

PR #696, opened 2022-10-26
Fix namespaced TreeElements

### LINKED ISSUE #695: `bind` calculates do not work for nodesets with prefixed nodes for cached `FormDef`
#### Software versions 
JavaRosa v4.0.0-SNAPSHOT

#### Problem description

Form entry when using a "cached" (serialized then deserialized) `FormDef` won't result in calculates being populated if their target nodeset contains a node with a prefix (like `/data/namespace:field`).

Using the below form...

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:example="http://example.com">
    <h:head>
        <h:title>Prefix calculate</h:title>
        <model odk:xforms-version="1.0.0">
            <instance>
                <data id="prefix-calculate" version="1">
                    <question/>
                    <example:calculate/>
                </data>
            </instance>
            <bind calculate="/data/question" nodeset="/data/example:calculate" type="string"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/question">
            <label>Question</label>
        </input>
    </h:body>
</h:html>
```

1. Parse the form with `XFormParser`
2. Serialize and deserialize the `FormDef` using `writeExternal`/`readExternal`
3. Use the deserialized `FormDef` for form entry (with `FormEntryController`), filling out the single question with an answer and then finalizing and serializing the instance (with `finalizeForm` and `XFormSerializingVisitor`)

The serialized instance will contain `<example:calculate />` rather than `<example:calculate>...</example:calculate>` with your answer. The calculate node is correctly populated when using a `FormDef` that comes from the parser directly.

#### Expected behavior

The node should be populated for a deserialized `FormDef`.

#### Other information 

Discovered using Collect, so there is some risk that the bug might actually be there. My understanding is that calculates are handled in JavaRosa. It would be good to write a failing test to confirm the problem in JavaRosa.

lognaturel:
Closes #695 and fixes some additional issues I discovered around path expressions with namespaced nodes. I considered breaking these up into multiple PRs but they feel easier to understand together. If any single commit feels riskier or harder to understand, I can separate it out.

Added a test for each of the fixes, ran the whole test suite. Have NOT tried in Collect yet.

The first commit fixes an issue with XPath path expressions that have namespace prefixes in them. In some case, the namespace prefix was not being considered correctly. The history there is that https://github.com/getodk/javarosa/pull/153/files added support for namespaced instance elements. Either that was incomplete or later performance-related changes broke this (related to https://github.com/getodk/javarosa/pull/192). I didn't see any alternatives here and it feels safe to me.

The second commit fixes the original issue. After deserialization, there was no record of the namespace prefix. I was initially confused because #155 added testing related to serialization as discussed at https://github.com/getodk/javarosa/pull/153#issuecomment-316678155 but after closer inspection I realized we had a misunderstanding about which serialization. The test verifies serializing the instance to XML, not caching the form definition. I saw two options to fix this: serialize the namespace prefix or serialize a map from prefix to namespace. I went with serializing the namespace prefix for each node. There's potential for redundancy but I don't expect namespaced instance nodes to be very common and we can revisit if we think it's worth taking on extra complexity. This also feels safe to me.

After fixing the first issue, I went looking for other comparisons against `TreeElement` names that might cause issues. I noticed that there were problems around attributes.

In the third commit, I added testing for `getAttribute`. I also discovered an existing test with a bad assertion that would have caught the issue I was noticing. Things got a little awkward here. `getAttribute` gets called by parsing code which wants to get attributes according to well-known node names and namespaces. But it also gets called when evaluating XPath path expressions. In that case, the namespace prefix and the node name are passed in and the namespace is always null. I considered breaking up those two cases but I wasn't confident enough that I could identify exactly when each case would be needed. Even though it's not very pretty, it feels safer to me to keep the existing single method with two cases. As it is now, this feels low risk but maybe a bit higher risk than the others. Another thing I found out doing this work is that when attributes on instance elements are in the default namespace, they're set to an empty string rather than `null`. I couldn't find where that happened so I added an explicit `|| (Objects.equals(attribute.namespace, "") && namespace == null)` case for that. Again, not the prettiest but it has the advantage of being explicit and relatively easy to verify.

Finally, it occurred to me to write a test to get the value of a namespaced attribute on an instance element. That was not working either. That's because the namespace prefix was not being saved on attribute `TreeElement`s. I was tempted to get into some deeper refactoring here but didn't. I kept the existing method signatures and made my changes additive except for `setBindAttribute` which really feels like an internal implementation detail. I think there's some risk here that the fix is incomplete.

All intentional changes are bug fixes as described above and by the tests. I overall feel like this is additive and pretty low risk. That said, if people are using custom namespaces and attributes in their forms, I could imagine some changes in behavior or regressions. This also does touch some broadly-used things like bind attributes. There's also always a bit of risk when touching serialization.

See tests.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/instance/TreeElement.java
+++ b/src/main/java/org/javarosa/core/model/instance/TreeElement.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 
+import java.util.Objects;
 import org.javarosa.core.model.Constants;
 import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.FormElementStateListener;
@@ -36,6 +37,7 @@ import org.javarosa.core.model.instance.utils.DefaultAnswerResolver;
 import org.javarosa.core.model.instance.utils.IAnswerResolver;
 import org.javarosa.core.model.instance.utils.ITreeVisitor;
 import org.javarosa.core.model.instance.utils.TreeElementChildrenList;
+import org.javarosa.core.model.instance.utils.TreeElementNameComparator;
 import org.javarosa.core.model.util.restorable.RestoreUtils;
 import org.javarosa.core.util.DataUtil;
 import org.javarosa.core.util.externalizable.DeserializationException;
@@ -137,7 +139,7 @@ import org.jetbrains.annotations.Nullable;
     public static TreeElement constructAttributeElement(String namespace, String name, String value) {
         TreeElement element = new TreeElement(name);
         element.setIsAttribute(true);
-        element.namespace = (namespace == null) ? "" : namespace;
+        element.namespace = namespace;
         element.multiplicity = TreeReference.INDEX_ATTRIBUTE;
         element.value = new UncastData(value);
         return element;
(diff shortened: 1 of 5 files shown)


lognaturel:
We've decided to close this for now. After discussion and reflection, we're no longer seeing as much value to using namespaces in the instance. See https://github.com/XLSForm/pyxform/issues/105#issuecomment-1305977323 for related discussion.

PR #697, opened 2022-11-11
Update entities features for new spec

seadowg:
https://github.com/getodk/xforms-spec/pull/299

Closes #


lognaturel (approved):
The test with a dynamic create expression wasn't setting the value of `/data/join` correctly. Since it was only checking for the no creation case which was satisfied with no change in `/data/join`, it kind of was testing what it was supposed to. I think it's helpful to have both the create and non-create case for that test so I went ahead and added that.

PR #698, opened 2022-11-16
Add tests for automatically giving values to attributes

lognaturel:
This is a testing-only change.

We were surprised to discover that preloading attributes is not supported. I think it's helpful to have an ignored test documenting this. `setvalue` provides similar functionality that does work with attributes.

### DIFF
--- a/src/test/java/org/javarosa/core/model/utils/test/QuestionPreloaderTest.java
+++ b/src/test/java/org/javarosa/core/model/utils/test/QuestionPreloaderTest.java
@@ -1,0 +1,41 @@
+package org.javarosa.core.model.utils.test;
+
+import static org.hamcrest.CoreMatchers.startsWith;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.javarosa.core.util.BindBuilderXFormsElement.bind;
+import static org.javarosa.core.util.XFormsElement.body;
+import static org.javarosa.core.util.XFormsElement.head;
+import static org.javarosa.core.util.XFormsElement.html;
+import static org.javarosa.core.util.XFormsElement.input;
+import static org.javarosa.core.util.XFormsElement.mainInstance;
+import static org.javarosa.core.util.XFormsElement.model;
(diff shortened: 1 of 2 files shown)


src/test/java/org/javarosa/core/model/utils/test/QuestionPreloaderTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.javarosa.core.util.XFormsElement.title;
+
+import java.io.IOException;
+import org.javarosa.core.test.Scenario;
+import org.javarosa.xform.parse.XFormParser;
+import org.junit.Ignore;
+import org.junit.Test;
+
+public class QuestionPreloaderTest {
+    @Ignore("Not supported")
```
seadowg, line 22: I think I'd generally prefer a test that documents the current behaviour (with a comment to illustrate that it's maybe not ideal) instead of an ignored one that documents a possible future (or here maybe "expected" behaviour). This test is great as a piece of documentation, but it won't give us any warning/indication if there's a change in behaviour because of the `@Ignore`.

seadowg (changes requested):

src/test/java/org/javarosa/core/model/utils/test/QuestionPreloaderTest.java
```diff
@@ -15,13 +16,30 @@
...
+            ),
+            body(
+                input("/data/element")
+            )));
+
+        assertThat(scenario.answerOf("/data/element").getDisplayText(), startsWith("uuid:"));
+    }
+
+    @Test
+    public void preloader_doesNotpreloadAttributes() throws IOException, XFormParser.ParseException {
```
seadowg, line 42: I think it'd be good to have a comment here highlighting that we might want this to **not** be the case one day. Makes it clear this test is documentation rather than intention.

seadowg (changes requested):

seadowg (approved):

PR #704, opened 2023-01-06
Add test of relevance on calculates

lognaturel:
This is a pattern that we broke in Enketo a while ago. I want to make sure it's explicitly tested here because it is a bit surprising.


src/test/java/org/javarosa/core/model/TriggerableDagTest.java
```diff
@@ -530,6 +530,33 @@ public void non_relevant_node_values_are_always_null_regardless_of_their_actual_
         assertThat(scenario.answerOf("/data/some-field"), is(intAnswer(42)));
     }
 
+    // Users use relevance on calculates to ensure that calculations are only run when the values
```
seadowg, line 533: Is it important to make sure that the calculation isn't executed or just that the calculation isn't "answered"? I'm guessing it's the latter (as that's what this test is about), but the comment suggests it might the former.

lognaturel: Great and deep question. Now I'm no longer 100% sure of the intent. This led me to some strange docs: https://github.com/getodk/docs/pull/1547 I will come back to it.

lognaturel: I clarified the comment wording to make it clear it's about the node not exposing a value. I've also added tests below to verify that multiplication and addition with empty values result in empty values.

lognaturel:
@latin-panda @garethbowen Could one of you please check to make sure that there are tests in Web Forms around this. It's connected to some questions about empty values that we've had recently.

latin-panda:
@lognaturel I've ported the missing tests to WF (https://github.com/getodk/central-frontend/pull/1916)

PR #707, opened 2023-01-12
Add LineString and Polygon parsing

### LINKED ISSUE #706: Add support for additional GeoJSON geometry object types
Support for the following [types](https://www.rfc-editor.org/rfc/rfc7946#section-3.1) should be added:
- [x] `LineString` - `coordinates` positions should be translated to `;` geometry string
- [x] `MultiPoint` - `coordinates` positions should be translated to `;` geometry string
- [x] `Polygon` - first linear ring in `coordinates` should be translated to `;` geometry string (any others should be ignored)

seadowg:
Closes #706

New tests.

Comments inline.

This is all additional, so not a lot of risk here, hopefully.

https://github.com/getodk/docs/issues/1541

### DIFF
--- a/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
+++ b/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
@@ -32,7 +32,7 @@ public class GeojsonGeometry {
     public String getOdkCoordinates() throws IOException {
         if (type.equals("Point")) {
             return coordinates.get(1) + " " + coordinates.get(0) + " 0 0";
-        } else if (type.equals("LineString") || type.equals("MultiPoint")) {
+        } else if (type.equals("LineString")) {
             StringJoiner stringJoiner = new StringJoiner("; ");
             for (Object item : coordinates) {
                 List<Object> point = (List<Object>) item;
@@ -54,7 +54,7 @@ public class GeojsonGeometry {
             }
 
         } else {
-            throw new IOException("Only Points, LineStrings, Polygons and MultiPoints are currently supported");
+            throw new IOException("Only Points, LineStrings and Polygons are currently supported");
         }
     }
 }
--- a/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
+++ b/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
@@ -32,7 +32,7 @@ public class GeojsonGeometry {
     public String getOdkCoordinates() throws IOException {
         if (type.equals("Point")) {
             return coordinates.get(1) + " " + coordinates.get(0) + " 0 0";
-        } else if (type.equals("LineString")) {
+        } else if (type.equals("LineString") || type.equals("MultiPoint")) {
             StringJoiner stringJoiner = new StringJoiner("; ");
             for (Object item : coordinates) {
                 List<Object> point = (List<Object>) item;
@@ -41,7 +41,7 @@ public class GeojsonGeometry {
 
             return stringJoiner.toString();
         } else {
-            throw new IOException("Only Points, LineString are currently supported");
+            throw new IOException("Only Points, LineStrings and MultiPoints are currently supported");
         }
     }
 }
--- a/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
+++ b/src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
@@ -41,13 +41,18 @@ public class GeojsonGeometry {
 
             return stringJoiner.toString();
         } else if (type.equals("Polygon")) {
-            StringJoiner stringJoiner = new StringJoiner("; ");
-            for (Object item : (List<Object>) coordinates.get(0)) {
-                List<Object> point = (List<Object>) item;
-                stringJoiner.add(point.get(1) + " " + point.get(0) + " 0 0");
+            if (!coordinates.isEmpty()) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
```diff
@@ -17,23 +17,44 @@
...
 @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
 public class GeojsonGeometry {
-    private String type;
-    private ArrayList<String> coordinates;
 
-    public String getType() {
-        return type;
-    }
+    private String type;
+    private ArrayList<Object> coordinates;
```
seadowg, line 30: I don't think there's a better way to parse properties that can have multiple types, unfortunately.

src/test/java/org/javarosa/core/model/instance/geojson/GeojsonGeometryTest.java
```diff
@@ -0,0 +1,37 @@
+package org.javarosa.core.model.instance.geojson;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class GeojsonGeometryTest {
```
seadowg, line 9: We could also add a test (and implementation) for validating that Polygons always have identical first and last points. I'm not sure that that makes sense however: Collect will have to dynamically detect a polygon by checking for a match anyway, so invalid polygons would just end up being represented as traces.

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
```diff
@@ -32,7 +32,7 @@ public class GeojsonGeometry {
     public String getOdkCoordinates() throws IOException {
         if (type.equals("Point")) {
             return coordinates.get(1) + " " + coordinates.get(0) + " 0 0";
-        } else if (type.equals("LineString")) {
+        } else if (type.equals("LineString") || type.equals("MultiPoint")) {
```
lognaturel, line 35: It occurs to me that `MultiPoint` is semantically distinct from `LineString` and we probably shouldn't support it (yet). It's intended to be displayed without the lines in between so we don't have a way of representing that in the ODK geo notation.

src/test/java/org/javarosa/core/model/instance/geojson/GeojsonGeometryTest.java
lognaturel: Does Collect check that first and last match before dropping last? If not, it probably should.

lognaturel (changes requested):
Unfortunately I think we should omit `MultiPoint` for now. Everything else looks good to me.

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
seadowg: > It's intended to be displayed without the lines in between so we don't have a way of representing that in the ODK geo notation.

It was pretty unclear to me in the spec whether that was the case or not. I guess it's an array of points that can be empty or just one point so that implies no "line". Is the "0-dimensional Point and MultiPoint" comment what suggests that to you or is there another part that lays that out in a clearer way?

Either way, part of me feels like there wouldn't be a problem in showing it with a line now and enhancing it later? Maybe there are cases where the line would be problematic that I haven't thought about.

src/test/java/org/javarosa/core/model/instance/geojson/GeojsonGeometryTest.java
seadowg: There's no implementation for showing shapes (it's a separate issue from showing traces/lines), but yes it'll need to do that.

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
lognaturel: Agreed the spec is not super clear but it was seeing the name again in context that made me realize the difference in intent. The Internet agrees: https://stackoverflow.com/questions/36685195/whats-the-difference-between-linestring-and-multipoint-in-geojson

> wouldn't be a problem in showing it with a line now and enhancing it later

It just doesn't make sense. Let's say the feature is "nurse stations" and there's a multipoint geometry. Drawing it with lines would not be useful I think it's more helpful to users to let them know we don't support what they're trying to do. In that case, they should disaggregate into one point per station.

That said, if you really feel strongly about keeping it in, this is not the hill I want to die on!

src/test/java/org/javarosa/core/model/instance/geojson/GeojsonGeometryTest.java
lognaturel: There is an implementation that shows existing shapes (defaults or previously-selected). But anyway, sounds like that's on your radar and you'll consolidate all that anyway.

src/main/java/org/javarosa/core/model/instance/geojson/GeojsonGeometry.java
seadowg: > It just doesn't make sense. Let's say the feature is "nurse stations" and there's a multipoint geometry. Drawing it with lines would not be useful I think it's more helpful to users to let them know we don't support what they're trying to do. In that case, they should disaggregate into one point per station.

That makes a lot of sense. I'll remove the support here.

lognaturel (approved):

PR #708, opened 2023-02-10
Add base64-decode

lognaturel:
Discussion at https://forum.getodk.org/t/proposal-add-centroid-and-base64binary-to-string-functions/40289/5

Added tests.

Name was selected with community input. Since we're not following an existing spec, I decided not to add an encoding parameter. In practice, most things are UTF-8 these days. We can add an encoding parameter later if needed.

We had talked about throwing an exception in case of invalid input but I decided to use a blank/null value instead. The problem with throwing an exception is that the exception can manifest as a Validate error and/or be thrown at undesirable moment if the input is not complete yet. With a blank value, it may be harder to track down unexpected values but I think it's an overall better experience.

I wanted to use standard Java `Base64` but that's not in Android until API 26. We already have a precedent for pulling code in from https://github.com/brsanthu/migbase64/blob/master/src/main/java/com/migcomponents/migbase64/Base64.java so I continued that.

Isolated addition so all risk is to the new code which could be wrong.

See test.

https://github.com/getodk/xforms-spec/pull/301

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/Encoding.java
+++ b/src/main/java/org/javarosa/xpath/expr/Encoding.java
@@ -22,75 +22,75 @@ import org.javarosa.xpath.XPathUnsupportedException;
  * Implements the hash encoding methods for XPathFuncExpr digest() function
  */
 enum Encoding {
-  HEX("hex") {
-    @Override
-    String encode(byte[] bytes) {
-      StringBuilder sb = new StringBuilder(bytes.length * 2);
-      for (byte b : bytes) {
-        sb.append(HEX_TBL[(b >> 4) & 0xF]);
-        sb.append(HEX_TBL[(b & 0xF)]);
-      }
-      return sb.toString();
+    HEX("hex") {
+        @Override
+        String encode(byte[] bytes) {
+            StringBuilder sb = new StringBuilder(bytes.length * 2);
+            for (byte b : bytes) {
+                sb.append(HEX_TBL[(b >> 4) & 0xF]);
+                sb.append(HEX_TBL[(b & 0xF)]);
+            }
+            return sb.toString();
+        }
+    },
+    BASE64("base64") {
+        @Override
+        String encode(byte[] sArr) {
+            // Copied from: https://github.com/brsanthu/migbase64/blob/master/src/main/java/com/migcomponents/migbase64/Base64.java
+            // Irrelevant after Java8
+            int sLen = sArr.length;
+            int sOff = 0;
+
+            if (sLen == 0)
+                return "";
+
+            int eLen = (sLen / 3) * 3;
+            int dLen = ((sLen - 1) / 3 + 1) << 2;
+            byte[] dArr = new byte[dLen];
+
+            for (int s = sOff, d = 0; s < sOff + eLen; ) {
+                int i = (sArr[s++] & 0xff) << 16 | (sArr[s++] & 0xff) << 8 | (sArr[s++] & 0xff);
+                dArr[d++] = (byte) BASE_64_TBL[(i >>> 18) & 0x3f];
+                dArr[d++] = (byte) BASE_64_TBL[(i >>> 12) & 0x3f];
+                dArr[d++] = (byte) BASE_64_TBL[(i >>> 6) & 0x3f];
+                dArr[d++] = (byte) BASE_64_TBL[i & 0x3f];
+            }
+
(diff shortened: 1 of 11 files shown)


src/test/java/org/javarosa/xpath/expr/XPathFuncExprBase64DecodeTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import org.javarosa.core.test.Scenario;
+import org.javarosa.xform.parse.XFormParser;
+import org.junit.Test;
+import org.junit.experimental.runners.Enclosed;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Enclosed.class)
```
lognaturel, line 26: I wanted to have both valid and invalid input cases in the same file. This is the cleanest way I found to do that.

seadowg (changes requested):

src/main/java/org/javarosa/xpath/expr/Encoding.java
```diff
@@ -15,82 +15,139 @@
...
+                    + Character.digit(bytes[i+1], 16));
+            }
+            return data;
+        }
+    },
+    BASE64("base64") {
+        // Copied from: https://github.com/brsanthu/migbase64/blob/master/src/main/java/com/migcomponents/migbase64/Base64.java
+        // Irrelevant after Java8
+        @Override
+        String encode(byte[] sArr) {
```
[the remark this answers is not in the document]
lognaturel: This is only a whitespace change to make indentation consistent with rest of project. `encode` was introduced and tested in https://github.com/getodk/javarosa/pull/247.

src/test/java/org/javarosa/xpath/expr/EncodingDecodeTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.junit.Assert.assertEquals;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Parameterized.class)
+public class EncodingDecodeTest {
```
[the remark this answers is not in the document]
lognaturel: I have a pretty strong preference to test at both levels because it's a pattern that exists throughout JR. There have been cases of functions that work when tested in isolation like this but not in forms. It's helpful to have the test coverage to differentiate the two, especially when there are regressions.

In this specific case, I further like the symmetry with `EncodingEncodeTest`.

src/test/java/org/javarosa/xpath/expr/XPathFuncExprBase64DecodeTest.java
lognaturel: I agree with you in some cases but I think it really depends on the shape of the tests. With this style of test where there's a standard form with different values subbed in, I find parameterized tests easier to work with. We have a lot of examples of `Parameterized` tests in JR.

That said, I'm ok moving to lowercase-p parameterized tests. I've done that in the latest commit. I think the important thing is that the form is only defined once. Hopefully this is a reasonable compromise for you.

src/main/java/org/javarosa/xpath/expr/Encoding.java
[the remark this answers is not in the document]
seadowg: Ah, sorry I got confused. I thought the whitespace change was just for the `HEX` stuff, but I see now that you only added `decode` to this file.

src/test/java/org/javarosa/xpath/expr/XPathFuncExprBase64DecodeTest.java
seadowg: I much prefer the new approach. I find using common builders/helpers a much nicer approach than an entire parametrized test, as you're still able to have a unique procedure for each case (without branching).

seadowg (approved):

src/test/java/org/javarosa/xpath/expr/XPathFuncExprBase64DecodeTest.java
lognaturel: Makes sense! We can move towards that as we introduce new tests.

PR #709, opened 2023-02-21
Upgrade Jackson

seadowg:
Upgrade Jackson to fix https://github.com/FasterXML/jackson-databind/issues/3412

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -135,10 +135,10 @@ if (!project.hasProperty("android")) {
     android {
         namespace 'com.example'
 
-        compileSdkVersion 30
+        compileSdkVersion 33
         defaultConfig {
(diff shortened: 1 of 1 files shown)


lognaturel (dismissed):

lognaturel:
Sigh, looks like your instinct to double check was right, @seadowg, this doesn't work. And actually now I see https://github.com/FasterXML/jackson-databind/issues/3412 does not suggest it would. There was some additional use of `getParameterCount()` introduced in .2 and removed in .3. What it looks like to me is that before https://github.com/getodk/javarosa/pull/707 (when we just parsed points), we avoided any functionality that resulted in that method being called. Now we use it. I can't yet spot the change.

PR #716, opened 2023-05-15
Downgrade Commons IO to prevent Android compatibility issues

seadowg:
Closes #

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -162,7 +162,9 @@ dependencies {
 
     // Upgrade to version higher than 1.4 when Collect minSDK >= 26
     implementation 'org.apache.commons:commons-csv:1.4'
-    implementation 'commons-io:commons-io:2.11.0'
+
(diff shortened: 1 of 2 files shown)


seadowg:
@getodk/testers do we test with a device with Android 7.1.1 or lower as part of regression? Just want to check whether this is a gap in our testing or not.

srujner:
@seadowg We used to have a single steps for a 5.1 in some test cases ~10, but we stopped checking it as we're mainly focusing on Android 10 and 13 and we have 8.1 for PRs

lognaturel (approved):

lognaturel:
The Android version checker is supposed to help us with things like this but it doesn’t because it’s a runtime exception!

I don’t think we need to change our process. I’ll continue using order devices and if we don’t catch these kinds of problems before releasing we’ll see them in logs.

PR #718, opened 2023-06-16
Updating codebase to use java.time.* instead of joda-date for internal interpretation of dates

jack-bolles:
Closes # n/a

All tests pass, more have been added

The java world has moved on from how Dates and Time should be addressed. the writers of joda-time defer to the java.time implementations.

#### How does this change affect users? Describe intentional changes to behavior and behavior that joda was used internally, with java.util.Date the object of record. That still holds true, as java.time clases are used "under the hood". 

One of the formats referred to in the code was not referenced anywhere, so was not migrated. This should be an area for the PR reviewer to look deeper. See doc/architecture/decisions/0003-limit-date-timeformats.md for more information

No

Documentation in github has been kept up to date, even enhanced. Documentation elsewhere has not. These are internal changes and are unlikely to need further documentation.


lognaturel:
Thanks very much for taking on this maintenance!

Could you please provide a bit of context on your usage of JavaRosa? Do you use it as part of Collect or another system? This will be one valuable piece of context as we consider how to review this to make sure we minimize risk for users.

It would also help to understand how you chose this particular area to work in. (E.g. are you trying to reduce dependencies in a downstream tool, do you just have recent experience with this migration in another context, etc.)

jack-bolles:
_**re "This will be one valuable piece of context as we consider how to review this to make sure we minimize risk for users."**_
A bit more context.... I believe each commit is a discrete piece of work and I ran the full test suite before every commit. I could have pushed to prod at anytime. Reviewing commit by commit makes much more sense than file by file. 

Outside of removing deprecation and other lint-like updates, the area impacted is well-contained and covered well by both existing tests and new ones added. Any gaps in testing probably would benefit the current javarosa codebase regardless of whether you accept this PR or not. 

What I don't have and would be useful to setup is a client for manual testing.

lognaturel:
Thanks so much for providing some background and for jumping in. I’m not entirely sure yet when we’ll be able to review but will try to do so in a timely manner!

lognaturel:
A quick nuts and bolts questions. We do have a lint check to build in an Android project with min API 21 so that gives reasonable confidence that at least the classes are available. I think it’s using desugaring, right? I’m not sure whether that’s something we already make use of (@seadowg?) and if not we’ll need to consider the APK size impact.

seadowg:
> I think it’s using desugaring, right? I’m not sure whether that’s something we already make use of in Collect (@seadowg?) and if not we’ll need to consider the APK size impact (unless unused classes are now excluded).

Not sure I understand the question here. What's using desugaring?

lognaturel:
Android added java.time as part of Android 26 (https://developer.android.com/reference/java/time/package-summary). So I assume this code works in Android through API desugaring (https://developer.android.com/studio/write/java8-support-table).

seadowg:
Ah right I understand. Yes JavaRosa enabled desugaring for it's Android checks and requires clients also enable it (Collect does).

seadowg (changes requested):
I've noticed that this PR introduces [Architecture Decision Records](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions.html). This approach hasn't been used in JavaRosa or any other ODK project, so it's quite a significant addition. Generally decisions made as part of a PR are documented in the PR description, through commit messages and also through Github's inline code comments. So we're not discussing ADRs **and** the switch away from jodatime, could you please remove the ADR changes and add equivalent discussions to the PR description (or inline comments if that feels more appropriate)?

If you feel strongly that we should be using ADRs (and would prefer to use it here), it would be better to create another PR (with just `0001-record-architecture-decision.md` for instance) that we can discuss/review in isolation.

lognaturel:
> JavaRosa enabled desugaring for it's Android checks and requires clients also enable it (Collect does).

Got it, thanks! So that might be part of the answer to our question about significant increases in APK size, right? Good to know this PR makes no change on that front

jack-bolles:
Happy to drop ADRs. Force of habit, as you may be able to tell by the above ADR 0000.

The decisions themselves are already recorded in the PR's description and the commit messages. The ADRs are there to give them context.

lognaturel:
I've started reviewing. I'm going to go commit by commit and leave notes in this comment as I go. I'm particularly looking for user-facing risk and changes that don't appear related to the time migration.

- https://github.com/getodk/javarosa/pull/718/commits/1f894843de7218742104bb0ab36e05c5bb806108 looks to me like a candidate for splitting off. It will be tricky for clients to adapt to the change to `ParseException`. There's a mix of formatting changes and substantive changes that makes it feel risky.
-

lognaturel:
Thanks again for taking this on, @jack-bolles!

This PR doesn't match the norms of this project so I've been trying to figure out how to approach it. In particular, you'll see looking back in history that it's uncommon for a PR to have more than 10-12 commits and that things like public API changes are typically considered in isolation. You can also read about some of this in [the contribution guide](https://github.com/getodk/javarosa/blob/master/CONTRIBUTING.md). That doesn't mean your way is bad, it's just different! The current maintainers are particularly risk-averse because we weren't there when much of the library was written and because the negative impact of regressions can be quite high.

Here are some high level thoughts:
* I don't think the change from joda to java.time should trigger a major version change because it's purely internal. hat means we shouldn't bundle public API changes with it.
* I'd like to see more analysis of risk specifically around the migration from joda to java.time. Are there known differences in precision? Null handling? Anything that could have inadvertent user-facing consequences? Any part of date/time code that might be lacking in test coverage? We typically use either an issue or the PR description for this kind of context.
* I think that if the switch in library were isolated and maybe split up into a few more explanatory commits it wouldn't look so risky. It's all the changes in API and commits that mix stylistic changes with substantive ones that make this feel really hard to review.

One possible next step would be to cherry pick some of the commits that are easy to evaluate as safe onto a new branch so we could consider those independently and hopefully quickly get to a merge. I kept a list above. Changes to Java 8, removing stale comments, etc, all seem valuable and relatively straightforward to confirm as safe.

Then I'm not totally sure. I think before investing further in the joda to java.time update it would be worth looking into some of my questions about risk above. The goal would be to get a sense of what edge cases might exist and backfill tests specifically for those. It might be good to do this in an issue. If there aren't very many areas of divergence and the existing tests look like they are likely to cover real usage in user forms, then it feels worth proceeding. @jack-bolles you'd have to consider what the best way to get to a really narrow PR would be and whether that feels like something you want to do. Cherry-picking commits onto a new branch might work well though I did notice a couple of big commits that also included stylistic changes. You might be able to split them if you've got a git environment you're comfortable in.

One thing that caught my eye is that you're interested in Kotlin. We have talked about moving to introducing new functionality here in Kotlin as we have with https://github.com/getodk/collect. So far the barrier has been getting the tooling working. If you're interested in helping out with that, @seadowg would have more details for you.

jack-bolles:
Thanks for your thoughtful comments, @lognaturel. I think the best approach, at least for me, is to treat this as a spike, put it to the side and start anew. 

One thing I've learned is that contributing as a 3rd party is much different than normal working with code you share ownership with. My natural tendency to only refactor (the 'tidy' commits) code one files you touch but to be thorough about it in that radius, rather than making a singular-but-codebase-wide change (e.g. one commit with all the java5 langauge updates) goes against the grain of OS contributions.

Using the philosophy of make things easy to change, _then_ make the change, I think the best approach is to do the code clean up - remove redundant commits, use features of Java 8, delete unused code if and only if behind a private accessor - then look to other things

I think coming up with a deprecation strategy makes sense for the project. I have an idea for that and can raise it in the forum.

jack-bolles:
I'll ask @seadowg about how I can help with tooling. I don't think that's the biggest issue with moving javarosa to Kotlin. The code has been done in a style that goes against the grain of Kotlin. We've all heard the adage that _a good C developer can write good C in any language_. Porting Java to Kotlin without taking into account immutability, null-safety, stronger type-safety,... will make the codebase worse (more verbose and harder to reason about) than better. at least IME. This feels like a longer discussion, and I'm happy to have it in slack or in the forum.

seadowg:
> I'll ask @seadowg about how I can help with tooling. I don't think that's the biggest issue with moving javarosa to Kotlin. The code has been done in a style that goes against the grain of Kotlin. We've all heard the adage that a good C developer can write good C in any language. Porting Java to Kotlin without taking into account immutability, null-safety, stronger type-safety,... will make the codebase worse (more verbose and harder to reason about) than better. at least IME. This feels like a longer discussion, and I'm happy to have it in slack or in the forum.

More than happy to chat about this! Agreed it makes sense to happen somewhere else. There was a good discussion on [the forum](https://forum.getodk.org/t/using-kotlin-for-odk-android-apps/18367/13) a while back for using Kotlin in the Java parts of the ODK ecosystem that we maintain (Collect and JavaRosa mainly). Hit me up on Slack (@seadowg there as well) and we can discuss complexities I've run into with introducing Kotlin into JavaRosa and we can try and identify some code that might benefit from it.

lognaturel:
> treat this as a spike, put it to the side and start anew.

This is a pattern the core team uses quite a bit: spike something out, evaluate risk and benefit of the resulting state, discuss what the focus of the change(s) should be, create and review small PRs. So if you're ok working that way too, I think we're generally very happy with that.

> One thing I've learned is that contributing as a 3rd party is much different than normal working with code you share ownership with

Yes, that is so very true. And I'd add that each code base has its own challenges and opportunities so even across the systems the ODK team maintains you'll find there are subtly different approaches at plan.

> My natural tendency to only refactor (the 'tidy' commits) code one files you touch but to be thorough about it in that radius, rather than making a singular-but-codebase-wide change (e.g. one commit with all the java5 langauge updates) goes against the grain of OS contributions.

I wouldn't necessarily say that! I think it depends on the nature of the changes, the state of the codebase at the time, maintainer availability and other factors. In general I'd say we welcome tidy commits in preparation for substantive work. It's very helpful to have those separated out. Ideally there's also a separation between changes that are very easily verifiable as equivalent (some of the ones I indicated as easy merges above) vs. changes that require understanding to verify (e.g. need to consider possibility of new null cases or something like that). Something smaller and more focused is always easier to review and get confidence in.

> javarosa's TimeData, DateData, and DateTimeData, each referenced in Collect. 

Aha, thanks, that's really useful to point out, I hadn't considered that part of the conversion. Currently all types available to forms are represented as `IAnswerData` classes in JavaRosa. I don't think we would consider changing that unless we were doing something like removing the caching that Collect/clients use to get faster form load (see `readExternal` and `writeExternal`). It would be helpful to get your take on why those could/should change.

jack-bolles:
> It would be helpful to get your take on why those could/should change.

In no particular order...

- Collect already roundtrips between `java.util.Date` and `joda-time` every time. see [`DateTimeWidget`](https://github.com/getodk/collect/blame/master/collect_app/src/main/java/org/odk/collect/android/widgets/DateTimeWidget.java). I suspect others using the library will have done something similar.
- `jodatime` is no longer being actively developed. The author wants everyone to move to `java.time`
- while I don't think `java.util.Date` will every be truly "deprecated", it won't be the basis for further enhancements or capbilities. eg a native android date picker or time picker won't use it. (In fact, the one that came out for Android 7 doesn't)
- There's a reasonable argument to be made that un/serialising should use string implementations rather than serialising objects (maybe that's the idea behing un/casting?). Generally that would allow for a smaller footprint (strings zip up smaller); separate storage from implementation, making it easier for a something 'serialised' under version x to be deserialised by version x++; allow interchange between ODK-friendly applications that aren't written on the jvm possible 
- A `java.util.Date` without further context can represent one of three types - `DateTimeData, DateData, TimeData`. They are not interchangeable. Moving to java.time equivalents are more precise and more accurate representations and c/would simplify the code that has to deduce and/or evaluate context.

lognaturel:
These arguments are for changing the internal representation of `DateTimeData`, `DateData`, `TimeData` from `java.util.Date` and therefore changing the return type of their `getValue` methods, right? That feels like it can be considered separately from replacing jodatime because the consequences are pretty different. Similarly, replacing the use of jodatime in Collect feels like a separate issue (that can be taken on at any time).

My current instinct is that it won't be worth the disruption of changing the return types of the `getValue` methods but maybe that can be filed as an issue for separate consideration and discussion.

jack-bolles:
Actually, I just re-named it. If you ever want to re-vist it, the code is here: https://github.com/jack-bolles/javarosa/tree/PR1-remove-joda-time

PR #719, opened 2023-06-28
Add ability to search expressions for functions and to questions

seadowg:
Work towards getodk/collect#5620

This change allows a client add a processor to `XFormParser` that would allow it to detect the presence of particular functions or properties of a question:

```kotlin
xFormParser.addProcessor(object : BindAttributeProcessor, QuestionProcessor {
    override fun getBindAttributes(): Set<Pair<String, String>> {
        return setOf(Pair("", "calculate"))
    }

    override fun processBindAttribute(name: String, value: String, binding: DataBinding) {
        val calculate = binding.calculate
        val containsFunc = calculate.expr.expr.containsFunc("myFunc")
    }

    override fun processQuestion(question: Question) {
        val containsAppearance = question.appearanceAttr.contains("myCustomAppearance")
    }
})
```

This will let us detect forms that do/don't contain `pulldata` or `search`.

I've also added a convenience implementation of a wrapper `IXFormParserFactory` (which will make Collect's code cleaner).

New tests.

We could have used the existing `BindAttributeProcessor`, but that would have meant parsing `calculate` in the processor which didn't feel ideal.

No user facing changes as this is opt in.

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -64,6 +64,7 @@ import org.javarosa.xpath.expr.XPathNumericLiteral;
 import org.javarosa.xpath.expr.XPathPathExpr;
 import org.javarosa.xpath.expr.XPathStringLiteral;
 import org.javarosa.xpath.parser.XPathSyntaxException;
+import org.jetbrains.annotations.NotNull;
 import org.kxml2.io.KXmlParser;
 import org.kxml2.kdom.Document;
 import org.kxml2.kdom.Element;
@@ -178,6 +179,7 @@ public class XFormParser implements IXFormParserFunctions {
     private final List<BindAttributeProcessor> bindAttributeProcessors = new ArrayList<>();
     private final List<FormDefProcessor> formDefProcessors = new ArrayList<>();
     private final List<ModelAttributeProcessor> modelAttributeProcessors = new ArrayList<>();
+    private final List<QuestionProcessor> questionProcessors = new ArrayList<>();
 
     /**
      * The string IDs of all instances that are referenced in a instance() function call in the primary instance
@@ -428,6 +430,10 @@ public class XFormParser implements IXFormParserFunctions {
 
         if (processor instanceof ModelAttributeProcessor) {
             addModelAttributeProcessor((ModelAttributeProcessor) processor);
+        }
+
+        if (processor instanceof QuestionProcessor) {
+            questionProcessors.add((QuestionProcessor) processor);
         }
     }
 
@@ -1172,6 +1178,7 @@ public class XFormParser implements IXFormParserFunctions {
 
         processAdditionalAttributes(question, e, usedAtts, passedThroughAtts);
 
+        questionProcessors.stream().forEach(questionProcessor -> questionProcessor.processQuestion(question));
         return question;
     }
 
@@ -2447,6 +2454,10 @@ public class XFormParser implements IXFormParserFunctions {
         Set<Pair<String, String>> getModelAttributes();
 
         void processModelAttribute(String name, String value) throws ParseException;
+    }
+
+    public interface QuestionProcessor extends Processor {
+        void processQuestion(@NotNull QuestionDef question);
     }
 
     public static class ParseException extends Exception {
--- a/src/main/java/org/javarosa/core/model/condition/IConditionExpr.java
+++ b/src/main/java/org/javarosa/core/model/condition/IConditionExpr.java
@@ -23,6 +23,7 @@ import org.javarosa.core.model.condition.pivot.UnpivotableExpressionException;
 import org.javarosa.core.model.instance.DataInstance;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.core.util.externalizable.Externalizable;
+import org.javarosa.xpath.expr.XPathExpression;
 
 /**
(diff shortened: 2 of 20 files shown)


seadowg:
Marking this as a draft for the moment. I think we'll also need a way to detect `search` in appearances.

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2435,6 +2458,14 @@ public interface FormDefProcessor extends Processor {
...
     }
 
+    public interface BindProcessor extends Processor {
+        void processBind(@NotNull DataBinding binding);
+    }
+
+    /**
+     * @deprecated Use {@link BindProcessor} instead
+     */
+    @Deprecated
```
seadowg, line 2468: We could potentially just remove this, but given someone could theoretically be using it, we'd need to a major version bump.

src/test/java/org/javarosa/entities/EntitiesTest.java
```diff
@@ -313,35 +311,4 @@ public void whenSaveToQuestionIsNotAnswered_entityPropertyIsEmptyString() throws
         assertThat(entities.size(), equalTo(1));
         assertThat(entities.get(0).properties, equalTo(asList(new Pair<>("name", ""))));
     }
-
-    @Test
-    public void savetoIsRemovedFromBindAttributesForClients() throws IOException, XFormParser.ParseException {
```
seadowg, line 318: I don't know why this was important? Happy to be corrected, but I don't see any reason to not allow accessing `saveto` from `DataBinding`.

lognaturel: Looks like it was introduced at the very beginning in https://github.com/getodk/javarosa/pull/691. I don't see a specific explanation but in my head it makes sense to want confidence that entity directives are handled by the JR processor and not available for clients to muck with. I see it as a nice way to enforce hygiene which is something that Collect has had trouble with in the past.

src/main/java/org/javarosa/xform/parse/XFormParser.java
lognaturel: I'm having trouble understanding the `BindAttributeProcessor` -> `BindProcessor` change. Currently `BindAttributeProcessor` makes it possible to create a processor for a specific bind attribute and namespace. It feels like creating a processor for `calculate` in the same way would be fine and good. I'm not seeing any advantage to the `BindProcessor` approach.

lognaturel:
> We could have used the existing BindAttributeProcessor, but that would have meant parsing calculate in the processor which didn't feel ideal.

As I wrote in a comment, I'm not understanding this. `processBindAttribute` takes in a `DataBinding`. The only real difference with the new `BindProcessor` that I can see is that `BindAttributeProcessor` specifies which attribute it cares about. Maybe it feels weird to iterate over all the attributes? But I think the alternative for attributes like `saveto` or the recording ones would be to iterate through `TreeElement`s (`additionalAttrs`) which feels worse.

src/main/java/org/javarosa/xform/parse/XFormParser.java
seadowg: I'd written a comment in the PR description about this:

> We could have used the existing BindAttributeProcessor, but that would have meant parsing calculate in the processor which didn't feel ideal.

Does that make sense? Happy to discuss further. I actually couldn't remember why I made this choice until I read that!

src/test/java/org/javarosa/entities/EntitiesTest.java
seadowg: That's fair. I'm not feeling as concerned about that now. Introducing these processors has given us a structured way to interact with form parsing, and my gut reaction is that removing attributes feels overly protective.

lognaturel: Sounds reasonable to me!

src/main/java/org/javarosa/xform/parse/XFormParser.java
lognaturel: Is this saying that the `value` for the `calculate` attribute is raw so it would require parsing? I think it would be ok for a `BindAttributeProcessor` to use the `DataBinding` that it's given to access the parsed `calculate` and ignore the `value`. Maybe `BindAttributeProcessor` could get a `processBindAttribute(String name, DataBinding binding)` method that's preferred for attributes that are already parsed into the `DataBinding`?

seadowg: > Is this saying that the value for the calculate attribute is raw so it would require parsing?

Yeah exactly. The `BindAttributeProcessor` would have to parse the contents of `calculate` (which would just be a string).

> I think it would be ok for a BindAttributeProcessor to use the DataBinding that it's given to access the parsed calculate and ignore the value. Maybe BindAttributeProcessor could get a processBindAttribute(String name, DataBinding binding) method that's preferred for attributes that are already parsed into the DataBinding?

Ah I guess it could! I think I still prefer having a `BindProcessor` as it just feels simpler to me. You hook into every bind and can make your decision about what you want to do rather than having to implement (and understand) `getBindAttributes` and `processBindAttribute`. Are you just feeling that the change isn't worth the disruption?

seadowg: I'll investigate implementing this with `BindAttributeProcessor`.

lognaturel (approved):

PR #729, opened 2023-08-25
Fix custom function handler support

seadowg:
This had been broken by #722.

New and existing tests.

I'd really like to deprecate `FormDef#getEvaluationContext` as it really doesn't feel like it belongs there. I might do an afternoon spike on that and see how far I can get, but it's definitely too much work to hold up a fix for.

This basically reverts the behavior to how it was pre #722 so this is hopefully fairly low risk. That said, changes to this are are pretty scary and can have subtle side effects.

### DIFF
--- a/src/test/java/org/javarosa/core/test/Scenario.java
+++ b/src/test/java/org/javarosa/core/test/Scenario.java
@@ -57,6 +57,7 @@ import org.javarosa.xpath.expr.XPathNumNegExpr;
 import org.javarosa.xpath.expr.XPathNumericLiteral;
 import org.javarosa.xpath.expr.XPathPathExpr;
 import org.javarosa.xpath.parser.XPathSyntaxException;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


src/test/java/org/javarosa/core/test/Scenario.java
```diff
@@ -438,6 +439,14 @@ public static Scenario init(String formName, XFormsElement form) throws IOExcept
         return Scenario.init(formFile);
     }
 
+    public static FormDef createFormDef(String formName, XFormsElement form) throws IOException, XFormParser.ParseException {
+        Path formFile = createTempDirectory("javarosa").resolve(formName + ".xml");
+        String xml = form.asXml();
+        System.out.println(xml);
```
lognaturel, line 445: We should probably clean these up at some point (it's also in the source `init` implementation)

lognaturel (approved):
I'm a bit unnerved by the fact that we didn't catch this in initial dev but I don't think there's any real process change to consider. It is good that Collect tests caught it.

PR #732, opened 2023-10-02
Fixed gradle deprecated features

grzesiek2010:
I have run `./gradlew installLocal`

I've just fixed deprecated features. 

It doesn't require testing.

No.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,15 +79,14 @@ tasks.register('installLocal') {
 
 if (!project.hasProperty("android")) {
     jar {
-        baseName = 'javarosa'
         // Be sure to update version in pom.xml to match
         // snapshot release = x.x.x-SNAPSHOT
         // production release = x.x.x
-        version = '4.3.0-SNAPSHOT'
-        archiveName = baseName + '-' + version + '.jar'
+        archiveVersion = '4.3.0-SNAPSHOT'
+        archiveBaseName = "javarosa"
 
         manifest {
-            attributes 'Manifest-Version': "$jar.version"
+            attributes 'Manifest-Version': "$jar.archiveVersion"
         }
     }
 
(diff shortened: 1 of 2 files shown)


grzesiek2010:
`installLocall` still fails for me if I don't remove `maven-javadoc-plugin` with:
> MavenReportException: Error while generating Javadoc: Unable to find javadoc command: The environment variable JAVA_HOME is not correctly set.

but it seem like there is everything fine with `JAVA_HOME`... does anyone else experience the same?

seadowg:
> but it seem like there is everything fine with JAVA_HOME... does anyone else experience the same?

Is this on Linux? Are you able to just run `javadoc` (or `which javadoc`) from the javarosa project dir? Also, what does `echo $JAVA_HOME` output?

grzesiek2010:
> Is this on Linux?

Yes.

`./gradlew javadoc` works fine for example. It used to fail for the same reason `JAVA_HOME is not correctly set` because it wasn't set indeed but I've fixed it and it works well. But `./gradlew installLocall` does not. 
`echo $JAVA_HOME ` says `/usr/lib/jvm/java-17-openjdk-amd64` I've tried some other configurations and to no avail. If it works for `./gradlew javadoc` it should also work for `./gradlew installLocall` I think.

seadowg:
> ./gradlew javadoc works fine for example.

I just mean literally `javadoc` on it's own. It looks like maven is having a hard time finding the Java Doc binary. `installLocal` uses a maven build, not a Gradle one so it might be that the maven build process is not working on your machine.

grzesiek2010:
> I just mean literally javadoc on it's own. It looks like maven is having a hard time finding the Java Doc binary. installLocal uses a maven build, not a Gradle one so it might be that the maven build process is not working on your machine.

Yeah, there is something wrong there but I've upgraded the plugin https://github.com/getodk/javarosa/pull/732/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R129 and it fixes the issue so I think it's ok and I don't have to think about it more at least for now.

seadowg (approved):

PR #733, opened 2023-11-03
Add ability to process xpath expressions during parse

seadowg:
Work toward getodk/collect#5620

This adds a new `XPathProcessor` for `XFormProcessor` that allows a client to inspect every XPath expression in an XForm during parse:

```java
parser.addProcessor(new XFormParser.XPathProcessor() {
    @Override
    public void processXPath(@NotNull XPathExpression xPathExpression) {
        // Do something
    }
});
```

I've also added a lock to `XFormParser#parse` that will prevent two forms being parsed at the same time (even using two different `XFormParser` instances) as `XPathProcessor` (and existing parse code) manipulates static state that could cause problems if that were to happen.

New tests.

We had initially considered just using our existing bind processing with `XPathExpression#containsFunc` to find `pulldata`, but we realised that we needed a more general solution to be able to detect a custom function anywhere in the form (as it could appear in any xpath expression). I'm not a fan of the static state I've added here, but it felt like a losing battle to avoid it given `XFormParser` was already not thread safe in any way. Down the line, we should look at improving that so clients like Collect can parse multiple forms in parallel.

Stopping two forms from being parsed simultaneously is probably the riskiest change here. As far as I know, Collect doesn't do this on purpose, but it might be that it's a possibility with things like auto update or match exactly. That said, if that was happening, I'd potentially rather we crash the app than end up with whatever could happen mixing state from two form parses.

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -59,6 +59,7 @@ import org.javarosa.xml.util.InvalidStructureException;
 import org.javarosa.xml.util.UnfullfilledRequirementsException;
 import org.javarosa.xpath.XPathConditional;
 import org.javarosa.xpath.XPathParseTool;
+import org.javarosa.xpath.expr.XPathExpression;
 import org.javarosa.xpath.expr.XPathFuncExpr;
 import org.javarosa.xpath.expr.XPathNumericLiteral;
 import org.javarosa.xpath.expr.XPathPathExpr;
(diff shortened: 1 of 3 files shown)


src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -182,6 +183,9 @@ public class XFormParser implements IXFormParserFunctions {
     private final List<FormDefProcessor> formDefProcessors = new ArrayList<>();
     private final List<ModelAttributeProcessor> modelAttributeProcessors = new ArrayList<>();
     private final List<QuestionProcessor> questionProcessors = new ArrayList<>();
+    private final List<XPathProcessor> xpathProcessors = new ArrayList<>();
+
+    public static final List<XPathProcessor> tempXPathProcessors = new ArrayList<>();
```
lognaturel, line 188: Took me a bit to understand that this is just so the processors can be accessed statically even though they need to be set on an instance to use the existing `addProcessor`.

lognaturel (approved):

PR #734, opened 2023-11-14
Return false instead of crashing when asking if a repeat can be added at a deleted position

grzesiek2010:
Closes #https://github.com/getodk/collect/issues/5539

I've added tests and also tested the fix with ODK Collect.

In ODK Collect when we remove a repeat we need to find the last relevant index. In this case we navigate back and check every possible index. If something does not exist we don't want any exceptions but just simple false value.

It's a safe fix so testing the described scenario should be enough.

any form with nested repeats and `repeat_count` like the one attached in https://github.com/getodk/collect/issues/5539

No.


lognaturel (approved):
I initially was concerned that this change might negatively affect certain cases but I've spent some time trying to come up with an example and I can't. I confirmed that it does fix the issue that was filed.

PR #748, opened 2024-02-22
Additions needed for offline entity creation

seadowg:
This adds two things needed to implement offline entities in Collect:
- Expose entity ID and label in `EntityFormFinalizationProcessor`
- Add processor for `ExternalDataInstance` parsing that can be added to a new `ExternalInstanceParser`

New tests.

There's a lot of different ways we could go to get offline entities working. This is a first simple pass that focuses on getting the feature working and doesn't take into account additional goals like reducing the memory footprint of external secondary instances.

This only adds things so should be very low risk.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,7 +82,7 @@ if (!project.hasProperty("android")) {
         // Be sure to update version in pom.xml to match
         // snapshot release = x.x.x-SNAPSHOT
         // production release = x.x.x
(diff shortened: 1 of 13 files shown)


seadowg:
@lognaturel I've updated this so that we can override how external instances are parsed rather than just the XForm as a whole so we can intercept the parse in both the fresh XForm and the cached case. This is ready to go!

PR #752, opened 2024-03-01
Trim values stored in choices

grzesiek2010:
Closes #https://github.com/getodk/collect/issues/5692

I've added a new test and tested the form attached to https://github.com/getodk/collect/issues/5692 manually.

There are multiple options when it comes to storing choices. They can be a part of a form file, they can be also stored in an external csv file using different mechanisms (the `search`/`pulldata` function, `select_one_from_file`/`select_one_external` question types). We don't want them to have values with redundant whitespace. The constructor is the common place used by all of the options so it seems to be the easiest way to make sure the values are trimmed no matter what option we use. 

It's rather a safe change so testing the form from https://github.com/getodk/collect/issues/5692 should be enough.

There is a form attached to https://github.com/getodk/collect/issues/5692

No.


lognaturel (approved):
I had a hard time thinking through possible consequences for this one for some reason! I do think it's the right thing to do. Let's get it in our beta and mention it.

PR #755, opened 2024-04-03
Expose action with entity

seadowg:
Work for getodk/collect#6031

This will allow a client to determine that an exposed entity is an update and handle edge cases around that (like the one described in the attached issue).

Updated tests.

This is quite a simple change. I considered also removing the versioning logic from JavaRosa now that clients will have access to the action and theoretically could handle it for themselves. However, I wanted to keep the entity exposed as simple to deal with as possible - an `UPDATE` is communicating "update the entity with this id to this version with this label and properties".

Shouldn't change anything on client side as the change is purely additive.

### DIFF
--- a/src/test/java/org/javarosa/entities/EntityFormParserTest.java
+++ b/src/test/java/org/javarosa/entities/EntityFormParserTest.java
@@ -52,8 +52,8 @@ public class EntityFormParserTest {
         XFormParser parser = new XFormParser(new InputStreamReader(new ByteArrayInputStream(form.asXml().getBytes())));
         FormDef formDef = parser.parse(null);
 
-        EntityFormParser.EntityAction dataset = EntityFormParser.parseAction(EntityFormParser.getEntityElement(formDef.getMainInstance()));
-        assertThat(dataset, equalTo(EntityFormParser.EntityAction.CREATE));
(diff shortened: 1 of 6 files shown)


lognaturel (approved):

src/test/java/org/javarosa/entities/EntityFormParserTest.java
```diff
@@ -52,8 +52,8 @@ public void parseAction_findsCreateWithTrueString() throws XFormParser.ParseExce
         XFormParser parser = new XFormParser(new InputStreamReader(new ByteArrayInputStream(form.asXml().getBytes())));
         FormDef formDef = parser.parse(null);
 
-        EntityFormParser.EntityAction dataset = EntityFormParser.parseAction(EntityFormParser.getEntityElement(formDef.getMainInstance()));
-        assertThat(dataset, equalTo(EntityFormParser.EntityAction.CREATE));
+        EntityAction dataset = EntityFormParser.parseAction(EntityFormParser.getEntityElement(formDef.getMainInstance()));
```
lognaturel, line 55: Not: dataset seems like the wrong var name here. Action?

lognaturel (approved):
I’ll let you merge in case you want to update wording first!

PR #756, opened 2024-04-11
Add ability for for clients to provide their own instance parsing

seadowg:
Work towards https://github.com/getodk/collect/issues/6029

This adds a new `FilterInstanceParser` interface that is now used by `GeoJsonExternalInstance` and `CsvExternalInstance` and allows clients to add their own implementations.

Verified with a spike in Collect.

I tired using an "interceptor"/"preprocessor" approach, but then realised that the way we parse CSV and GeoJSON files is already a "chain" of parsers so formalising that made the most sense.

Shouldn't change behaviour - it adds the ability for clients to change behaviour from their side.

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/ExternalInstanceParser.java
+++ b/src/main/java/org/javarosa/xform/parse/ExternalInstanceParser.java
@@ -15,8 +15,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static java.util.Arrays.asList;
 
@@ -58,10 +56,7 @@ public class ExternalInstanceParser {
      * (via {@link FileInstanceParser#isSupported(String, String)}) first.
      */
     public void addFileInstanceParser(FileInstanceParser fileInstanceParser) {
-        fileInstanceParsers = Stream.concat(
-            Stream.of(fileInstanceParser),
-            fileInstanceParsers.stream()
-        ).collect(Collectors.toList());
+        fileInstanceParsers.add(0, fileInstanceParser);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/javarosa/xform/parse/ExternalInstanceParser.java
```diff
@@ -36,6 +53,17 @@ public void addProcessor(Processor processor) {
...
 
+    /**
+     * Adds {@link FileInstanceParser} before others. The last added {@link FileInstanceParser} will be checked
+     * (via {@link FileInstanceParser#isSupported(String, String)}) first.
+     */
+    public void addFileInstanceParser(FileInstanceParser fileInstanceParser) {
+        fileInstanceParsers = Stream.concat(
+            Stream.of(fileInstanceParser),
+            fileInstanceParsers.stream()
+        ).collect(Collectors.toList());
```
lognaturel, line 64: This got a little bit of an eyebrow raise from me. 😅 Is it taking a streaming approach too far? Do you want to avoid `fileInstanceParsers.add(0, fileInstanceParser)` to avoid direct mutation? Is it substantively different here? Is it to be consistent with other stream-based list processing?

I don't feel strongly but wanted to share my eyebrow raise.

lognaturel (approved):

src/main/java/org/javarosa/xform/parse/ExternalInstanceParser.java
seadowg: It's "completely forgetting about `add(0, fileInstanceParser)` and trying to append two lists". Remember we don't have to deal with any of this insanity in Collect any longer!

seadowg: Derp. Can't do that because it's an immutable list. Sigh.

PR #758, opened 2024-04-12
Use first repeat instance(s) for jr:choice-name choices

### LINKED ISSUE #757: jr:choice-name call using indexed-repeat outside a repeat with unbound reference as second parameter crashes
[Forum post](https://forum.getodk.org/t/format-note-summary-for-repeats-in-collect-and-enketo/46097?u=ln), report from support@

Some folks have expressions like `jr:choice-name(indexed-repeat(${choice}, ${thing}, 1), '${choice}')` in existing forms. This is intended to get the label of a selected value in a specific repeat instance from outside the repeat. People use it to display a summary of repeat data.

This generally worked ok before https://github.com/XLSForm/pyxform/issues/603 when `jr:choice-name` was often used with inline choice lists. Now, an XLSForm that worked before gets converted to use secondary instances instead and that crashes because the reference we use to identify the current choices is unbound (it references a full nodeset rather than a specific node). With secondary instance choice lists, there could conceivably be a choice filter on the list that makes the choices different depending on the repeat instance.

lognaturel:
Closes #757

Added tests

My initial instinct was that using a reference to a repeat nodeset rather than a specific repeat instance to specify where choices should be taken from is a form design issue. But then I realized that a standards-compliant XPath engine would use the first repeat instance(s) when given a nodeset reference where it needs a single node reference. 

This doesn't work well with choice filters but I think that's ok because it also wouldn't have been possible before. My priority here is to ensure that XLSForms that used to work continue to work without modification.

My recommendation for writing these kinds of `choice-name` calls would be to make them relative from within the repeat instance rather than making it from outside. 

This only affects `choice-name` calls given a reference parameter that is a nodeset reference. I think the case in the test is the only one in which this kind of reference would make sense. There's also the analogous nested case and that's why there's a for loop to qualify every unboud reference level.

https://docs.google.com/spreadsheets/d/1MZvF9ynsOdGMWeZF8YsjuoUK9E0t83LZMi5-RsBOEiI/edit#gid=1068911091

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -16,6 +16,25 @@
 
 package org.javarosa.core.model;
 
+import static java.util.Collections.emptyList;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Queue;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.javarosa.core.log.WrappedException;
 import org.javarosa.core.model.TriggerableDag.EventNotifierAccessor;
 import org.javarosa.core.model.actions.ActionController;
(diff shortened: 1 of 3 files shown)


src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -16,6 +16,25 @@
 
 package org.javarosa.core.model;
 
+import static java.util.Collections.emptyList;
```
lognaturel, line 19: I think I'm likely the one who messed up this order previously. 😬 My IDE should now enforce the order that's mostly consistent across existing files.

src/test/java/org/javarosa/core/model/ChoiceNameTest.java
```diff
@@ -117,4 +119,75 @@ public class ChoiceNameTest {
         scenario.answer("/data/city", "grenoble");
         assertThat(scenario.answerOf("/data/city_name"), is(stringAnswer("Grenoble")));
     }
+
+    @Test public void choiceNameCallWithIndexedRepeatAndStaticChoices_worksWithMultipleRepeats() throws IOException, XFormParser.ParseException {
```
lognaturel, line 123: This is not affected by the code change. It's there to document behavior that has always worked.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -866,10 +865,18 @@ public Object eval(Object[] args, EvaluationContext ec) {
                                 // the situations where that context would be used don't make sense for trying to reverse a
                                 // select value back to a label in an unrelated expression
                                 if (ref.isAmbiguous()) {
-                                    // SurveyCTO: We need a absolute "ref" to populate the dynamic choices,
-                                    // like we do when we populate those at FormEntryPrompt (line 251).
-                                    // The "ref" here is ambiguous, so we need to make it concrete first.
+                                    // ref, the reference used to specify where choices are defined, could be an absolute
```
seadowg, line 868: This `IFunctionHandler` implementation should be moved out to another file now we're touching it again.

seadowg (changes requested):

seadowg (approved):

PR #761, opened 2024-05-10
Add multi-argument and single string distance variants

### LINKED ISSUE #740: Allow geo functions to take string arguments
Currently the `distance` and `area` functions require a nodeset argument. That means they can't be used in predicates, for example. I believe we could use the same code that currently handles a 1-element nodeset on a literal string value.

lognaturel:
Closes #740

Added a tests, tried [sample form in Collect](https://docs.google.com/spreadsheets/d/1Dv9lD33Bfqz4XqmTNMJXm4NK9Lw0TloKph1juU0Kins/edit#gid=1068911091).

See https://github.com/getodk/xforms-spec/pull/311 for spec consideration.

I went back and forth on where to include the logic around argument quantity and types. I decided to put it in `XPathFuncExpr` because that's most inline with what other functions do and it feels helpful to get an overview of variants at that top level.

Regression risks center around the existing distance function. It's possible the changes break that existing variant in some subtle way.

https://docs.google.com/spreadsheets/d/1Dv9lD33Bfqz4XqmTNMJXm4NK9Lw0TloKph1juU0Kins/edit#gid=1068911091

https://github.com/getodk/docs/issues/1795

### DIFF
--- a/src/test/java/org/javarosa/core/util/GeoDistanceTest.java
+++ b/src/test/java/org/javarosa/core/util/GeoDistanceTest.java
@@ -209,26 +209,55 @@ public class GeoDistanceTest {
     }
 
     @Test
-    public void distance_isComputedForMultipleArguments() throws IOException, XFormParser.ParseException {
+    public void distance_isComputedForMultiplePathArguments() throws IOException, XFormParser.ParseException {
         Scenario scenario = Scenario.init("string distance", html(
             head(
                 title("Multi parameter distance"),
                 model(
                     mainInstance(t("data id=\"string-distance\"",
                         t("point1", "38.253094215699576 21.756382658677467 0 0"),
+                        t("point2", "38.25021274773806 21.756382658677467 0 0"),
                         t("point3", "38.25007793942195 21.763892843919166 0 0"),
                         t("point4", "38.25290886154963 21.763935759263404 0 0"),
                         t("point5", "38.25146813817506 21.758421137528785 0 0"),
                         t("distance")
                     )),
                     bind("/data/point1").type("geopoint"),
+                    bind("/data/point2").type("geopoint"),
                     bind("/data/point3").type("geopoint"),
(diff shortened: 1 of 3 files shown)


tiritea:
>An alternative would be to always only allow a single arg but to allow that arg to be a literal string rather than only a nodeset.

it actually seems more intuitive to me that the (singular) argument to the existing XPath distance() function should be able to support either a node/nodeset (for example, containing a geotrace value), or a literal string of the same. Many functions behave this way...

seadowg:
Removing myself as a reviewer until we've merged https://github.com/getodk/xforms-spec/pull/311

src/test/java/org/javarosa/core/util/GeoDistanceTest.java
```diff
@@ -149,6 +153,90 @@ public void distance_isComputedForString() throws IOException, XFormParser.Parse
...
+                        t("point3", "38.25007793942195 21.763892843919166 0 0"),
+                        t("point4", "38.25290886154963 21.763935759263404 0 0"),
+                        t("point5", "38.25146813817506 21.758421137528785 0 0"),
+                        t("distance")
+                    )),
+                    bind("/data/point1").type("geopoint"),
+                    bind("/data/point3").type("geopoint"),
+                    bind("/data/point4").type("geopoint"),
+                    bind("/data/point5").type("geopoint"),
+                    bind("/data/distance").type("decimal").calculate("distance(/data/point1, '38.25021274773806 21.756382658677467 0 0', /data/point3, /data/point4, /data/point5)")
```
seadowg, line 228: I'm guessing "point 2" is not pulled from a node to test that literals work fine here? I think if that's a concern it should move out to a separate test (or maybe a unit test of `XPathFuncExpr`) so it's not hidden in here.

seadowg (changes requested):

seadowg (approved):

PR #762, opened 2024-05-13
Don't create/update entities without an ID

seadowg:
- **Add failing test for null ID entities**
- **Don't return entities without an ID when updating**
- **Fix null ID on create case**

Closes #


src/test/java/org/javarosa/entities/EntitiesTest.java
```diff
@@ -127,6 +127,42 @@ public void fillingFormWithCreate_makesEntityAvailable() throws IOException, XFo
...
+            body(
+                input("/data/id")
+            )
+        ));
+
+        scenario.getFormEntryController().addPostProcessor(new EntityFormFinalizationProcessor());
+        scenario.finalizeInstance();
+
+        List<Entity> entities = scenario.getFormEntryController().getModel().getExtras().get(Entities.class).getEntities();
+        assertThat(entities.size(), equalTo(0));
```
seadowg, line 163: @lognaturel I considered exposing a blank ID to a client, but I'm not actually sure that makes sense. Given that `update` and `id` could both be conditional in a form, I'm not sure this is an exceptional case. It seems intuitive to me that if I should be able to write a form that conditionally updates based on whether an entity was ever selected or not for instance (like in this example form).

We could get strict and say that if `update` is truthy, then `id` cannot be null though? The [spec](https://getodk.github.io/xforms-spec/entities) would need some small updates for that, and I think if that's the case we'd want to fail to finalize (as the form would "invalid").

PR #763, opened 2024-05-13
Handle entities without IDs

seadowg:
- **Add failing test for null ID entities**
- **Don't return entities without an ID when updating**
- **Fix null ID on create case**
- **Allow clients to handle null entity IDs**

Closes #

### DIFF
--- a/src/test/java/org/javarosa/entities/EntitiesTest.java
+++ b/src/test/java/org/javarosa/entities/EntitiesTest.java
@@ -398,7 +398,7 @@ public class EntitiesTest {
                         t("data id=\"create-entity-form\"",
                             t("name"),
                             t("meta",
-                                t("entities:entity dataset=\"people\" create=\"1\"")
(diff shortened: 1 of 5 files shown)


src/test/java/org/javarosa/entities/EntitiesTest.java
```diff
@@ -213,6 +213,39 @@ public void fillingFormWithUpdate_andNoLabel_makesEntityAvailableWithNullLabel()
         assertThat(entities.get(0).properties, equalTo(asList(new Pair<>("name", "Tom Wambsgans"))));
     }
 
+    @Test
+    public void fillingFormWithUpdate_withNullId_doesNotMakeEntityAvailable() throws IOException, XFormParser.ParseException {
+        Scenario scenario = Scenario.init("Create entity form", XFormsElement.html(
```
lognaturel, line 259: Title should be update not create

lognaturel (approved):
Minor naming issue and then this looks ready to merge to me. I’d prefer to squash because the direction changed but will leave that up to you.

PR #765, opened 2024-05-28
Use Maven plugin for local publish

seadowg:
This replaces `installLocal` with `publishToMavenLocal`. This is a good baby step towards replacing Maven entirely with Gradle which will simplify development workflows.

We'll need to make a few updates on the Collect side after this, but that's fine to do after (and won't make sense to do before).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,6 @@ out
 # mvn
 target
 secrets.xml
-pom-local.xml
 
 # logging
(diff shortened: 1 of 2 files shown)


lognaturel (approved):
Was trying to verify this with Collect forgetting the comment that it's not supposed to work yet. I have confirmed that the expected artifacts are created.

PR #766, opened 2024-06-03
Include Maven compile scope dependencies in Gradle JAR

seadowg:
The `build.gradle` was declaring dependencies in an incompatible way to Maven. In Maven, the default scope (`compile`) includes the depdencies in the JAR. This is more analgous to Gradle's `api` than `implementation` (or `compileOnly`) which was being used and would mean that downstream dependencies would not recieve these. This creates problems for Collect which uses JavaRosa's included dependencies (like `kxml` for instance).

Down the line, we should move back to `implementation`, but I think that makes more sense once we've completely removed Maven.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,13 +48,11 @@
       <groupId>net.sf.kxml</groupId>
       <artifactId>kxml2</artifactId>
       <version>2.3.0</version>
-      <scope>compile</scope>
     </dependency>
     <dependency>
        <groupId>joda-time</groupId>
        <artifactId>joda-time</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -48,13 +48,11 @@
       <groupId>net.sf.kxml</groupId>
       <artifactId>kxml2</artifactId>
       <version>2.3.0</version>
-      <scope>compile</scope>
```
seadowg, line 51: This was probably confusing things: `compile` is the default and doesn't need to be specified.

lognaturel (approved):

PR #770, opened 2024-06-10
Fix crash for non-`String` entity labels

seadowg:
Closes https://github.com/getodk/collect/issues/6167

New tests!

Not a lot to discuss here. We now always convert to a string instead of casting, which should work for all the possible types.

The change is pretty specific, and will only affect entity forms.

### DIFF
--- a/src/test/java/org/javarosa/entities/EntityFormParserTest.java
+++ b/src/test/java/org/javarosa/entities/EntityFormParserTest.java
@@ -1,92 +1,30 @@
 package org.javarosa.entities;
 
-import kotlin.Pair;
-import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.data.IntegerData;
 import org.javarosa.core.model.instance.TreeElement;
-import org.javarosa.core.util.XFormsElement;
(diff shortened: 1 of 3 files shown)


src/test/java/org/javarosa/entities/EntityFormParserTest.java
```diff
@@ -1,90 +1,41 @@
...
+        entityElement.setAttribute(null, "update", "true");
 
-        EntityAction dataset = EntityFormParser.parseAction(EntityFormParser.getEntityElement(formDef.getMainInstance()));
+        EntityAction dataset = EntityFormParser.parseAction(entityElement);
         assertThat(dataset, equalTo(EntityAction.UPDATE));
     }
+
+    @Test
+    public void parseLabel_whenLabelIsAnInt_convertsToString() {
+        TreeElement labelElement = new TreeElement("label");
```
grzesiek2010, line 33: I think we should have consts for values like this one `label`  and maybe for some others too like `id`

grzesiek2010 (changes requested):

lognaturel (approved):

PR #775, opened 2024-07-08
Add partial elements

seadowg:
Work towards https://github.com/getodk/collect/issues/5623

This adds the ability to provide "partial" parses of external instances using a custom `FileInstanceParser`:

```java
class MyFileInstanceParser implements ExternalInstanceParser.FileInstanceParser {

        @Override
        public TreeElement parse(@NotNull String instanceId, @NotNull String path) throws IOException {
            return parse(instanceId, path, false);
        }

        @Override
        public TreeElement parse(@NotNull String instanceId, @NotNull String path, boolean partial) throws IOException {
            TreeElement root = new TreeElement("root", 0);

            TreeElement value = new TreeElement("value");
            TreeElement label = new TreeElement("label");

            if (!partial) {
                value.setValue(new StringData("blah"));
                label.setValue(new StringData("Blah"));
            }

            TreeElement item = new TreeElement("item", 0, partial);
            item.addChild(value);
            item.addChild(label);

            root.addChild(item);
        }
}
```

JavaRosa will allow partial parses for parsing/deserializing forms (by passing `partial` as `true`), but will require a "full" parse (`partial` as `false`) if it needs to resolve a reference currently involving a partial `TreeElement`. 

Clients can also populate partial elements at an instance level by calling `DataInstance#populatePartialElements`. This will allow plugins to populate partials before they are resolved to avoid full parses (from a `FilterStrategy` for example).

New tests. I've also built a toy implementation in Collect to validate that the API feels correct.

There's a few different approaches that I've discussed in different chats and also in https://github.com/getodk/collect/issues/5623 (like fallbacks for individual elements rather than full parses for example). This approach has ended up winning out, so unless there's anything that feels incorrect here I don't think there's that much that needs to be pointed out.

There should be very little risk here as the new code will only kick in if "activated" by a client. The big risk is that there's some edge case I've missed that we'll run into down the road when adding in these new features.

### DIFF
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -26,6 +26,5 @@ Inspect external instances (their ID and parsed XML) after parsing or provide cu
 
 ### API
 - `ExternalInstanceParser#addFileInstanceParser`
-- `ExternalInstanceparsser#addProcessor`
 
 The default `ExternalInstanceParser` can be overridden by creating an implementation of `ExternalInstanceParserFactory` and calling `XFormUtils.setExternalInstanceParserFactory` with it.
--- a/src/test/java/org/javarosa/plugins/InstancePluginTest.java
+++ b/src/test/java/org/javarosa/plugins/InstancePluginTest.java
@@ -1,0 +1,271 @@
+package org.javarosa.plugins;
+
+import kotlin.Pair;
+import org.javarosa.core.model.SelectChoice;
+import org.javarosa.core.model.data.StringData;
+import org.javarosa.core.model.instance.DataInstance;
+import org.javarosa.core.model.instance.TreeElement;
+import org.javarosa.test.Scenario;
+import org.javarosa.test.TempFileUtils;
+import org.javarosa.xform.parse.ExternalInstanceParser;
+import org.javarosa.xform.parse.ExternalInstanceParserFactory;
+import org.javarosa.xform.parse.XFormParser;
+import org.javarosa.xform.util.XFormUtils;
+import org.jetbrains.annotations.NotNull;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+
+import static java.util.Arrays.asList;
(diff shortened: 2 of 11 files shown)


PLUGINS.md
```diff
@@ -26,6 +26,5 @@ Inspect external instances (their ID and parsed XML) after parsing or provide cu
 
 ### API
 - `ExternalInstanceParser#addFileInstanceParser`
-- `ExternalInstanceparsser#addProcessor`
```
seadowg, line 29: It feels like `addFileInstanceParser` replaced this, so I wanted to take the opportunity to clean up the API.

lognaturel (approved):
I still don’t feel fully comfortable with having to first create partial elements and then fill them in as opposed to only creating the elements that need to exist. But let’s experiment with this and see how it feels!

Slight preference for squashing.

src/test/java/org/javarosa/plugins/InstancePluginTest.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    @Test
+    public void supportsPartialElements() throws IOException, XFormParser.ParseException {
+        externalInstanceParserFactory.setFileInstanceParser(new FakeFileInstanceParser(asList(
+            new Pair<>("0", "Item 0"),
+            new Pair<>("1", "Item 1")
+        ), true));
+
+        File tempFile = TempFileUtils.createTempFile("fale-instance", "fake");
```
lognaturel, line 59: Fale -> fake

seadowg:
> I still don’t feel fully comfortable with having to first create partial elements and then fill them in as opposed to only creating the elements that need to exist. But let’s experiment with this and see how it feels and performs, I may very well warm up to it! I’ll also refresh my memory on the reference verification that happens at form parse time.

Agreed! I realized when working on https://github.com/getodk/javarosa/pull/775/commits/5f9fc09483d751a1b5b3174eacf4c8f66cbb4637 that our current implementation should also work for a version where a `FileInstanceParser` produces totally empty `item` elements. It's just the verification that's blocking us. Hopefully we can remove/move that so we can get that extract but of gains.

PR #776, opened 2024-07-08
Add indexed-repeat tests

lognaturel:
- [x] Modify first test to use a relative reference for index and add a group around the repeat so the contextualization is clear
- [x] Add nested repeat tests
       https://github.com/getodk/web-forms/pull/195/commits/541cd1200818c3024eb98a5aa973dc6142acfecd#diff-d7f5e2a9fa88fc71566d393928de170be5700174a890bc68e5195a4f8c6e1657R335
- [x] Demonstrate that the order of the repeat/index pairs doesn't matter (included in nested repeat test above)
- [x] Multiple repeat/index pairs where the repeat refs are relative (existing in `IndexedRepeatRelativeRefsTest`)

These are test-only additions intended to capture the intended `indexed-repeat` behavior. It also updates the function's comment to better reflect the language that we use these days to talk about repeats.

I wasn't there when the function was added by my understanding is that it was introduced so end users wouldn't have to write XPath predicates in the common case where they want to use a numeric index to access a specific repeat instance's value. XPath predicates are harder to reason about because the predicate expression is evaluated in the context of the path that's being filtered. It's particularly baffling when using parallel repeats because you end up having to write something like `position(.) = position(current()/..)`.

The `indexed-repeat` function makes it so the index expression is evaluated using the bound node's context which is intuitive to users. The first parameter is the full reference to a field that needs to get indexed, and then there are some number of pairs of repeat references and indexes.

A simple algorithm to think about the intended behavior (not exactly what JR does):
- contextualize all the parameters based on the single node binding
- evaluate all the index parameters (param 3, 5, 7, etc), these should now be literal integers
- for each repeat reference parameter (param 2, 4, 6, etc), add a positional predicate at that reference level in the base reference (param 1)

### DIFF
--- a/src/test/java/org/javarosa/xpath/expr/IndexedRepeatTest.java
+++ b/src/test/java/org/javarosa/xpath/expr/IndexedRepeatTest.java
@@ -1,8 +1,12 @@
 package org.javarosa.xpath.expr;
 
+import org.javarosa.test.Scenario;
+import org.javarosa.xpath.XPathTypeMismatchException;
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.instanceOf;
-import static org.hamcrest.CoreMatchers.is;
 import static org.javarosa.core.test.AnswerDataMatchers.stringAnswer;
 import static org.javarosa.test.BindBuilderXFormsElement.bind;
 import static org.javarosa.test.XFormsElement.body;
@@ -16,10 +20,6 @@ import static org.javarosa.test.XFormsElement.repeat;
 import static org.javarosa.test.XFormsElement.t;
 import static org.javarosa.test.XFormsElement.title;
 import static org.junit.Assert.fail;
-
-import org.javarosa.test.Scenario;
-import org.javarosa.xpath.XPathTypeMismatchException;
-import org.junit.Test;
 
 public class IndexedRepeatTest {
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/org/javarosa/xpath/expr/IndexedRepeatTest.java
```diff
@@ -0,0 +1,135 @@
...
+                head(
+                    title("indexed-repeat"),
+                    model(
+                        mainInstance(t("data id=\"indexed-repeat\"",
+                            t("outside"),
+                            t("repeat",
+                                t("inside")),
+                            t("calc")
+                        )),
+                        bind("calc").calculate("indexed-repeat(/data/outside, /data/repeat, 1)")
```
lognaturel, line 38: calc -> /data/calc

src/test/java/org/javarosa/xpath/expr/IndexedRepeatTest.java
```diff
@@ -0,0 +1,135 @@
...
+                    )),
+                    bind("/data/repeat2/from_repeat1").calculate("indexed-repeat(/data/repeat1/inside1, /data/repeat1, position(..))")
+                )
+            ),
+            body(
+                repeat("/data/repeat1",
+                    input("/data/repeat1/inside1")),
+
+                repeat("/data/repeat2",
+                    input("/data/repeat2"))
```
lognaturel, line 116: /data/repeat2/inside2

seadowg (approved):

PR #780, opened 2024-07-15
Use same initialization order in Scenario as Collect

seadowg:
This switches the initialization order in `Scenario` to match Collect (`FormDef#initialize` is called after creating the `FormEntryController`). It also makes it possible to pass in a custom function for constructing the `FormEntryController` to allow tests to use plugins (like `#addFilterStrategy`).

### DIFF
--- a/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/src/main/java/org/javarosa/core/model/FormDef.java
@@ -1629,11 +1629,20 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
         return extras;
     }
 
+    /**
+     * @deprecated use {@link FormEntryController#addFilterStrategy(FilterStrategy)} instead
+     */
+    @Deprecated
     public void addFilterStrategy(FilterStrategy filterStrategy) {
         customFilterStrategies.add(filterStrategy);
         resetEvaluationContext();
     }
 
+
+    /**
+     * @deprecated use {@link FormEntryController#addFunctionHandler(IFunctionHandler)} instead
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1629,11 +1629,20 @@ public Extras<Externalizable> getExtras() {
         return extras;
     }
 
+    /**
+     * @deprecated use {@link FormEntryController#addFilterStrategy(FilterStrategy)} instead
```
lognaturel, line 1633: Since we're doing a major version bump anyway, what do you think about removing these right away? It's very unlikely there are existing consumers and if there are it's an easy change. We'd just want to make sure to track these breaking changes somewhere so we can document them.

lognaturel (approved):
This is a really great time to make these kinds of changes since we're just exposing Scenario for the first time as an intentional public interface and we are planning a major version bump 👍

src/main/java/org/javarosa/core/model/FormDef.java
seadowg: I'd thought about that, but we'd need to do significant work to move `EvaluationContext` ownership to `FormEntryController` (as that's what these methods affect). I gave doing a minimal change (where we still end up with a coupling between `FormDef` and `FormEntryController` but the `EvaluationContext` moves) a go, but it was pretty involved.

lognaturel: Ah, ok, I thought it was an easier change that that. Sounds good.

PR #781, opened 2024-07-15
Document current verifyItemsetBindings behavior

lognaturel:
Only the first item in a secondary instance gets its label and value references verified.

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
+++ b/src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
@@ -336,24 +336,24 @@ class FormInstanceParser {
 
             //make sure the labelref is tested against the right instance
             //check if it's not the main instance
-            DataInstance fi = null;
+            DataInstance secondaryInstance;
             if (itemset.labelRef.getInstanceName() != null) {
-                fi = formDef.getNonMainInstance(itemset.labelRef.getInstanceName());
-                if (fi == null) {
+                secondaryInstance = formDef.getNonMainInstance(itemset.labelRef.getInstanceName());
+                if (secondaryInstance == null) {
                     throw new XFormParseException("Instance: " + itemset.labelRef.getInstanceName() + " Does not exists");
                 }
             } else {
-                fi = instance;
+                secondaryInstance = instance;
             }
 
             // Don't try to validate references if the external instance could not be resolved. We allow parsing a form
             // with placeholder external secondary instances for cases where a ReferenceManager can't be configured.
-            if (!(fi instanceof ExternalDataInstance) || !((ExternalDataInstance) fi).isUsingPlaceholder()) {
-                if (fi.getTemplatePath(itemset.labelRef) == null) {
+            if (!(secondaryInstance instanceof ExternalDataInstance) || !((ExternalDataInstance) secondaryInstance).isUsingPlaceholder()) {
+                if (secondaryInstance.getTemplatePath(itemset.labelRef) == null) {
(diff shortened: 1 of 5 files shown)


src/test/java/org/javarosa/xform/parse/ExternalSecondaryInstanceParseTest.java
```diff
@@ -134,6 +134,59 @@ public void xformParseException_whenItemsetConfiguresValueOrLabelNotInExternalIn
...
+        try {
+            Scenario.init("Some form", html(
+                head(
+                    title("Some form"),
+                    model(
+                        mainInstance(t("data id=\"some-form\"",
+                            t("first")
+                        )),
+
+                        t("instance id=\"mixed-schema\" src=\"jr://file/mixed-schema2.xml\""),
```
seadowg, line 173: Can we use inline instances here or would they not get verified? If we can, they'd make the test much easier to read in my opinion.

seadowg (changes requested):

src/test/java/org/javarosa/xform/parse/ExternalSecondaryInstanceParseTest.java
lognaturel: Coming back at you from the past!

In https://github.com/getodk/javarosa/commit/847865c78512af9b65211ea332d23e17c119fa07#diff-e045a71eed0a5d3fd849d1ca17f98246297bcae6d2d6b62203e17ec297365b03R348, I changed the logic to only verify itemset bindings for external secondary instances. I think that was a mistake so thanks for getting me to investigate.

I have changed the condition so all itemset bindings are verified. I've added test coverage for internal secondary instances. There was actually already some coverage for external secondary instances at https://github.com/getodk/javarosa/blob/a5ea1e95f322165e95e22ff01ea0495fbcfed296/src/test/java/org/javarosa/xform/parse/ExternalSecondaryInstanceParseTest.java#L108

src/test/java/org/javarosa/core/model/SelectChoiceTest.java
```diff
@@ -111,7 +114,7 @@ public void getChild_returnsEmptyString_whenChoicesAreFromSecondaryInstance_andR
                         t("data id='select-empty'",
                             t("select"))),
                     instance("choices",
-                        t("item", t("label", "Item"), t("property", ""))
+                        t("item", t("label", "Item"), t("name", "item"), t("property", ""))
```
lognaturel, line 117: This was a mistake introduced because of the lack of binding check for internal secondary instances!

seadowg (changes requested):
Looks like there are conflicts, so I'll have a proper look after those are resolved.

src/main/java/org/javarosa/xform/parse/FormInstanceParser.java
```diff
@@ -348,7 +348,7 @@ private void verifyItemsetBindings(FormInstance instance) {
 
             // Don't try to validate references if the external instance could not be resolved. We allow parsing a form
             // with placeholder external secondary instances for cases where a ReferenceManager can't be configured.
-            if (fi instanceof ExternalDataInstance && !((ExternalDataInstance) fi).isUsingPlaceholder()) {
+            if (!(fi instanceof ExternalDataInstance) || !((ExternalDataInstance) fi).isUsingPlaceholder()) {
```
seadowg, line 351: While we're here, let's change that `fi` variable to something more readable. It's especially mad here where we end up with and `if` in front of it.

seadowg (changes requested):
Looks good! Just one wee tweak that'd be good to make before moving on.

seadowg (approved):

PR #782, opened 2024-07-16
Add way of providing external instances without an existing file

seadowg:
This adds a new interface `InstanceProvider` which can be used to provide custom instance "parsing", but without needing an actual file to be referenced by `src`. This makes it a lot easier to write tests for custom parsing (we don't need to deal with `ReferenceManager`) that don't actually care about the file in `src`, and will allow us to also provide instances without an underlying file. 

New tests.

We alternatively could have made `FileInstanceParser` implementations handle converting `src` to an actual file path, but this would have meant changing that interface and also duplicating a bunch of code that's currently shared.

This just adds a new plugin interface, so there's not a lot of risk.

### DIFF
--- a/src/main/java/org/javarosa/test/XFormsElement.java
+++ b/src/main/java/org/javarosa/test/XFormsElement.java
@@ -44,7 +44,7 @@ public interface XFormsElement {
         Map<String, String> attributes = new HashMap<>();
         String[] words = name.split(" ");
         for (String word : asList(words).subList(1, words.length)) {
-            String[] parts = word.split("(?<!\\))=(\"|')");
+            String[] parts = word.split("(?<!\\))=(\"|')", 2);
             attributes.put(parts[0], parts[1].substring(0, parts[1].length() - 1));
         }
         return attributes;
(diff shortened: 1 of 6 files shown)


src/main/java/org/javarosa/test/XFormsElement.java
```diff
@@ -44,7 +44,7 @@ static Map<String, String> parseAttributes(String name) {
         Map<String, String> attributes = new HashMap<>();
         String[] words = name.split(" ");
         for (String word : asList(words).subList(1, words.length)) {
-            String[] parts = word.split("(?<!\\))=(\"|')");
+            String[] parts = word.split("(?<!\\))=(\"|')", 2);
```
seadowg, line 47: I ran into this problem when working on a related tests in Collect. Basically `select1Dynamic` did not work with nodeset expressions like `instance('things')/root/item[property='value']` because it would split on the second `=`.

lognaturel: 😮 That's a useful `split` signature I did not know about!

lognaturel (approved):

src/main/java/org/javarosa/test/XFormsElement.java
seadowg: Just discovered it myself!

PR #786, opened 2024-07-24
Add commit SHA to SNAPSHOT versions

seadowg:
This means we'll end up with versions like `5.0.0-SNAPSHOT-8ad757e` which will allow clients to explicitly update rather than end up potentially getting a new version every build.

This should also allow clients to relax their caching strategies to hold versions for longer and not rely on constantly checking dependency infrastructure for new "versions" of a snapshot.

My understanding is that this approach **should** work with OSSRH, but please correct me if I'm wrong.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,18 +53,6 @@ targetCompatibility = '1.8'
 sourceCompatibility = '1.8'
 
 if (!project.hasProperty("android")) {
-    jar {
-        // Be sure to update version in pom.xml to match
-        // snapshot release = x.x.x-SNAPSHOT
-        // production release = x.x.x
-        archiveVersion = '5.5.0-SNAPSHOT'
-        archiveBaseName = "javarosa"
-
-        manifest {
-            attributes 'Manifest-Version': "$jar.archiveVersion"
-        }
-    }
-
     // TODO: does not build UML diagrams
     javadoc {
         failOnError = false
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -3,9 +3,9 @@
   <groupId>org.getodk</groupId>
   <artifactId>javarosa</artifactId>
   <!-- Be sure to update version in build.gradle to match -->
-  <!-- snapshot release = x.x.x-SNAPSHOT -->
+  <!-- snapshot release = x.x.x-SNAPSHOT-${revision} -->
   <!-- production release = x.x.x -->
-  <version>5.0.0-SNAPSHOT</version>
+  <version>5.0.0-SNAPSHOT-${revision}</version>
```
seadowg, line 8: There's magic properties that allow this to work: https://maven.apache.org/maven-ci-friendly.html

build.gradle
```diff
@@ -53,18 +53,6 @@ targetCompatibility = '1.8'
 sourceCompatibility = '1.8'
 
 if (!project.hasProperty("android")) {
-    jar {
```
seadowg, line 56: We won't need this now we're using the `maven-publish` plugin (which explicitly sets versions).

lognaturel (approved):
I don’t know of a reason why this wouldn’t work. Let’s merge and try it. We can always revert if it doesn’t.

PR #787, opened 2024-07-25
Close CSVParser properly and switch to record wise parsing to reduce memory use

seadowg:
Does what it says on the tin! It seems like the CSV parsing code we had was using a different approach from the one we'd want for low memory (possibly to optimize for speed). It was clear from reading [the docs](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVParser.html) that we were firstly not closing the resource properly, and secondly could be parsing files on a record-by-record basis. 

I'd initially been looking to just write up an issue for this, but then realized the changes needed were very small.

I've created a `v4.4.x` branch for this to merge into as I'm thinking that we could do a hotfix release here and a corresponding one in Collect.


lognaturel (approved):
Good find!

PR #789, opened 2024-08-01
jr:count value should always be coerced to number

### LINKED ISSUE #783: jr:count value should always be coerced to number
#### Software versions 
JavaRosa v4.4.0

#### Problem description
When a number from an external CSV is used as a `jr:count` value, that number is not respected.

See sample form [on the forum](https://forum.getodk.org/t/number-value-from-entity-list-does-not-work-as-a-repeat-count-in-collect/48175).

#### Expected behavior
Any value type should be coerced to a number.

grzesiek2010:
Closes #783 

I've improved automated tests and performed manual ones using the form attached below.

This improved `AnswerDataUtil#answerDataToInt` so that it supports parsing `StringData` to integers. It was a missing piece and that's why it was not possible to use numbers from an external CSV in `repeat_count`.

It should just add support for using numbers from an external CSV in `repeat_count`, nothing else should be affected.

I've used this: 
[families.csv](https://github.com/user-attachments/files/16461053/families.csv)
[families_form.xlsx](https://github.com/user-attachments/files/16461054/families_form.xlsx)

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/data/helper/AnswerDataUtil.java
+++ b/src/main/java/org/javarosa/core/model/data/helper/AnswerDataUtil.java
@@ -15,6 +15,13 @@ public class AnswerDataUtil {
             return (int) Math.floor((Double) count);
         } else if (count instanceof Long) {
             return (int) ((Long) count).longValue();
+        } else if (count instanceof String) {
+            try {
+                double value = Double.parseDouble((String) count);
+                return (int) Math.floor(value);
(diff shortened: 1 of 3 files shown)


seadowg (approved):

grzesiek2010:
I think this should be tested, do you agree @seadowg?

seadowg:
@grzesiek2010 yup sounds good!

PR #798, opened 2024-10-09
Fix ambiguous constructors

seadowg:
This fixes a problem introduced by https://github.com/getodk/javarosa/pull/794 when using JavaRosa with Kotlin: Java's `int` and `@NotNull Integer` both map to Kotlin's `Int` which makes overloaded constructors using those types ambiguous (this goes for any primitive type and its corresponding "boxed" type).

### DIFF
--- a/src/main/java/org/javarosa/core/model/data/DecimalData.java
+++ b/src/main/java/org/javarosa/core/model/data/DecimalData.java
@@ -45,9 +45,6 @@ public class DecimalData implements IAnswerData {
     public DecimalData(double d) {
         this.d = d;
     }
-    public DecimalData(@NotNull Double d) {
-        setValue(d);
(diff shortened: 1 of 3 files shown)


lognaturel (approved):
Is this a breaking change? Or maybe the `@NotNull` addition was technically a breaking change? This would be an issue only for Java clients passing in nullable objects, right?

Not a big deal because we're working towards a major release, just want to get the release notes right.

seadowg:
> Is this a breaking change? Or maybe the @NotNull addition was technically a breaking change? This would be an issue only for Java clients passing in nullable objects, right?

The `@NotNull` change was breaking, but passing a `null` into those constructors would explode at runtime anyway (as there is a check in `setValue`).

PR #799, opened 2024-10-11
Always include secondary instances to support custom functions that references them

seadowg:
Closes #


seadowg:
We won't be making this change quite yet.

PR #801, opened 2024-10-14
hash un-numeric input when used as PRNG seed, fixes #800

### LINKED ISSUE #800: Seed for randomization derived from non-numeric-looking string is always 0
Related: getodk/web-forms#240
Related: getodk/web-forms#49

[Here is](https://staging.getodk.cloud/#/projects/112/forms/seeded_rando) a form that uses [choice randomization](https://xlsform.org/en/#randomize-choices).
Yes, the referenced seed node is a text input! And if you put non-numerical looking text in it, by my reading of OpenRosa, it'll evaluate as a NaN double, which then becomes 0 when converted to the Long that the Park-Miller PRNG takes as a seed. Which means that basically any text that doesn't happen to look like a number results in the same seed.

It's not a huge disaster (which is why it went unnoticed?) but on the other hand there's no indication for this behaviour; the spec (not xform, but xlsform) doesn't say "only use integer / numeric types or numberish-looking text", and neither does `ODK Validate` say anything about it, nor is there a runtime warning (not that we'd want one).
Thus to a user doing a superficial test of their survey it *does* look like the choice list gets randomized when they put in some text in the designated seed field. They might find out with more extensive testing that almost any string (any non-numeric-looking string) results in the same sort order though, and only at that point they might think about how what was promised relates to these observations, and how those oberservations relate to what they want. And that sort of ambiguity is probably not what we want.

I stumbled upon this in the context of getodk/web-forms#49.
Of the solution alternatives I could come up with ("do nothing but port the behaviour to webforms", "clarify current behaviour in the spec", and "fix things") @lognaturel is leaning towards "fix things", which I thought about as:

- in the current fallthrough case (string can't be parsed into a Double) currently resulting in NaN, we take the original string value instead, hash-digest it (in utf8-encoding), and read the first 8 bytes of the resulting digest as a double. The goal is to achieve stable and reproducible (cross-platform) seeded randomization regardless of the nature of the input, while giving the same results for inputs that are currently not squashed to `0` (and thus being backwards-compatible).

brontolosone:
Closes #800

As discussed on Slack, but there may be more to discuss!

Ran the tests, added some tests too. Ideally I'd put this JR version in an ODKCollect to witness some choice randomization change when altering the seed value in [this test form](https://staging.getodk.cloud/#/projects/112/forms/seeded_rando), but I didn't manage that today ;-)

The nice thing is that it doesn't change existing randomizations that were working well (eg, strings interpretable as numerics). It only changes randomizations for which the seed wasn't working (eg, strings not interpretable as numerics).

In very specific use cases this can cause an unwelcome change.

For instance: In some survey the randomization of the order of choice fields is part of the survey. Also, it wasn't actually working due to issue #800, but anyway. Now if ODKCollect is updated with this new OpenRosa version halfway through the survey, some users will have seen randomization X for seed Z while others will have seen randomization Y for that same seed Z. If the research draws conclusions based on either X or Y, then the researchers need to know which version of ODKCollect was used so they know (or can deduce) which choice randomization went with seed Z.

as mentioned above, running ODKCollect on [this form](https://staging.getodk.cloud/#/projects/112/forms/seeded_rando) and witnessing the choice list reshuffle based on string input variations in the seed field, would be extra convincing.

Maybe. We could say something like:

> "As of ODKCollect version X, the choice randomization based on using an RNG seed from some other field actually works for more inputs (notably, any text). If you have ongoing research that depends on this type of randomization, and where the analysis is dependent on the specific choice list ordering that used to go with certain seeds, you may want to test with this new version of ODKCollect to see if your choice list randomization will change, and decide to not upgrade ODKCollect while this research is ongoing".

### DIFF
--- a/src/main/java/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/main/java/org/javarosa/core/model/ItemsetBinding.java
@@ -2,6 +2,7 @@ package org.javarosa.core.model;
 
 import static org.javarosa.core.model.FormDef.getAbsRef;
 import static org.javarosa.xform.parse.RandomizeHelper.shuffle;
+import static org.javarosa.xpath.expr.XPathFuncExpr.toLongHash;
 import static org.javarosa.xpath.expr.XPathFuncExpr.toNumeric;
 
 import java.io.DataInputStream;
@@ -36,6 +37,7 @@ import org.javarosa.debug.Event;
 import org.javarosa.model.xform.XPathReference;
 import org.javarosa.xpath.XPathConditional;
 import org.javarosa.xpath.XPathException;
+import org.javarosa.xpath.XPathNodeset;
 import org.javarosa.xpath.expr.XPathNumericLiteral;
 import org.javarosa.xpath.expr.XPathPathExpr;
 
@@ -309,10 +311,21 @@ public class ItemsetBinding implements Externalizable, Localizable {
     }
 
     private Long resolveRandomSeed(DataInstance model, EvaluationContext ec) {
+        XPathNodeset seedNode = null;
         if (randomSeedNumericExpr != null)
             return ((Double) randomSeedNumericExpr.eval(model, ec)).longValue();
-        if (randomSeedPathExpr != null)
-            return toNumeric(randomSeedPathExpr.eval(model, ec)).longValue();
+        if (randomSeedPathExpr != null) {
+            seedNode = randomSeedPathExpr.eval(model, ec);
+            Double asDouble = toNumeric(seedNode);
+            if (asDouble == Double.NaN) {
+                // Reasonable attempts at reading the node's value as a number failed.
+                // Fall back to deriving the seed from it using hashing.
+                // See https://github.com/getodk/javarosa/issues/800
+                return toLongHash(seedNode);
+            } else {
+                return asDouble.longValue();
+            }
+        }
         return null;
     }
 
--- a/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -22,6 +22,8 @@ import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Date;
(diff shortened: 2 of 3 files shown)


lognaturel (approved):
Thanks!

lognaturel:
This has been kicking around in my head since I merged it and I just realized that it doesn't do anything! Because of the way that selects are represented, there are two code paths to seeds being cleaned and converted to `Long`: one in `XPathFuncExpr` for function calls used in general expressions and one in `XFormParser` and `ItemsetBinding` for function calls used in itemset declarations. This change targets the second one but that code path already throws in the case of non-numeric literals before getting to the new additions.

Here are some tests showing this discrepancy:
```
    // The current implementation for randomize for an ItemsetBinding requires a literal numeric value and does not
    // attempt to convert
    @Test
    public void throwsWhenSeedForItemsetExpressionNotNumeric() throws IOException, XFormParser.ParseException {
        try {
            Scenario scenario = Scenario.init("Randomize non-numeric seed", html(
                head(
                    title("Randomize non-numeric seed"),
                    model(
                        mainInstance(t("data id=\"rand-non-numeric\"",
                            t("choice")
                        )),
                        instance("choices",
                            item("a", "A"),
                            item("b", "B")
                        ),
                        bind("/data/choice").type("string")
                    )
                ),
                body(
                    select1Dynamic("/data/choice", "randomize(instance('choices')/root/item, '0')")
                )
            ));

            System.out.println(scenario.choicesOf("/data/choice"));
            fail("Expecting XPathTypeMismatchException");
        } catch (XPathTypeMismatchException e) {
            // Expected
        }
    }

    // When a randomize expression is used in a context other than in an itemset binding, the seed value is converted to
    // numeric
    @Test
    public void nonNumericSeedConvertsToIntWhenUsedInExpression() throws IOException, XFormParser.ParseException {
        Scenario scenario = Scenario.init("Randomize non-numeric seed", html(
            head(
                title("Randomize non-numeric seed"),
                model(
                    mainInstance(t("data id=\"rand-non-numeric\"",
                        t("choice")
                    )),
                    instance("choices",
                        item("a", "A"),
                        item("b", "B")
                    ),
                    bind("/data/choice").type("string").calculate("selected-at(join(' ', randomize(instance('choices')/root/item/label, '1')), 0)")
                )
            ),
            body(
                input("/data/choice")
            )
        ));

        assertThat(scenario.answerOf("/data/choice").getDisplayText(), is("B"));
    }
```

lognaturel:
I initially thought that we should roll this change back because we're just about to freeze for Collect regression testing. Having explored this more, though, I realize that there are some related fixes and improvements we can make that would be nice to be part of the upcoming major version change. We'd want this to go through quality assurance so I'm not totally sure whether or not it makes the cut but would be interested in getting your impressions on https://github.com/getodk/javarosa/pull/804 @brontolosone

PR #802, opened 2024-10-15
Fixed detecting partial elements that should be populated

grzesiek2010:
Closes https://github.com/getodk/collect/issues/6409

I've added automated tests and tested the fix with Collect.

In Collect our approach to loading a long list of choices (entities) is to initially load only the first element, then populate the rest of the list as needed. See: [LocalEntitiesInstanceAdapter](https://github.com/getodk/collect/blob/master/entities/src/main/java/org/odk/collect/entities/javarosa/intance/LocalEntitiesInstanceAdapter.kt#L18). We refer to these not loaded elements as `partial` ones. For example, if the list should contain 10 elements, we load the first one, while the remaining 9 are simple `TreeElement` placeholders. All elements in the list are marked as partial, even the first one (though this may not be necessary in Collect). 
When we encounter a partial element and need the full list, we load it at that point. The issue we faced was with the mechanism for detecting partial elements - it failed to correctly identify elements without children.
Everything worked fine as long as the first element was still marked as partial, since it had children (as mentioned earlier, it was fully loaded). However, in Collect, the first element could be replaced, causing the flag to change from true to false. Additionally, as noted earlier, it likely doesn't make sense to mark that first element as partial at all. When we later tried to detect other partial elements, none of them had children and therefore couldn't be detected.

I fixed the issue by checking the specific node that needs to be returned in `DataInstance#resolveReference`. The issue with the check at the beginning of the for loop was that elements without children were being ignored.

I believe it should just fix the issue but it would be good to test other forms with entities to make sure the choices are correctly displayed.

The one from https://github.com/getodk/collect/issues/6409

No.


src/test/java/org/javarosa/plugins/InstancePluginTest.java
```diff
@@ -86,21 +86,36 @@ public void instanceProvider_supportsPartialElements() throws IOException, XForm
         assertThat(firstItem.isPartial(), equalTo(true));
         assertThat(firstItem.getNumChildren(), equalTo(2));
         assertThat(firstItem.getChildAt(0).getName(), equalTo("value"));
-        assertThat(firstItem.getChildAt(0).getValue(), equalTo(null));
+        assertThat(firstItem.getChildAt(0).getValue(), equalTo(new StringData("0")));
```
seadowg, line 89: I think I'd prefer to keep the existing tests as simpler and just test that we support partial elements and then have the last test use a different fake (or make the existing one configurable) to set up the special case where the first element in a nodeset is partial.

seadowg (changes requested):

seadowg (approved):

PR #805, opened 2024-10-22
Extend randomize seed tests

brontolosone:
Tests in #804 could serendipitously ~fail~ succeed while they shouldn't since randomization took place over only 2 items (thus randomize(x, y) would result in the same shuffled order with probability 0.5 when varying y).

This expands tests to use a larger collection so that entropy is `8!` rather than 2 when the  total ordering rather than just the identity of the first element of the shuffled list is tested (reducing the "lucky factor" to 1/40320 if I'm not mistaken).

Update: Preserved classic behaviour for the special case of zero-length strings as seed input, and added/modified tests for that behaviour.

### DIFF
--- a/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -705,12 +705,18 @@ public class XPathFuncExpr extends XPathExpression {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lognaturel (approved):
Love these, thank you!

lognaturel:
Do you want to add on the empty case here as well?

lognaturel (approved):

PR #810, opened 2024-12-05
Remove usage of integer type which in JR is alias for int

lognaturel:
I accidentally introduced test forms with the `integer` bind type. This is technically supported (in the xsd namespace) in the W3C XForms spec. JavaRosa defines the alias. Most other tools in the ecosystem would not recognize it so I think we should aim not to use it in tests.


seadowg:
@lognaturel was I added on purpose here?

lognaturel:
Yes please! Sorry, should have added context. I just need a 👍 from a maintainer -- this is removing usage of an uncommon alias. I'd originally tagged @eyelidlessness because he pointed this out as being confusing but his approval won't let us merge.

PR #811, opened 2024-12-11
Spaces after CSV delimiters are not stripped

lognaturel:
Tests illustrating that spaces after CSV delimiters are not stripped. This is something users sometimes run into when they build CSVs by hand. To make matters more confusing, Enketo does strip whitespace so issues are only caught when testing with Collect.

For now I want to make sure we have tests so that the behavior is documented and can't change without being noticed.

I followed the test style in that suite.

### DIFF
--- a/src/test/java/org/javarosa/core/model/instance/CsvExternalInstanceTest.java
+++ b/src/test/java/org/javarosa/core/model/instance/CsvExternalInstanceTest.java
@@ -29,6 +29,12 @@ public class CsvExternalInstanceTest {
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


seadowg (changes requested):
Definitely good to add these tests! How would you feel about reworking them slightly so that each test writes (and then parses) its own CSV file? It's easy enough to track with the diff here, but it's awkward to reason about the test with it not being clear why individual child nodes are being picked out without it.

lognaturel:
Replaced by https://github.com/getodk/javarosa/pull/853

PR #814, opened 2025-01-29
Fixed deserialization of forms containing selects with default answers

grzesiek2010:
Closes #https://github.com/getodk/collect/issues/6571

I've added an automated test and performed manual tests based on the steps described in the issue.

According to my findings, the issue stems from setting the instance name for tree elements in the form instance after deserialization. When reading the form from XML, everything works correctly.  

According to the provided reference:  
image  

For elements in the primary instance, the instance name should be `null`. However, during deserialization, when attempting to read the reference (`ref`) of a select-one question located in the primary instance, the reference appears as:  

```
instance(Internal choices)/data/select-from-secondary-instance
```  

In this case, the instance name is mistakenly set to the ID of a child in the primary instance. According to the comment in the reference, the instance name should be `null`, and the reference should instead be (exactly as it appears when we read form from XML):  

```
/data/select-from-secondary-instance
```  

This difference in TreeReferences triggers an exception because the expected question cannot be found. I believe it worked correctly for a long time only because the `equals` method used for comparing references was faulty. Once that method was fixed in [[this pull request](https://github.com/getodk/javarosa/pull/578)](https://github.com/getodk/javarosa/pull/578), the issue began to occur.

It shouldn't affect users at all. It should only fix the problem with the deserialization of forms containing selects with default answers.

The form is attached to the issue.

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/FormDefSerializationTest.java
+++ b/src/test/java/org/javarosa/core/model/FormDefSerializationTest.java
@@ -24,12 +24,15 @@ import static org.javarosa.test.XFormsElement.body;
 import static org.javarosa.test.XFormsElement.head;
 import static org.javarosa.test.XFormsElement.html;
 import static org.javarosa.test.XFormsElement.input;
+import static org.javarosa.test.XFormsElement.instance;
 import static org.javarosa.test.XFormsElement.mainInstance;
 import static org.javarosa.test.XFormsElement.model;
+import static org.javarosa.test.XFormsElement.select1Dynamic;
 import static org.javarosa.test.XFormsElement.t;
 import static org.javarosa.test.XFormsElement.title;
 
 import java.io.IOException;
+
 import org.javarosa.test.Scenario;
 import org.javarosa.core.util.externalizable.DeserializationException;
 import org.javarosa.xform.parse.XFormParser;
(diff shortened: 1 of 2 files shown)


src/test/java/org/javarosa/core/model/FormDefSerializationTest.java
```diff
@@ -77,6 +80,36 @@ public class FormDefSerializationTest {
...
+                        )
+                    )
+                )
+            ),
+            body(
+                select1Dynamic("/data/select-from-secondary-instance", "instance('secondary-instance')/root/item")
+            ))
+        );
+
+        Scenario deserialized = scenario.serializeAndDeserializeForm();
```
grzesiek2010, line 106: The test failed here.
It would be ideal to have a test that fails when checking the instance name or reference of the `select_one` question. However, the issue occurs only after deserialization, and before any validation can be performed, the deserialization process throws an exception.

src/test/java/org/javarosa/core/model/FormDefSerializationTest.java
```diff
@@ -77,6 +80,32 @@ public class FormDefSerializationTest {
...
+                        )
+                    )
+                )
+            ),
+            body(
+                select1Dynamic("/data/select-from-secondary-instance", "instance('secondary-instance')/root/item")
+            ))
+        );
+
+        scenario.serializeAndDeserializeForm();
```
grzesiek2010, line 106: It would be ideal to have a test that fails when checking the instance name or reference of the `select_one` question. However, the issue occurs only after deserialization, and before any validation can be performed, the deserialization process throws an exception.

lognaturel: This is an integration test and that feels ok. I wish the test didn't end with `_worksCorrectly` but I also don't have a great alternate suggestion so I'm inclined to go with it!

lognaturel:
Wow, this is a blast from the past.

I couldn't immediately figure out how your description as related to default values so I debugged the test myself. I now see that when there's a default value, there's a block in `attachControlsToInstanceData` for going from the serialized selected option to the full internal representation of that option. This requires the reference to the question which is broken after deserialization for the reasons you describe.

I don't really understand why and how the `name` for the `FormInstance` is being set (e.g. to "Internal choices" in this case). This troubles me but I'm inclined not to dig deeper unless you think it's important, @grzesiek2010.

lognaturel (approved):

grzesiek2010:
> I don't really understand why and how the name for the FormInstance is being set (e.g. to "Internal choices" in this case). This troubles me but I'm inclined not to dig deeper unless you think it's important, @grzesiek2010.

I don't fully understand why it was implemented that way, so I can't be completely confident that the fix is safe. Based on my tests and debugging, I'm pretty sure it is, but I wanted to check if you might know something about it.

There is a difference between `TreeElement#instanceName`:
image

and `DataInstance#name`:
image
(`instanceName` should be null for primary instances but `name` should take the form title in such cases)

so setting `TreeElement#instanceName` using `DataInstance#name` in [#setRoot](https://github.com/getodk/javarosa/pull/814/files#diff-a4b12cc7c5cbcc3dd21df9095e7c05257f13bc52162f3c2545208235caf7bc70L95) seemed to me like an obvious bug but maybe there was some rationale behind it.

PR #815, opened 2025-02-04
Fixed deserialization of forms containing triggers with no value

grzesiek2010:
Closes #https://github.com/getodk/collect/issues/6583

I've added an automated test and also performed manual tests.

The difference between using an empty string in a trigger versus a null value was as follows:  

- With an empty string, `SetValueAction#value` was set to `XPathStringLiteral("")`, while `SetValueAction#explicitValue` was `null`. When serialized, `explicitValue` was saved as an empty string, and `value` remained `XPathStringLiteral("")` (see [this code](https://github.com/getodk/javarosa/compare/master...grzesiek2010:javarosa:setvalue_serialize?expand=1#diff-95152bc467c506aae66b25edf2944f8b9b202c0eb2a5594e415e32bfa8ebb7c9L136)).  
- With a null trigger value, `SetValueAction#value` was `null`, and `SetValueAction#explicitValue` was an empty string. During serialization (see the link below), only `explicitValue` was saved, while `value` was skipped.  

During deserialization in the second case, an exception occurred because `value` was missing entirely.  

I found the code unnecessarily complex - having both `value` and `explicitValue` was redundant. To simplify, I removed `explicitValue` from `SetValueAction` and now always use `SetValueAction#value` with `XPathStringLiteral("")`, regardless of whether the trigger contained an empty string or null.  

The form attached at https://github.com/getodk/collect/issues/6583

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/actions/SetValueAction.java
+++ b/src/main/java/org/javarosa/core/model/actions/SetValueAction.java
@@ -37,8 +37,6 @@ public class SetValueAction extends Action {
     /** the value to be assigned to the target when this action is triggered **/
     private XPathExpression value;
 
-    private String explicitValue;
-
     public static final String ELEMENT_NAME = "setvalue";
 
     public SetValueAction() {
@@ -49,12 +47,6 @@ public class SetValueAction extends Action {
         super(ELEMENT_NAME);
         this.target = target;
         this.value = value;
-    }
-
-    public SetValueAction(TreeReference target, String explicitValue) {
-        super(ELEMENT_NAME);
-        this.target = target;
-        this.explicitValue = explicitValue;
     }
 
     public static IElementHandler getHandler() {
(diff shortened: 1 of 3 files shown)


lognaturel:
The reason that there are two paths is because the value can be set two ways:
1) As a literal string:
<setvalue bind="/data/q1">foo('bar')</setvalue>

2) An XPath path:
<setvalue bind="/data/q1" value="foo('bar')" />

In the first case, the literal text "foo('bar')" should be set. In the second case, there should be an error because the foo function isn't known. I believe your change makes these equivalent which they shouldn't be.

If we did want to consider changing this code, we should also change the calling code in `parseSetValueAction`.

Let's address the issue without changing the literal string / XPath path distinction.

grzesiek2010:
The second case hasn't changed at all. The only difference is that when a literal string (case 1) is used I wrap it into `XPathStringLiteral` (as it is a string value either way) so that it is handled in the same way no matter if it's case 1 or 2.
I don't think it's wrong and we have a bunch of tests in [SetValueActionTest](https://github.com/getodk/javarosa/blob/master/src/test/java/org/javarosa/core/model/actions/SetValueActionTest.java#L48) that seem to proof I'm right.

lognaturel:
Aha, thanks for the explanation. I missed this very important line: https://github.com/getodk/javarosa/pull/815/files#diff-b3a15b4196d9bf871e885e41f1eea08ab39ffa54e1b5f3b93de8f041a893bc60R892

lognaturel (approved):

PR #817, opened 2025-02-18
Improve geopoint tests

lognaturel:
Closes #

### DIFF
--- a/src/test/java/org/javarosa/core/model/data/test/GeoPointDataTests.java
+++ b/src/test/java/org/javarosa/core/model/data/test/GeoPointDataTests.java
@@ -1,10 +1,23 @@
 package org.javarosa.core.model.data.test;
 
+import org.javarosa.core.model.data.GeoPointData;
+import org.javarosa.test.Scenario;
+import org.javarosa.xform.parse.XFormParser;
+import org.junit.Test;
+
+import java.io.IOException;
+
 import static org.hamcrest.CoreMatchers.is;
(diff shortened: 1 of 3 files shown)


src/test/java/org/javarosa/core/model/data/test/GeoPointDataTests.java
```diff
@@ -20,4 +33,39 @@ public void getDisplayText_whenAllComponentsAreZero_returnsEmptyString() {
...
+            body(
+                input("/data/q1"),
+                input("/data/note")
+            )
+        ));
+
+        scenario.answer("/data/q1","1.234 5.678");
+        assertThat(scenario.getAnswerNode("/data/note").isRelevant(), is(false));
+
+        scenario.answer("/data/q1", "1.234 5.678 0 0");
```
latin-panda, line 68: Suggestion:

- Having a test where the altitude is not zero: `1.234 5.678 0 0`
- Having another test where values can produce NaN (e.g. then value is nullable, field is not relevant): `ABC 23.55, PP 0`

latin-panda (changes requested):
Looks good to me! 

Do you think the suggestion below makes sense?

src/test/java/org/javarosa/core/model/data/test/GeoPointDataTests.java
latin-panda: @lognaturel, would you prefer to leave this suggestion for another PR?

PR #819, opened 2025-03-14
Add `deprecatedID` and update `instanceID` when editing a finalized form.

grzesiek2010:
I've added automated tests and also tested the changes with ODK Collect.

So far, form initialization has only distinguished between two states (new form/saved form). We need a third state to identify when a finalized form is being edited, ensuring proper handling of `instanceID` and `deprecatedID`.
I don't think there's anything we could add externally as a plugin (as we briefly considered) since it's better for JR to be aware of those three form initialization states.

The implemented changes should be safe, as the new behavior applies only to the `FINALIZED_FORM_EDIT` mode, which has not been used before.

No.

No.


seadowg (commented):
Just a question on how this works with Collect: basically Collect will create a `FormDef` and then add the existing data in via `FormLoaderTask#importData` right? What we're doing here is then having `initialize` manipulate the `instanceId`/`deprecatedId` after that's happened. 

The back and forth there is a little hard to follow (because Collect handles the existing instance load for some reason). I'm wondering if as part of this change we should also move what happens in `importData` into `initialize` (and have the latter take the existing instance as an optional parameter). It seems strange to have JavaRosa now aware of edits but that it needs a client to handle most of the work. I do understand this is blocking though, so if that feels like a heavy lift we can create a separate issue for it.

grzesiek2010:
> Just a question on how this works with Collect: basically Collect will create a FormDef and then add the existing data in via FormLoaderTask#importData right? What we're doing here is then having initialize manipulate the instanceId/deprecatedId after that's happened.

Collect calls `FormDef#initialize` whenever a form is opened, as seen in [[loadDraft](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryUseCases.kt#L98)](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/formentry/FormEntryUseCases.kt#L98). I haven't looked into how `FormDef` is created, as it's outside the scope of what I need.  

This isn't supported in Collect yet, but once this PR is merged, I'll be able to call `FormDef#initialize(FormInitializationMode.FINALIZED_FORM_EDIT)` during editing, allowing JR to adjust the IDs (`instanceID` and `deprecatedID`), and that's all.

> The back and forth there is a little hard to follow (because Collect handles the existing instance load for some reason). I'm wondering if as part of this change we should also move what happens in importData into initialize (and have the latter take the existing instance as an optional parameter). It seems strange to have JavaRosa now aware of edits but that it needs a client to handle most of the work. I do understand this is blocking though, so if that feels like a heavy lift we can create a separate issue for it.

I don't quite follow your reasoning - perhaps we just have different perspectives on edits? In Collect, I started implementing it by cloning the form I want to edit while keeping the original intact: https://github.com/getodk/collect/pull/6646/files#diff-652a7923f1bdcf3c4d1150b5f9ecf17cdcfade959c8f433cc94e8526929cd96fR13.  
Is that approach incorrect? That’s what both the issue and Figma suggest. Thanks to that calling `FormDef#initialize` with a proper parameter is enough.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1097,6 +1099,17 @@ public void initialize(FormInitializationMode formInitializationMode, InstanceIn
...
             preloadInstance(mainInstance.getRoot());
+        } else if (formInitializationMode == FormInitializationMode.FINALIZED_FORM_EDIT) {
+            TreeElement metaSection = mainInstance.getRoot().getFirstChild("meta");
+            if (metaSection != null) {
+                TreeElement instanceId = metaSection.getFirstChild("instanceID");
+                if (instanceId  != null) {
+                    TreeElement deprecatedId = new TreeElement("deprecatedID");
+                    deprecatedId.setValue(instanceId.getValue());
+                    metaSection.addChild(deprecatedId);
+                    instanceId.setValue(new StringData("uuid:" + PropertyUtils.genUUID()));
```
seadowg, line 1110: Should we be setting the `instanceId` explicitly here? What if a form uses a non UUID based `instanceId` (calculated based on data in the form for example)? Would that be valid? It's probably worth checking how this works in Central.

src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -497,5 +498,12 @@ public Object eval(Object[] args, EvaluationContext ec) {
 
         assertThat(deprecatedID, is(originalInstanceID));
         assertThat(newInstanceID, not(originalInstanceID));
+
+        // Subsequent initialization - nothing should change
+        editFinalizedFormScenario = Scenario.from(newFormScenario.getFormDef(), FormInitializationMode.FINALIZED_FORM_EDIT);
```
seadowg, line 503: What situations do we end initializing twice in?

grzesiek2010: I added it just in case. If handled correctly by clients (like ODK Collect), this shouldn't happen.  
In Collect, when editing a form, `FormDef#initialize(FormInitializationMode.FINALIZED_FORM_EDIT)` should be called, and this should be the only case. The only tricky scenario I can think of is when a user edits a form, and while still in the editing process, the app is killed and restored. In that case, we would need to ensure Collect doesn't call `FormDef#initialize(FormInitializationMode.FINALIZED_FORM_EDIT)` again.  
I would need to check that to be sure, but handling it in JR ensures safety.

src/main/java/org/javarosa/core/model/FormDef.java
grzesiek2010: Do people actually do that? I've never seen a form like that before.  
I believe we should continue using `uuid`, as it ensures uniqueness and is a safer approach. If the instance ID is generated based on custom logic, it may not always be unique, and handling it would be more complex.

src/test/java/org/javarosa/core/model/test/FormDefTest.java
seadowg: I think this might be related to https://github.com/getodk/javarosa/pull/819#discussion_r2005390910 actually: depend ending on how we set `instanceId`, it might not matter if `initialize` is called multiple times.

src/main/java/org/javarosa/core/model/FormDef.java
grzesiek2010: You can edit an xml file to achieve something like that but it is not possible with xls right? @lognaturel any thoughts?

src/test/java/org/javarosa/core/model/test/FormDefTest.java
lognaturel: The thing we have to be absolutely sure of is that deprecatedId can only be set exactly once when copying over the original instance to edit.

src/main/java/org/javarosa/core/model/FormDef.java
lognaturel: If I recall correctly, Central does allow any value for instanceid and deprecatedid. The precedent to look at is probably more in other clients. Enketo completely ignores any expression bound to instanceid and always injects a uuid (both for edits and for new submissions). Web Forms does honor alternative bound expressions and just reruns the uuid preload in the edit case.

I think injecting a UUID for the edit case is fine, if folks do inject custom instanceIDs in their form XML I think the likelihood of them considering edits in that expression is very low. Injecting a UUID lets them have a custom id for the base submission and then still build up a coherent chain of edits. I also think it's ok for clients to vary for this edge case because if users are sophisticated enough to want the custom functionality they'll know to ask if it's missing.

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1083,25 +1084,39 @@ private static void getElementsFromReferences(Set<TreeReference> referencesRemai
...
      *
-     * @param newInstance true if the form is to be used for a new entry interaction,
-     *                    false if it is using an existing IDataModel
+     * @param formInitializationMode Specifies the mode in which the form is being initialized:
+     *                               NEW_FORM: The form is being initialized for a new entry interaction.
+     *                               DRAFT_FORM_EDIT: The form is being edited from a saved draft.
+     *                               FINALIZED_FORM_EDIT: The form is being edited after being finalized.
      */
-    public void initialize(boolean newInstance, InstanceInitializationFactory factory) {
+    public void initialize(FormInitializationMode formInitializationMode) {
```
lognaturel, line 1092: This means we're looking at a major version bump

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1083,25 +1084,39 @@ private static void getElementsFromReferences(Set<TreeReference> referencesRemai
     /**
      * meant to be called after deserialization and initialization of handlers
      *
-     * @param newInstance true if the form is to be used for a new entry interaction,
-     *                    false if it is using an existing IDataModel
+     * @param formInitializationMode Specifies the mode in which the form is being initialized:
```
lognaturel, line 1087: Feels like this doc comment should be a link to the type so the values don't have to be inline here where they could become out of sync. I don't think the value comments are necessary since the names are descriptive. You could specifically call out that finalized_form_edit populates deprecatedid

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1083,25 +1084,39 @@ private static void getElementsFromReferences(Set<TreeReference> referencesRemai
...
-    public void initialize(boolean newInstance, InstanceInitializationFactory factory) {
+    public void initialize(FormInitializationMode formInitializationMode) {
         HashMap<String, DataInstance> formInstances = getFormInstances();
         for (String instanceId : formInstances.keySet()) {
             DataInstance instance = formInstances.get(instanceId);
-            instance.initialize(factory, instanceId);
+            instance.initialize(instanceId);
         }
-        if (newInstance) {// only preload new forms (we may have to revisit
+        if (formInitializationMode == FormInitializationMode.NEW_FORM) {// only preload new forms (we may have to revisit
```
lognaturel, line 1098: While here, maybe clean up this comment? Maybe the revisiting was about edits, or if not we don't know what the intent was

src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1083,25 +1084,39 @@ private static void getElementsFromReferences(Set<TreeReference> referencesRemai
...
             // this)
             preloadInstance(mainInstance.getRoot());
+        } else if (formInitializationMode == FormInitializationMode.FINALIZED_FORM_EDIT) {
+            TreeElement metaSection = mainInstance.getRoot().getFirstChild("meta");
+            if (metaSection != null) {
+                TreeElement instanceId = metaSection.getFirstChild("instanceID");
+                TreeElement deprecatedId = metaSection.getFirstChild("deprecatedID");
+                if (instanceId  != null && deprecatedId == null) {
```
lognaturel, line 1106: I think you need code and tests to handle a 2nd or third edit -- a submission that already has a deprecatedid

src/test/java/org/javarosa/core/model/test/FormDefTest.java
lognaturel: Ah, I see, this won't work, I don't think. We need to be able to handle edits to submissions that already have a deprecatedid. It's going to be the client's responsibility to only call FINALIZED_FORM_EDIT exactly once at copy time.

lognaturel (changes requested):

lognaturel:
Overall this looks ok to me! I do want to call out that it will be a major version bump. Could we consider also keeping the boolean variant and marking it as deprecated?

src/test/java/org/javarosa/core/model/test/FormDefTest.java
grzesiek2010: >  We need to be able to handle edits to submissions that already have a deprecatedid.

What scenario would that be? We have this issue: https://github.com/getodk/collect/issues/6594, which states that once a form has been edited, it can't be edited again.

lognaturel: That issue is about making sure that if you already have a 6th edit to an original submission, you can't make edits to that original submission, the 1st edit, the second, etc. You may only continue making edits to edit 6. Edit 6 does have a deprecatedid.

grzesiek2010: Ahh ok so once an edit is finalized it also can be edited, right! Ok so I will unblock that scenario.

src/main/java/org/javarosa/core/model/FormDef.java
grzesiek2010: Done.

src/main/java/org/javarosa/core/model/FormDef.java
grzesiek2010: Done.

src/main/java/org/javarosa/core/model/FormDef.java
grzesiek2010: Done.

src/main/java/org/javarosa/core/model/FormDef.java
seadowg: > I think injecting a UUID for the edit case is fine, if folks do inject custom instanceIDs in their form XML I think the likelihood of them considering edits in that expression is very low. Injecting a UUID lets them have a custom id for the base submission and then still build up a coherent chain of edits. I also think it's ok for clients to vary for this edge case because if users are sophisticated enough to want the custom functionality they'll know to ask if it's missing.

Right that makes sense. I think what web forms does is more "correct", but we can move to that later if we need to and if Enketo injected a UUID then there's certainly precedent for doing it this way.

seadowg (changes requested):
As well as the inline comments, I think we should bump the version in this PR as well just to be safe (as @lognaturel pointed out).

src/main/java/org/javarosa/core/model/FormDef.java
grzesiek2010: The last released version was 4.x but it is already bumped to 5.x so isn't it enough?

src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -463,4 +468,63 @@ public Object eval(Object[] args, EvaluationContext ec) {
...
+                )
+            ),
+            body(
+                input("/data/a")
+            )));
+
+        TreeElement meta  = newFormScenario.getFormDef().getMainInstance().getRoot().getFirstChild("meta");
+        IAnswerData originalInstanceID = meta.getFirstChild("instanceID").getValue();
+
+        Scenario editFinalizedFormScenario = Scenario.from(newFormScenario.getFormDef(), FormInitializationMode.FINALIZED_FORM_EDIT);
```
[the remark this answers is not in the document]
grzesiek2010: Done.

src/main/java/org/javarosa/core/model/FormDef.java
lognaturel: Wow, somehow I didn't tag the v5.0.0 release. It was published and included at https://github.com/getodk/collect/compare/v2024.2.4...v2024.3.0#diff-d4be40bb49cbd2e0529fd0efc0469d2074f8d77b306d8e3ceb62cc395f20e2cdR37 I went back and tagged that commit. It's not entirely clear to me how this is even possible but I guess we can look into it if it happens again.

A related issue is that we forgot to release 5.1.0 so we released Collect v2025.1.0 on a snapshot: https://github.com/getodk/collect/compare/v2024.3.6...v2025.1.0#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfR49 Not the end of the world but I'll work to make sure that doesn't happen again.

Could we consider keeping the original signature and marking it as deprecated, to be removed with the next major version bump? Seems we don't lose much by doing that and it better captures our intent that this is an additive change.

grzesiek2010: > Could we consider keeping the original signature and marking it as deprecated, to be removed with the next major version bump?

Done.

src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -463,4 +469,69 @@ public Object eval(Object[] args, EvaluationContext ec) {
...
+                input("/data/a")
+            )));
+
+        formDef.getMainInstance().getRoot().getFirstChild("meta").getFirstChild("instanceID").setAnswer(new StringData("originalInstanceId"));
+        formDef.initialize(FormInitializationMode.FINALIZED_FORM_EDIT);
+
+        IAnswerData originalInstanceID = formDef.getMainInstance().getRoot().getFirstChild("meta").getFirstChild("instanceID").getValue();
+        IAnswerData originalDeprecatedID = formDef.getMainInstance().getRoot().getFirstChild("meta").getFirstChild("deprecatedID").getValue();
+
+        formDef.initialize(FormInitializationMode.FINALIZED_FORM_EDIT);
```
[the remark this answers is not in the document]
grzesiek2010: Done.

src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -463,4 +469,74 @@ public Object eval(Object[] args, EvaluationContext ec) {
...
+        formDef.getMainInstance().getRoot().getFirstChild("meta").getFirstChild("instanceID").setAnswer(new StringData("originalInstanceId"));
+        formDef.initialize(FormInitializationMode.FINALIZED_FORM_EDIT);
+
+        IAnswerData originalInstanceID = formDef.getMainInstance().getRoot().getFirstChild("meta").getFirstChild("instanceID").getValue();
+        IAnswerData originalDeprecatedID = formDef.getMainInstance().getRoot().getFirstChild("meta").getFirstChild("deprecatedID").getValue();
+
+        formDef = Scenario.createFormDef("Simplest", formXml);
+
+        formDef.getMainInstance().getRoot().getFirstChild("meta").getFirstChild("instanceID").setAnswer(originalInstanceID);
+        formDef.initialize(FormInitializationMode.FINALIZED_FORM_EDIT);
```
seadowg, line 533: A `deprecatedId` (probably `originalDeprecatedId`) needs to be set before this to set up the test properly.

seadowg (changes requested):

src/test/java/org/javarosa/core/model/test/FormDefTest.java
seadowg: Ah of course! Follow the process from my [earlier comment](https://github.com/getodk/javarosa/pull/819#discussion_r2007312071), it would get added by Collect before `initialize` I think (like how `instanceId` would be set)? With the current test I can change the implementation so we only set the `deprecatedId` if it doesn't already exist and everything passes.

seadowg (changes requested):
https://github.com/getodk/javarosa/pull/819#discussion_r2012307308

src/test/java/org/javarosa/core/model/test/FormDefTest.java
grzesiek2010: > it would get added by Collect before initialize I think

The `deprecatedID` will always be added and the `instanceID` updated when the `#initialize` method is called, particularly when editing a finalized form.

> With the current test I can change the implementation so we only set the deprecatedId if it doesn't already exist and everything passes.

As mentioned earlier, the `deprecatedID` will never exist when `#initialize` is called, regardless of whether it's the first or a subsequent edit. The `instanceID` is part of the form’s meta section (in the blank form), so it always exists and has a value because the data was imported earlier. However, the `deprecatedID` does not exist at that point, which is expected. We create it every time `#initialize` is called.

seadowg: > As mentioned earlier, the deprecatedID will never exist when #initialize is called

A finalized instance that's an edit of a finalized form will include a `deprecatedId` in its XML right? If I then try and edit that instance, I think `deprecatedId` would be added into `meta` during [`importData`](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java#L421). I might be misunderstanding how finalized edits will work in Collect though.

seadowg (changes requested):
https://github.com/getodk/javarosa/pull/819#discussion_r2012629914

src/test/java/org/javarosa/core/model/test/FormDefTest.java
grzesiek2010: > A finalized instance that's an edit of a finalized form will include a deprecatedId in its XML right?

Yes a finalized one but not the blank one.

> If I then try and edit that instance, I think deprecatedId would be added into meta during [importData](https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java#L421). I might be misunderstanding how finalized edits will work in Collect though.

No #importData does not add that tree element. It is always added dynamically in #initialize when we edit a finalized form. I could probably update #importData to work like that but I don't think I need it as it seems to work well without it.

seadowg: > No #importData does not add that tree element.

Not explicitly, but `importData` overrides the main instance in the `FormDef` that's created from parsing the blank form ([here](https://github.com/getodk/collect/blob/e263c1edb6f48d47af454d6f9f7b175b2e9ee83d/collect_app/src/main/java/org/odk/collect/android/tasks/FormLoaderTask.java#L488)) which I'm assuming is how data from the saved form is "loaded" into form entry. 

If you debug at line 485 (above the line linked above) and at line 493 you'll see that the `instanceId` is not set, but then it is at 493 (after the root of the main instance has been overridden). My guess is that if a saved form contained a `deprecatedId`, we also end up with that in the main instance. Is that not the case? Do only nodes that exist in the blank form get included? I think it'd be good to confirm, either way because if not, then the code doesn't need to care about `deprecatedId` ever existing (other than in multi `initialize` calls which shouldn't happen, but we could add an explicit test for) and could just be:

```java
} else if (formInitializationMode == FormInitializationMode.FINALIZED_FORM_EDIT) {
    TreeElement metaSection = mainInstance.getRoot().getFirstChild("meta");
    if (metaSection != null) {
        TreeElement instanceId = metaSection.getFirstChild("instanceID");
        if (instanceId != null) {
            TreeElement deprecatedId = metaSection.getFirstChild("deprecatedID");
            deprecatedId = new TreeElement("deprecatedID");
            deprecatedId.setAnswer(instanceId.getValue());
            metaSection.addChild(deprecatedId);
            
            instanceId.setAnswer(new StringData("uuid:" + PropertyUtils.genUUID()));
        }
    }
}
```

Right now there's no test that drives out the changes made in https://github.com/getodk/javarosa/pull/819/commits/ec41a0618bfffb869cc78374bc30348017083848 to make adding `deprecatedId`/ conditional and that's confusing me.

> I could probably update #importData to work like that but I don't think I need it as it seems to work well without it.

No I like the structure you have here! I'm just not confident about how the "import" process works and how it will interact with these changes.

> Plus #importData is a mechanism used in Collect not in JR.

That's true, but I'm assuming we're going to use `importData` for editing finalized instances, so how it operates does affect this.

seadowg (changes requested):

src/test/java/org/javarosa/core/model/test/FormDefTest.java
seadowg: Great! Thanks for looking into it. I guess `importData` must only "import" things if there is an existing node in the blank form that corresponds to them.

> Right now there's no test that drives out the changes made in https://github.com/getodk/javarosa/commit/ec41a0618bfffb869cc78374bc30348017083848 to make adding deprecatedId/ conditional and that's confusing me.

This is still true. Surely if `deprecatedId` will never exist before an `initialize` call we can just simplify the code? Otherwise, the only reason I can see for the current `deprecatedId == null` check is to account for multiple `initialize` calls. Taking that check out doesn't cause any tests to fail currently.

grzesiek2010: Yeah I think we can simplify the code https://github.com/getodk/javarosa/pull/819/commits/2f45b69ac67d625c4b724b044f6bb9839fb99c6b

seadowg (approved):

grzesiek2010:
@lognaturel it’s now awaiting your final review. I’m not rushing you, just letting you know that the long discussion with @seadowg has finally ended :smile:

lognaturel (approved):

lognaturel:
A long but good discussion!

PR #820, opened 2025-04-03
Fixed initialization of the form definition for edits that are already modifications of finalized forms

grzesiek2010:
Closes #

I've improved the tests and tested the fix with ODK Collect.

When we edit a form that is already an edit of a finalized form (it already contains `deprecatedID`) we need to set that id but we don't need to add it.

It doesn't require testing.

No.

No.


seadowg (commented):
@grzesiek2010 is it potentially worth getting this all working in Collect using `./gradlew publishToMavenLocal` before reviewing the JavaRosa changes? That avoids multiple PRs/reviews if things change while implementing on the Collect side.

seadowg (approved):

PR #823, opened 2025-05-08
Move from OSSRH to Central

yanokwa:


lognaturel:
Sigh, Central is too popular of a name. @yanokwa do you know whether it's true that Sonatype Central's gradle support is still unreliable? It would be super nice if we could get rid of Maven for builds.

yanokwa:
There's no official Gradle support. They are [working on it](https://central.sonatype.org/publish/publish-portal-gradle/) and recommend [JReleaser](https://jreleaser.org/guide/latest/examples/maven/maven-central.html#_portal_publisher_api) as an alternative.

Doesn't seem worth it to switch to something third party and then switch again when official support arrives.

lognaturel (approved):

PR #829, opened 2025-09-24
`odk-new-repeat` event in model should always be rejected

lognaturel:
[The spec ](https://getodk.github.io/xforms-spec/#the-odk-new-repeat-event) says "Actions triggered by odk-new-repeat must be nested in the repeat form control." We only enforced this when `odk-new-repeat` was the only triggering event but there can be more than one.

I'm opening this for consideration and communication but I'm not convinced we should merge it. There are forms with dynamic defaults in repeats that were generated by `pyxform` that would stop working in Collect. That could be pretty disruptive. Users would have to reconvert with updated pyxform.

Wrote and ran a new test.

This uses string comparison which would break if a new event that adds to `odk-new-repeat` were introduced. That's highly unlikely so I think it's the simplest solution. Otherwise we could split the space-separated string into a collection.

It's going to break existing forms.

Any form with dynamic defaults in repeats or Entities from repeats.


lognaturel:
Thanks for the pyxform context, @lindsay-stevens!

This continues to feel like a risky change, even if it would be a more correct one one. @seadowg pointed out that if we wanted to do this, we should start by using analytics to see how common the case is. We could use form definition extras to do this.

The extra flexibility isn't harming users so I'm closing this a day before its one year anniversary!

PR #832, opened 2025-09-25
Fix handling date types

### LINKED ISSUE #831: Calculation for `time` field throwing Invalid format error
For the following Form definition, I get the error: `Invalid format: "16:15:00.000-04:00" is malformed at ":15:00.000-04:00"`

```
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" xmlns:odk="http://www.opendatakit.org/xforms">
    <h:head>
        <h:title>timefieldbug</h:title>
        <model odk:xforms-version="1.0.0">
            <instance>
                <data id="timefieldbug" version="20250917095930">
                    <timeout/>
                    <meta>
                        <instanceID/>
                    </meta>
                </data>
            </instance>
            <bind nodeset="/data/timeout" type="time" calculate="&quot;16:15:00.000-04:00&quot;"/>
            <bind nodeset="/data/meta/instanceID" type="string" readonly="true()" jr:preload="uid"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/timeout">
            <label>timeout</label>
        </input>
    </h:body>
</h:html>
```

Note that `time` field (type="time") saves the data in the same format. In Submission XML, I can see `<timeout>16:15:00.000-04:00</timeout>`. I expect `time` field to parse the data in the same format.

Interestingly, time field expects data to include date component as well for parsing, following Form definition works correctly:

```
<?xml version="1.0"?>

<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" xmlns:odk="http://www.opendatakit.org/xforms">
    <h:head>
        <h:title>timefieldbug</h:title>
        <model odk:xforms-version="1.0.0">
            <instance>
                <data id="timefieldbug" version="20250917101105">
                    <timeout/>
                    <meta>
                        <instanceID/>
                    </meta>
                </data>
            </instance>
            <bind nodeset="/data/timeout" type="time" calculate="&quot;2025-01-01T16:15:00.000&quot;"/>
            <bind nodeset="/data/meta/instanceID" type="string" readonly="true()" jr:preload="uid"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/timeout">
            <label>timeout</label>
        </input>
    </h:body>
</h:html>
```

XLSForm: 

[timefieldbug.xlsx](https://github.com/user-attachments/files/22387697/timefieldbug.xlsx)

**Collect Version**: v2025.3.0 Beta 4

grzesiek2010:
Closes #831 

I've added automated tests and tested the fix manually with Collect.

As briefly described in one of the comments on the issue, the problem was that sometimes the wrong data type was returned for a given question, for example, `StringData` instead of `TimeData`. I couldn’t think of any other approach. We needed to handle the data types returned by `time`, `date`, and `dateTime` questions correctly, and that’s what I did.

This should just fix the issue, but it would be good to test the three related question types for regressions (`time`, `date` and `dateTime`).

[dateTimes.xlsx](https://github.com/user-attachments/files/22548597/dateTimes.xlsx)

No.

### DIFF
--- a/src/test/java/org/javarosa/core/model/DateTimeTest.java
+++ b/src/test/java/org/javarosa/core/model/DateTimeTest.java
@@ -13,8 +13,14 @@ import java.io.IOException;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.javarosa.test.BindBuilderXFormsElement.bind;
-import static org.javarosa.test.XFormsElement.*;
+import static org.javarosa.test.XFormsElement.body;
+import static org.javarosa.test.XFormsElement.head;
+import static org.javarosa.test.XFormsElement.html;
 import static org.javarosa.test.XFormsElement.input;
+import static org.javarosa.test.XFormsElement.mainInstance;
+import static org.javarosa.test.XFormsElement.model;
+import static org.javarosa.test.XFormsElement.t;
+import static org.javarosa.test.XFormsElement.title;
 
 public class DateTimeTest {
     @Test
--- a/src/test/java/org/javarosa/core/model/DateTimeTest.java
+++ b/src/test/java/org/javarosa/core/model/DateTimeTest.java
@@ -30,28 +30,28 @@ public class DateTimeTest {
                 title("Time form"),
                 model(
                     mainInstance(t("data id=\"time-form\"",
-                        t("time1"),
-                        t("time2", "23:14:00.000+02:00"),
-                        t("time3")
+                        t("calculateLiteral"),
+                        t("empty", "23:14:00.000+02:00"),
+                        t("calculateReference")
                     )),
-                    bind("/data/time1").type("time").calculate("&quot;23:14:00.000+02:00&quot;"),
-                    bind("/data/time2").type("time"),
-                    bind("/data/time3").type("time").calculate("/data/time2")
+                    bind("/data/calculateLiteral").type("time").calculate("&quot;23:14:00.000+02:00&quot;"),
+                    bind("/data/empty").type("time"),
+                    bind("/data/calculateReference").type("time").calculate("/data/empty")
                 )
             ),
             body(
-                input("/data/time1"),
-                input("/data/time2"),
-                input("/data/time3")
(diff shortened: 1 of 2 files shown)


src/test/java/org/javarosa/core/model/DateTimeTest.java
```diff
@@ -0,0 +1,121 @@
...
+                        t("time2", "23:14:00.000+02:00"),
+                        t("time3")
+                    )),
+                    bind("/data/time1").type("time").calculate("&quot;23:14:00.000+02:00&quot;"),
+                    bind("/data/time2").type("time"),
+                    bind("/data/time3").type("time").calculate("/data/time2")
+                )
+            ),
+            body(
+                input("/data/time1"),
```
seadowg, line 43: Just to make these tests a little more readable, maybe we could name the questions around what they test. Instead of `time1` and so on, it could be:

- `calculateLiteral`
- `empty`
- `calculateReference`

seadowg (approved):

src/test/java/org/javarosa/core/model/DateTimeTest.java
grzesiek2010: Done.

seadowg:
@grzesiek2010 we'll want to do a new release and update Collect right?

seadowg:
It was originally reported on the forum: https://forum.getodk.org/t/odk-entities-troubleshooting-outage-reporting-create-update-in-one-form/56573. I'm thinking we should do a hotfix with this unless it feels too risky for some reason.

PR #840, opened 2025-12-19
Add equals to geo types

seadowg:
This simplifies testing components that use these types.

### DIFF
--- a/src/main/java/org/javarosa/core/model/data/GeoPointData.java
+++ b/src/main/java/org/javarosa/core/model/data/GeoPointData.java
@@ -25,7 +25,9 @@ import org.jetbrains.annotations.NotNull;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
 
 /**
@@ -190,8 +192,19 @@ public class GeoPointData implements IAnswerData, IExprDataType {
         return getDisplayText();
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (o == null || getClass() != o.getClass()) return false;
+        GeoPointData that = (GeoPointData) o;
+        return len == that.len && Objects.deepEquals(gp, that.gp);
+    }
 
-  public double getPart(int i) {
+    @Override
+    public int hashCode() {
+        return Objects.hash(Arrays.hashCode(gp), len);
+    }
+
+    public double getPart(int i) {
     if (i < len) {
       return gp[i];
     } else {
--- a/src/main/java/org/javarosa/core/model/data/GeoPointData.java
+++ b/src/main/java/org/javarosa/core/model/data/GeoPointData.java
@@ -27,7 +27,6 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
 
 
 /**
@@ -193,15 +192,18 @@ public class GeoPointData implements IAnswerData, IExprDataType {
     }
 
     @Override
-    public boolean equals(Object o) {
-        if (o == null || getClass() != o.getClass()) return false;
+    public final boolean equals(Object o) {
+        if (!(o instanceof GeoPointData)) return false;
+
         GeoPointData that = (GeoPointData) o;
-        return len == that.len && Objects.deepEquals(gp, that.gp);
+        return len == that.len && Arrays.equals(gp, that.gp);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(Arrays.hashCode(gp), len);
+        int result = Arrays.hashCode(gp);
+        result = 31 * result + len;
+        return result;
     }
 
     public double getPart(int i) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/javarosa/core/model/data/GeoPointData.java
```diff
@@ -190,8 +192,19 @@ public String toString() {
         return getDisplayText();
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (o == null || getClass() != o.getClass()) return false;
```
lognaturel: I'm somewhat surprised to see typed equality. Is that what the `equals` methods for other types do? They don't first try to convert to the target type and only return false if that fails?

seadowg: > They don't first try to convert to the target type and only return false if that fails?

I'm not sure that I understand you here. This checks that the `this` and `o` have exactly the same type and then returns false if not. If that return doesn't happen, `o` is converted and then the `points` are compared. Is that not equivalent to what you're describing? 

This code was generated by IntelliJ's equals/hashcode generator and it does the exact type match vs the polymorphic `instanceof` check. You can try generating the code yourself with right click > Generate > "equals() and hashCode()". There's also tests checking this gives us the equality functionality we need.

lognaturel: I'm across too many languages with different type systems so sorry if this is a nonsense question.

If you compare `StringData(1 1 1 1)` to the `GeoPointData` equivalent, the `equals` check as written here will fail even if they're both referenced by vars of type `IAnswerData` I believe, right? Is that's what is desired?

lognaturel: With XPath type coercion (which is admittedly idiosyncratic), the two should be equivalent so that's why this makes me nervous. I know form-level equality expressions are evaluated elsewhere but having two different definitions of equality makes me nervous.

lognaturel: Ok but I guess because `.equals` and `hashCode` are Java concepts they should follow Java's typing so it wouldn't make sense to try casting first, that's just the XPath expectation.

Could you consider using display string equality to avoid this clash of type systems? Or maybe we are confident this would never be used by form evaluating code so maybe it doesn't matter.

lognaturel (approved):
I was initially expecting XPath style equality but I now realize these `equals` methods are not used for form evaluation. I've confirmed that other subclasses of `IAnswerData` have these.

src/main/java/org/javarosa/core/model/data/GeoTraceData.java
```diff
@@ -205,4 +207,16 @@ public Double toNumeric() {
     public String toString() {
         return getDisplayText();
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == null || getClass() != o.getClass()) return false;
```
grzesiek2010: You could add a reference equality check as the first step:
`if (this == o) return true;`

src/main/java/org/javarosa/core/model/data/GeoPointData.java
```diff
@@ -190,8 +192,19 @@ public String toString() {
         return getDisplayText();
     }
 
+    @Override
+    public boolean equals(Object o) {
```
grzesiek2010: There are tests for `GeoShapeData` and `GeoTraceData` but not for `GeoPointData`. Was that intentional to test `GeoPointData` only with shape and trace?

src/main/java/org/javarosa/core/model/data/GeoShapeData.java
```diff
@@ -199,4 +201,16 @@ public Double toNumeric() {
     public String toString() {
         return getDisplayText();
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == null || getClass() != o.getClass()) return false;
```
grzesiek2010: You could add a reference equality check as the first step:
`if (this == o) return true;`

src/main/java/org/javarosa/core/model/data/GeoPointData.java
grzesiek2010: You could add a reference equality check as the first step:
`if (this == o) return true;`

grzesiek2010 (changes requested):

src/main/java/org/javarosa/core/model/data/GeoPointData.java
seadowg: > If you compare StringData(1 1 1 1) to the GeoPointData equivalent, the equals check as written here will fail even if they're both referenced by vars of type IAnswerData I believe, right? Is that's what is desired?

Ah I get you! So yes, it might be desirable to have `StringData` and `GeoPointData` with the same "raw" value be equivalent, but that's currently not the case anyway - because `equals` wasn't implemented it was only ever using reference equality before.

src/main/java/org/javarosa/core/model/data/GeoTraceData.java
seadowg: I'd like to just use the generated code if possible as it means that you can regenerate if you make a change to the class without a significant diff. I've regenerated the equals/hashcode methods using the "IntelliJ default" setting which seems like the simplest.

grzesiek2010 (approved):

PR #841, opened 2026-02-11
Add support for tick itemset for range

lognaturel:
In support of https://github.com/getodk/collect/issues/7013

Added and ran tests.

We considered having a special `tick-labelset` attribute to specify the nodeset with label items but @garethbowen nudged us to consider reusing the `itemset` child pattern already used for selects and rank.

The names of methods in JavaRosa don't feel quite right for this because they're focused on `itemset`s defining select choices but I think it's understandable enough. I considered adding a `getChildItems` method to `FormEntryPrompt` that just delegates to `getSelectChoices` but felt that didn't add enough to justify further expanding the API. 

It's a focused, additive change so risk is low. There should be no change to existing behavior. Collect should now be able to use `getSelectChoices` on a `range` to get its tick labels.

See test. This will be hard to validate outside of a test before Collect has support.

Not yet.

### DIFF
--- a/src/main/java/org/javarosa/xform/parse/XFormParser.java
+++ b/src/main/java/org/javarosa/xform/parse/XFormParser.java
@@ -1170,13 +1170,6 @@ public class XFormParser implements IXFormParserFunctions {
             }
         }
 
-        final boolean hasItems =
-            controlType == CONTROL_SELECT_MULTI
-                || controlType == CONTROL_RANK
-                || controlType == CONTROL_SELECT_ONE;
-
-        final boolean hasOptionalItemset = controlType == CONTROL_RANGE;
-
         question.setControlType(controlType);
         question.setAppearanceAttr(e.getAttributeValue(null, APPEARANCE_ATTR));
 
@@ -1189,15 +1182,16 @@ public class XFormParser implements IXFormParserFunctions {
                 parseQuestionLabel(question, child);
             } else if ("hint".equals(childName)) {
                 parseHint(question, child);
-            } else if (hasItems && "item".equals(childName)) {
+            } else if ((hasRequiredItems(controlType) || hasOptionalItems(controlType)) && "item".equals(childName)) {
                 parseItem(question, child);
-            } else if ((hasItems || hasOptionalItemset) && "itemset".equals(childName)) {
+            } else if ((hasRequiredItems(controlType) || hasOptionalItems(controlType)) && "itemset".equals(childName)) {
                 parseItemset(question, child, parent);
             } else if (actionHandlers.containsKey(childName)) {
                 actionHandlers.get(childName).handle(this, child, question);
             }
         }
-        if (hasItems) {
+
+        if (hasRequiredItems(controlType)) {
             if (question.getNumChoices() > 0 && question.getDynamicChoices() != null) {
                 throw new XFormParseException("Select question contains both literal choices and <itemset>");
             } else if (question.getNumChoices() == 0 && question.getDynamicChoices() == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1170,11 +1170,13 @@ private QuestionDef parseControl(IFormElement parent, Element e, int controlType
             }
         }
 
-        boolean isItem =
+        final boolean hasItems =
             controlType == CONTROL_SELECT_MULTI
                 || controlType == CONTROL_RANK
                 || controlType == CONTROL_SELECT_ONE;
 
+        final boolean hasOptionalItemset = controlType == CONTROL_RANGE;
```
lognaturel: This is "itemset" whereas the one above is "items" because we're not allowing inline items for range.

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1187,15 +1189,15 @@ private QuestionDef parseControl(IFormElement parent, Element e, int controlType
...
                 parseItem(question, child);
-            } else if (isItem && "itemset".equals(childName)) {
+            } else if ((hasItems || hasOptionalItemset) && "itemset".equals(childName)) {
                 parseItemset(question, child, parent);
             } else if (actionHandlers.containsKey(childName)) {
                 actionHandlers.get(childName).handle(this, child, question);
             }
         }
-        if (isItem) {
+        if (hasItems) {
```
lognaturel: For controls that have optional items, we don't want to do this extra validation.

src/test/java/org/javarosa/xform/parse/XFormParserTest.java
```diff
@@ -1,37 +1,7 @@
...
-import static org.junit.Assert.assertNull;
-
-import java.io.File;
-import java.io.IOException;
-import java.io.UncheckedIOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.List;
 import org.javarosa.core.model.FormDef;
+import org.javarosa.core.model.FormIndex;
```
lognaturel, line 4: Again with the import noise. As you can see, my config matches the import order in `XFormParser`. At some point we must have ended up with different configs. I think this is the more common order but I could also turn off automatic import ordering if we think these changes are too noisy.

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1187,15 +1189,15 @@ private QuestionDef parseControl(IFormElement parent, Element e, int controlType
...
-            } else if (isItem && "item".equals(childName)) {
+            } else if (hasItems && "item".equals(childName)) {
                 parseItem(question, child);
-            } else if (isItem && "itemset".equals(childName)) {
+            } else if ((hasItems || hasOptionalItemset) && "itemset".equals(childName)) {
                 parseItemset(question, child, parent);
             } else if (actionHandlers.containsKey(childName)) {
                 actionHandlers.get(childName).handle(this, child, question);
             }
         }
```
seadowg, line 1192: Nitpick: could you add whitespace here?

src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1170,11 +1170,13 @@ private QuestionDef parseControl(IFormElement parent, Element e, int controlType
...
 
-        boolean isItem =
+        final boolean hasItems =
```
seadowg: I think it'd be good to take the opportunity to pull out helpers for at least these `has...` methods.

src/test/java/org/javarosa/xform/parse/XFormParserTest.java
seadowg: I get no changes when doing "Optimize imports" so it's probably just that we're both using the default. That works for our style checks so it's all good I think!

seadowg (approved):

src/main/java/org/javarosa/xform/parse/XFormParser.java
lognaturel: Realized we don't have a great reason for this so making it consistent instead.

PR #842, opened 2026-02-25
Fix string representation of geotrace and geoshape data

grzesiek2010:
I've tested the changes with Collect.

Points should be separated by semicolons, without any whitespace between them.

It should simply fix the string representation of getotrace/geoshape data.

Any for with geotrace/geoshape questions.

No.


seadowg (approved):
Nice fix! Seems really off to me that `getDisplayText` is used to serialize these data types, but it does look like that's the case (from `XFormAnswerDataSerializer`).

PR #843, opened 2026-03-12
Add support for new tick-interval and placeholder range parameters

grzesiek2010:
Added and ran tests.

Nothing important to discuss here. I've just added support for the new parameters in the same way it already existed for the existing ones.

It doesn't require testing.

No.

No.

### DIFF
--- a/src/test/resources/range-form-itemset.xml
+++ b/src/test/resources/range-form-itemset.xml
@@ -28,7 +28,7 @@
     </h:head>
 
     <h:body>
-        <range ref="/stats/balance" start="-2.0" end="2.0" step="0.5">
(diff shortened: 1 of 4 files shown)


src/test/resources/range-form-itemset.xml
```diff
@@ -28,7 +28,7 @@
     </h:head>
 
     <h:body>
-        <range ref="/stats/balance" start="-2.0" end="2.0" step="0.5">
+        <range ref="/stats/balance" start="-2.0" end="2.0" step="0.5" tick-interval="2.0" placeholder="0.0">
```
seadowg: Probably better to use a non-zero value for the test here.

seadowg (changes requested):

seadowg (approved):

PR #845, opened 2026-03-19
Do not remove answers that are not in choices list for RangeQuestion

grzesiek2010:
I've added automated tests and also tested the fix with Collect.

Comments below.

It shouldn’t cause any regression in the features we currently support. Once range questions with labels become available in Collect, this should prevent removing answers that don’t have corresponding choices.

The one from https://github.com/getodk/javarosa/pull/841 (from tests), for example.

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/main/java/org/javarosa/core/model/ItemsetBinding.java
@@ -2,9 +2,7 @@ package org.javarosa.core.model;
 
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IConditionExpr;
-import org.javarosa.core.model.data.DecimalData;
 import org.javarosa.core.model.data.IAnswerData;
-import org.javarosa.core.model.data.IntegerData;
 import org.javarosa.core.model.data.MultipleItemsData;
 import org.javarosa.core.model.data.SelectOneData;
 import org.javarosa.core.model.data.helper.Selection;
@@ -257,12 +255,8 @@ public class ItemsetBinding implements Externalizable, Localizable {
      */
     private void updateQuestionAnswerInModel(FormDef formDef, TreeReference curQRef, Map<String, SelectChoice> selectChoicesForAnswer) {
         IAnswerData originalValue = formDef.getMainInstance().resolveReference(curQRef).getValue();
-        if (originalValue instanceof IntegerData || originalValue instanceof DecimalData) {
-            // Do not perform any updates for RangeQuestion with choices
-            return;
-        }
-        
-        if (selectChoicesForAnswer != null) {
+
+        if (originalValue instanceof MultipleItemsData || originalValue instanceof SelectOneData) {
             IAnswerData boundAndFilteredValue;
             if (originalValue instanceof MultipleItemsData) {
                 boundAndFilteredValue = getFilteredAndBoundSelections((MultipleItemsData) originalValue, selectChoicesForAnswer);
--- a/src/main/java/org/javarosa/core/model/ItemsetBinding.java
+++ b/src/main/java/org/javarosa/core/model/ItemsetBinding.java
@@ -2,7 +2,9 @@ package org.javarosa.core.model;
 
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IConditionExpr;
+import org.javarosa.core.model.data.DecimalData;
 import org.javarosa.core.model.data.IAnswerData;
+import org.javarosa.core.model.data.IntegerData;
 import org.javarosa.core.model.data.MultipleItemsData;
 import org.javarosa.core.model.data.SelectOneData;
 import org.javarosa.core.model.data.helper.Selection;
@@ -255,7 +257,11 @@ public class ItemsetBinding implements Externalizable, Localizable {
      */
     private void updateQuestionAnswerInModel(FormDef formDef, TreeReference curQRef, Map<String, SelectChoice> selectChoicesForAnswer) {
         IAnswerData originalValue = formDef.getMainInstance().resolveReference(curQRef).getValue();
-
+        if (originalValue instanceof IntegerData || originalValue instanceof DecimalData) {
+            // Do not perform any updates for RangeQuestion with choices
+            return;
+        }
+        
         if (selectChoicesForAnswer != null) {
             IAnswerData boundAndFilteredValue;
             if (originalValue instanceof MultipleItemsData) {
--- a/src/test/java/org/javarosa/core/model/RangeQuestionTest.java
+++ b/src/test/java/org/javarosa/core/model/RangeQuestionTest.java
@@ -20,11 +20,11 @@ import static org.javarosa.test.XFormsElement.title;
 public class RangeQuestionTest {
     @Test
     public void answerIsPreservedWhenRangeQuestionHasIncompleteChoices() throws Exception {
-        Scenario scenario = Scenario.init("indefinite repeat", html(
+        Scenario scenario = Scenario.init("range question with incomplete choices", html(
             head(
-                title("Indefinite repeat"),
+                title("Range question with incomplete choices"),
                 model(
-                    mainInstance(t("data id=\"indefinite-repeat\"",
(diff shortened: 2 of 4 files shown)


src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -255,7 +257,11 @@ private Map<String, SelectChoice> initializeAnswerMap(FormDef formDef, TreeRefer
      */
     private void updateQuestionAnswerInModel(FormDef formDef, TreeReference curQRef, Map<String, SelectChoice> selectChoicesForAnswer) {
         IAnswerData originalValue = formDef.getMainInstance().resolveReference(curQRef).getValue();
-
+        if (originalValue instanceof IntegerData || originalValue instanceof DecimalData) {
```
grzesiek2010: It would be better to check the question control type to ensure it’s a RangeQuestion, but that would require additional changes since it’s not easily accessible here. So this might be a simpler solution, which I hope is sufficient in this case.

seadowg: Are all other question types that use `ItemsetBinding` using the `string` data type? I can't think of a test that breaks this if so.

lognaturel: Agreed, this makes me nervous. I think the select bind types are indeed aliases for string so this is likely to work but it smells! Maybe this functionality is in the wrong place and the itemset binding could be responsible just for telling the question that choices have changed and the question could filter out an invalid value if appropriate?

grzesiek2010: @lognaturel do you like this solution better: [https://github.com/getodk/javarosa/pull/846](https://github.com/getodk/javarosa/pull/846), or did you have something else in mind?

I’m a bit torn. On one hand, moving some code out of `ItemsetBinding` makes sense. On the other hand, we currently have a single `QuestionDef` for almost all question types (except range), so moving choice-related logic there doesn’t feel quite right.

Ideally, it would be better to introduce specialized classes for those question types, but that would require more extensive changes/refactoring.

grzesiek2010: An alternative approach, if we want to keep the changes here limited (the first solution), would be to check for `SelectOneData`, `MultipleItemsData`, or `StringData` instead of checking whether `originalValue instanceof IntegerData || originalValue instanceof DecimalData` and returning early.

This wouldn’t require any refactoring, and checking for the data types we expect makes a bit more sense than checking for the ones we want to exclude.

grzesiek2010: @lognaturel @seadowg 
Let's maybe fix this with the smallest possible change:

> An alternative approach, if we want to keep the changes here limited (the first solution), would be to check for SelectOneData, MultipleItemsData, or StringData instead of checking whether originalValue instanceof IntegerData || originalValue instanceof DecimalData and returning early.

and spend more time on it once we add support for itemsets for geo questions, as that feature will require similar changes.
I’ve mentioned this in a few places already, but it will give us a broader perspective on the issue and unblock me so I can finish the new range widget for Collect v2026.2.

seadowg: Is it not possible to derive the question type here from the `FormDef` and the `TreeReference`? That would at least feel safer, although I'd argue this current solution is fine unless someone can think of a test that would break it.

lognaturel: I do prefer the general shape of #846. I agree that it's weird to move select-specific functionality in a general question class but there already is a bunch of select code in there. Maybe we could consider pulling out something like a `HasItems` interface with default methods (basically a mixin).

If you two prefer something like this to start, that seems fine. Either way, I'd like to request a test with a secondary instance that has integer values if there isn't one already. The answer value is stored as a string but I think it would be good to have an explicit test (e.g. name: 1, label: "Yes"; name: 2, label: "No")

lognaturel: Ok, let's go with checking for the types we want to include, add the test I mentioned and close #846 for now.

lognaturel: To expand a bit on the test that would be nice to see. It would be good to have a few options with numeric values, select one, filter the list so that the previously-selected option is no longer in the list, confirm that the value is blanked out. Pretty sure we don't have a test like that yet.

src/main/java/org/javarosa/core/model/ItemsetBinding.java
```diff
@@ -256,7 +256,7 @@ private Map<String, SelectChoice> initializeAnswerMap(FormDef formDef, TreeRefer
     private void updateQuestionAnswerInModel(FormDef formDef, TreeReference curQRef, Map<String, SelectChoice> selectChoicesForAnswer) {
         IAnswerData originalValue = formDef.getMainInstance().resolveReference(curQRef).getValue();
 
-        if (selectChoicesForAnswer != null) {
+        if (originalValue instanceof MultipleItemsData || originalValue instanceof SelectOneData) {
```
grzesiek2010, line 259: I believe only these two types should be possible here. I had some failing tests that relied on `StringData`, but I think that was just a bug in the tests - saving string answers instead of `SelectOneData`. This shouldn’t happen in real use, right?

src/test/java/org/javarosa/core/model/SelectOneChoiceFilterTest.java
```diff
@@ -99,8 +99,8 @@ public void clearingValueAtLevel1_ShouldClearChoicesAtLevels2And3() {
         assertThat(scenario.choicesOf("/data/level2"), empty());
         assertThat(scenario.choicesOf("/data/level3"), empty());
 
-        scenario.answer("/data/level1", "a");
-        scenario.answer("/data/level2", "aa");
+        scenario.answer("/data/level1", scenario.choicesOf("/data/level1").get(0));
```
grzesiek2010, line 102: There are many more tests we could update since they save string answers for select questions, but I only updated the ones that were failing.

src/main/java/org/javarosa/core/model/ItemsetBinding.java
grzesiek2010: There are tests for filtering and clearing answers if corresponding choices are no longer available, like `changingValueAtLevel2_ShouldClearLevel3_IfChoiceNoLongerAvailable`.

I’m not quite sure why we would need to test numeric values specifically when updating this part of the code. If I’m correct that only `MultipleItemsData` and `SelectOneData` can occur here (see this comment: https://github.com/getodk/javarosa/pull/845#discussion_r2994942205), then those data types have been using strings under the hood from the beginning. If there were any issues with that, they would likely have shown up elsewhere already.

src/main/java/org/javarosa/core/model/ItemsetBinding.java
lognaturel: This is related to why I was asking for that test above! We don't really know with certainty that all code paths use the select control types to determine the data type. Even if we're confident all paths within JavaRosa do, could be accidentally doing something that means the data types are sometimes not select item data (as `Scenario` does). For example, I think we should also test serializing and deserializing a submission.

src/test/java/org/javarosa/core/model/RangeQuestionTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.javarosa.test.XFormsElement.item;
+import static org.javarosa.test.XFormsElement.mainInstance;
+import static org.javarosa.test.XFormsElement.model;
+import static org.javarosa.test.XFormsElement.t;
+import static org.javarosa.test.XFormsElement.title;
+
+public class RangeQuestionTest {
+    @Test
+    public void answerIsPreservedWhenRangeQuestionHasIncompleteChoices() throws Exception {
+        Scenario scenario = Scenario.init("indefinite repeat", html(
```
lognaturel: Update this, title, formid

lognaturel (approved):
Approving to unblock follow-on work but I'm not 100% sure that we should release with this, will keep thinking about the potential risk.

PR #849, opened 2026-04-29
Use Java implementation of UUID

seadowg:
As far as I could see, the old implementation did indeed meet the spec for a version 4 UUID, but I don't see a reason for us to take on the risk of rolling our own solution here. My guess would be that this code might have been written before `UUID` was added or widely adopted.


lognaturel (approved):
This does lead to an improvement because `UUID.randomUUID` uses `SecureRandom`, guaranteeing higher entropy and cryptographic security. Not a big deal in this context but agree it's best to leverage the core lib when we can.

PR #851, opened 2026-05-19
Allow any question to optionally have choices

seadowg:
Work towards https://github.com/getodk/collect/issues/7118

I've tried this out with my draft PR (https://github.com/getodk/collect/pull/7235) for https://github.com/getodk/collect/issues/7118 to check that geo questions now have attached choices.

We discussed just adding the geo questions to the optional list, but it honestly felt like the more complex solution given that all questions (`FormEntryPrompt` instances) have a `selectChoices` property.

There should not be any effect on behaviour other than allowing us to build the new features in https://github.com/getodk/collect/issues/7118. I've actually only had to delete untested code to get this working (the test I added is backfilling missing coverage).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,8 +122,8 @@ if (!project.hasProperty("android")) {
         compileSdk = 36
 
         defaultConfig {
-            minSdk = 21
-            targetSdk = 33
+            minSdk = 26
+            targetSdk = 36
             versionCode = 1
             versionName = "1.0"
             testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
@@ -131,8 +131,8 @@ if (!project.hasProperty("android")) {
 
         compileOptions {
             coreLibraryDesugaringEnabled = true // For APIs requiring > API 21
-            sourceCompatibility = JavaVersion.VERSION_1_8
-            targetCompatibility = JavaVersion.VERSION_1_8
+            sourceCompatibility = JavaVersion.VERSION_17
+            targetCompatibility = JavaVersion.VERSION_17
         }
 
         buildTypes {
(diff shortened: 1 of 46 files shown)


src/test/java/org/javarosa/core/model/OptionalChoicesQuestionTest.java
```diff
@@ -17,9 +17,9 @@
 import static org.javarosa.test.XFormsElement.t;
 import static org.javarosa.test.XFormsElement.title;
 
-public class RangeQuestionTest {
+public class OptionalChoicesQuestionTest {
     @Test
-    public void answerIsPreservedWhenRangeQuestionHasIncompleteChoices() throws Exception {
+    public void answerIsPreservedWhenQuestionHasIncompleteItemsetChoices() throws Exception {
         Scenario scenario = Scenario.init("range question with incomplete choices", html(
```
grzesiek2010: Let's update titles and ids here and in the new test below to match their test titles.

grzesiek2010 (changes requested):

src/test/java/org/javarosa/core/model/OptionalChoicesQuestionTest.java
seadowg: These titles/IDs don't actually server any purpose. I'll make changes so that we don't need the boilerplate for them.

build.gradle
```diff
@@ -122,17 +122,17 @@ if (!project.hasProperty("android")) {
         compileSdk = 36
 
         defaultConfig {
-            minSdk = 21
-            targetSdk = 33
+            minSdk = 26
```
seadowg, line 125: This was out of date and we need 26 for the Java standard lib temp directory.

grzesiek2010 (approved):

PR #852, opened 2026-06-05
Handle unparseable date, time and dateTime values gracefully in wrapData

grzesiek2010:
Closes #https://github.com/getodk/collect/issues/7239

I've added unit tests in `DateTimeTest` covering all three temporal types (`date`, `time`, `dateTime`) and also tested the changes manually with Collect.

`DateUtils.parseDate/parseTime/parseDateTime` already signal `unparseable` by returning `null`. The bug was that `wrapData` fed that null straight into a `DateData`/`TimeData`/`DateTimeData` constructor, whose `setValue` rejects `null`. Returning `null` from `wrapData` instead is the smallest, most consistent fix: it matches how `wrapData` already handles empty strings, and it honours the `IAnswerData` contract that a non-existent response is represented by a `null` reference (not a data object wrapping `null`).

When a calculation yields a value that can't be parsed as a `date`/`time`/`dateTime`, the target node is now left blank instead of throwing an exception that might cause a crash.

There is one linked in https://github.com/getodk/collect/issues/7239

No.

### DIFF
--- a/src/main/java/org/javarosa/core/model/data/IAnswerData.java
+++ b/src/main/java/org/javarosa/core/model/data/IAnswerData.java
@@ -111,7 +111,8 @@ public interface IAnswerData extends Externalizable {
             return new MultipleItemsData().cast(new UncastData(String.valueOf(val)));
         } else if (dataType == TIME) {
             if (val instanceof String) {
-                return new TimeData().cast(new UncastData(String.valueOf(val)));
+                Date time = DateUtils.parseTime(String.valueOf(val));
+                return time == null ? null : new TimeData(time);
             } else {
                 return new TimeData((Date) val);
             }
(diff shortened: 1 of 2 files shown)


src/main/java/org/javarosa/core/model/data/IAnswerData.java
```diff
@@ -111,19 +111,22 @@ static IAnswerData wrapData(Object val, int intDataType) {
             return new MultipleItemsData().cast(new UncastData(String.valueOf(val)));
         } else if (dataType == TIME) {
             if (val instanceof String) {
-                return new TimeData().cast(new UncastData(String.valueOf(val)));
+                Date time = DateUtils.parseTime(String.valueOf(val));
```
seadowg, line 114: Just to make sure I'm understanding this correctly: we don't modify the `cast` behaviour (to just return `null`) as `cast` should always succeed or throw an `IllegalArgumentException`?

grzesiek2010: Yes, exactly, your understanding is correct.

seadowg (approved):

PR #854, opened 2026-06-19
Display time component of datetime

### LINKED ISSUE #282: User sees additional time for saved form with time component of now()
#### Problem description
User sees additional time for saved form with time component of now()

1. Start form: [ShowTime.xml.txt](https://github.com/opendatakit/javarosa/files/1837940/ShowTime.xml.txt)
2. User sees 

| | |
| --- | --- |
| screenshot_20180322-153427 | screenshot_20180322-153430|

3. User saves form
4. User reopens saved form

| | |
| --- | --- |
|screenshot_20180322-153442 | screenshot_20180322-153438 | 

#### Expected behavior
User does not see additional time for saved form

### LINKED ISSUE #680: Time component of dateTime not included when displayed
This seems like a duplicate of https://github.com/getodk/javarosa/issues/226 or at least related.

type | name | label
-- | -- | --
datetime | start_datetime | Start datetime
note | note | start_datetime=${start_datetime}

Only the date component is shown. Collect v2022.2

lognaturel:
Closes #282 
Closes #680

Newly-added tests.

This is the least-invasive, least-risky approach I could think of for both the fix and the tests. In the fullness of time it would be great to refactor this date/time code. For example, `LocalDate` would be a better underlying type for `DateData`. Ideally we wouldn't have to set the global time and timezone to write tests. But I think it makes sense to add more testing of this kind first before taking on a bigger refactor.

This is intended to be very narrow and just display the time component for datetime values used in outputs. This matches Enketo and Web Forms and makes it easier to compare behavior between the 3.

[now.xlsx](https://github.com/user-attachments/files/29150376/now.xlsx) is a good one to illustrate why we're making the change.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,23 +56,6 @@ if (!project.hasProperty("android")) {
         }
     }
 
-    test {
-        exclude '**/DateTimeTest.class'
-    }
-
-    tasks.register('isolatedDateTimeTest', Test) {
-        description = 'Runs DateTimeTest without parallelization'
-        group = 'verification'
-
-        testClassesDirs = sourceSets.test.output.classesDirs
-        classpath = sourceSets.test.runtimeClasspath
-
-        include '**/DateTimeTest.class'
-        maxParallelForks = 1
-    }
-
-    check.dependsOn isolatedDateTimeTest
-
     tasks.named('compileJava', JavaCompile) {
         options.encoding = "UTF-8"
         options.compilerArgs << "-Xlint:deprecation"
--- a/build.gradle
+++ b/build.gradle
@@ -56,6 +56,23 @@ if (!project.hasProperty("android")) {
         }
     }
 
+    test {
+        exclude '**/DateTimeTest.class'
+    }
+
+    tasks.register('isolatedDateTimeTest', Test) {
+        description = 'Runs DateTimeTest without parallelization'
+        group = 'verification'
+
+        testClassesDirs = sourceSets.test.output.classesDirs
+        classpath = sourceSets.test.runtimeClasspath
+
+        include '**/DateTimeTest.class'
+        maxParallelForks = 1
+    }
+
+    check.dependsOn isolatedDateTimeTest
+
     tasks.named('compileJava', JavaCompile) {
         options.encoding = "UTF-8"
         options.compilerArgs << "-Xlint:deprecation"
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -56,6 +56,23 @@ if (!project.hasProperty("android")) {
         }
     }
 
+    test {
+        exclude '**/DateTimeTest.class'
```
seadowg: Did you actually run into problems that motivated the changes to this file. That would confuse me for two reasons:

1. `maxParallelForks` should default to 1
2. `maxParallelForks`being greater than 1 should spawn separate **process** for each test "fork" - messing with state static state (like with `TimeZone.setDefault`) should not be an issue

Maybe I'm missing a detail here?

seadowg (changes requested):

build.gradle
lognaturel: Good question, should have written about this in the PR description.

I briefly explored whether I could find a way not to modify global state but anything I could come up with was going to be a big change. I felt somewhat sheepish about changing global state and my understanding is that there are some cases like killing a test in which `@After` might not run. So I thought isolating that test class could offer greater protection against that edge case. Really what I wanted was JUnit 5's `@Isolated` tag but we use JUnit 4.

When I read about how to isolate test classes in JUnit 4, I saw examples with explicit `maxParallelForks` of 1 but you're right that since we're talking about forked JVM processes that doesn't seem like it would help anything.

I'm taking all of that attempt at blocking parallelization out for now. I expect to follow up with more tests that modifying timezone and time and I'll see what kind of protection is needed then if applicable.

seadowg: > I briefly explored whether I could find a way not to modify global state but anything I could come up with was going to be a big change. I felt somewhat sheepish about changing global state and my understanding is that there are some cases like killing a test in which @After might not run.

The only cases where `@After` won't run is if the test process itself is killed. For modifying static state, this doesn't any reprecussions (the process doesn't exist so there's not state to mess up). The only time you need to worry about it is when you're modifying state that exists outside the test process. An example would be test setup/teardown that mess with animation speeds in Android emulator.

Generally, when messing with time you want to avoid static state by just treating it like any other dependency to the logic under test (which it is), but obviously reworking all of JavaRosa to take time as an input would be a big job to block this fix on. 

For the moment, the biggest risk with tests like this that do static static setup/teardown is messing up that setup/teardown code. The best way to avoid that in my opinion is to encapsulate it in a JUnit `@Rule` so that it's contained and reusable. This is a perfect use case for subclassing `ExternalResource`.

seadowg (changes requested):
https://github.com/getodk/javarosa/pull/854#discussion_r3457904078

lognaturel:
Pulled out rule and added one more test to document edge case.

PR #859, opened 2026-08-13
Remove synchronization on java.lang.Boolean

### LINKED ISSUE #858: `XPathLazyNodeset` gives appearance of being thread-safe but is not
`XPathLazyNodeset` synchronizes on its `evaluated` field:
```
Boolean evaluated = Boolean.FALSE;

synchronized (evaluated) {
    ...
    evaluated = true;
}
```

This doesn't actually provide thread safety because `Boolean.FALSE` and `Boolean.TRUE` are shared cached objects. Assigning `evaluated = true` also changes the object used as the monitor so two threads could synchronize on different objects.

As far as I can tell, `XPathLazyNodeset` is local, transient evaluation state. In general, JavaRosa itself is single-threaded and is used by Collect in a single-threaded way so the likelihood of an instance being accessed concurrently is very low. Other consumers could possibly try to use individual parts of JR in a multithreaded way but that seems very unlikely.

There's unlikely to be user impact so this isn't high-priority to change but https://github.com/getodk/javarosa/issues/857 points out that in JDK 28 (~March 2027), this code will be an error. We could:
* Wait until/if we upgrade to a JDK where a change becomes necessary
* Remove the failed attempt at synchronization by removing the `synchronized` blocks entirely
* Swap in a lock `Object`

I prefer doing nothing or removing the attempt at synchronization unless we have solid evidence that a single `XPathLazyNodeset` can be accessed by multiple threads.

Yqwed:
This makes sure that `evaluated` is accessed while holding the same monitor.
Also makes this class value classes ready.

Closes https://github.com/getodk/javarosa/issues/858

./gradlew build

Fixes immediate problem.

Makes their code JEP-401 (value classes) ready.

Probably not.

No

### DIFF
--- a/src/main/java/org/javarosa/xpath/XPathLazyNodeset.java
+++ b/src/main/java/org/javarosa/xpath/XPathLazyNodeset.java
@@ -29,6 +29,7 @@ import org.javarosa.xpath.expr.XPathPathExpr;
  */
 public class XPathLazyNodeset extends XPathNodeset {
 
+    private final Object lock = new Object();
     Boolean evaluated = Boolean.FALSE;
     TreeReference unExpandedRef;
 
@@ -46,7 +47,7 @@ public class XPathLazyNodeset extends XPathNodeset {
 
 
     private void performEvaluation() {
-        synchronized(evaluated) {
+        synchronized(lock) {
             if(evaluated.booleanValue()) {
                 return;
             }
@@ -71,7 +72,7 @@ public class XPathLazyNodeset extends XPathNodeset {
      * existed, but didn't, rather than a reference which could not represent a real node).
      */
     public Object unpack () {
-        synchronized(evaluated) {
+        synchronized(lock) {
             if(evaluated.booleanValue()) {
                 return super.unpack();
             }


lognaturel:
Thanks for the interest in this, @Yqwed. Closing because it doesn't match the preferred approaches at https://github.com/getodk/javarosa/issues/858. If we wanted to maintain the synchronization attempt we would want to remove `evaluated` entirely. 

>  Also makes this class value classes ready.

This is not a goal we have at the moment.

Yqwed:
> If we wanted to maintain the synchronization attempt we would want to remove evaluated entirely.

Current implementation of `performEvaluation` makes sure that it's called only once and that's tracked in `evaluated`. As a maintainer you know better.

> This is not a goal we have at the moment.

To be fair I find your prioritization quite perplexing - value classes will be released quite soon and your users won't be able to test their apps due to these synchronization and instead you decided to focus on something which I assume no one reported before and probably not a problem at all given your suggestion

> Remove the failed attempt at synchronization by removing the synchronized blocks entirely

But, again, you're the maintainer it's all up to you.

lognaturel:
> performEvaluation makes sure that it's called only once and that's tracked in evaluated

That's right, thanks!

PR #861, opened 2026-08-20
Bump version for release

seadowg:
Does what it says on the tin! We're going with 6.0.0 as we've made a few changes that could arguably be considered breaking.


lognaturel (approved):
Thanks!

PR #863, opened 2026-09-22
Test nodeset used as string

lognaturel:
Adds tests to illustrate https://github.com/getodk/javarosa/issues/862

Verified expected behavior against XPath 1.0 spec.

I could have just put the tests on the issue but I think that having them in the code is most likely to be helpful.

No user-facing changes.

No.

No. The issue acts as another entry point into this area.


src/test/java/org/javarosa/core/model/TriggerableDagTest.java
```diff
@@ -1763,6 +1763,43 @@ public void issue_119_target_question_should_be_relevant() throws IOException, X
         assertThat(scenario.getAnswerNode("/data/outer/inner"), is(nonRelevant()));
         assertThat(scenario.getAnswerNode("/data/outer/inner/target_question"), is(relevant()));
     }
+
+    @Ignore("Group node dependencies should include all descendant nodes (issue 862)")
```
lognaturel, line 1767: Ported from Web Forms: https://github.com/getodk/central-frontend/pull/1913/changes#diff-17041ecb2711c4266d0287e6531a122e6ded477250841d0c376ef141fd70d260R244

lognaturel: FYI @latin-panda

